PR #152, opened 2014-09-19
WIP: Add source builds for HDF5 on Linux

GlenHertz:
Hi, 

I don't have root access so I added support to build HDF5 from sources.   I tried the above but it failed.  Then I went into the src/ directory and ran `./configure --prefix=/path/to/.julia/v0.3/HDF5/deps/usr` and `make install`.  Then it works from Julia.  So I don't understand how to use the build system well enough to figure out this trivial error.  Hopefully someone knows and we can get HDF5 working when a user doesn't have root access.

Glen

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -3,9 +3,29 @@ using BinDeps
 @BinDeps.setup
 
 @linux_only begin
-    hdf5 = library_dependency("libhdf5")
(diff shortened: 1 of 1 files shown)


timholy:
Seems to have been superseded by #291.

PR #177, opened 2014-11-14
add support for the Nix package manager

rened:
Will require https://github.com/JuliaLang/BinDeps.jl/pull/115 to work.


rened:
I will reopen this once I have a homebrew-like nix manager running.

PR #191, opened 2014-12-23
Demo of custom serialization

timholy:
I needed to create a custom format for a type I'm working on, so I thought while learning how to do this I'd create a demo for the benefit of others. I suspect the main issue worth discussing is whether this is the best approach, or would it be better/simpler to create custom `read` and `write` methods? CC @simonster.

Once this settles, I'll also add this material to the documentation.

### DIFF
--- a/src/JLD.jl
+++ b/src/JLD.jl
@@ -7,7 +7,7 @@ using HDF5, Compat
 # Add methods to...
 import HDF5: close, dump, exists, file, getindex, setindex!, g_create, g_open, o_delete, name, names, read, size, write,
              HDF5ReferenceObj, HDF5BitsKind, ismmappable, readmmap
-import Base: length, endof, show, done, next, start, delete!, sizeof
+import Base: convert, length, endof, show, done, next, start, delete!, sizeof
 
 # .jld files written before v"0.4.0-dev+1419" might have Uint32 instead of UInt32 as the typename string.
 # See julia issue #8907
@@ -102,14 +102,6 @@ immutable TypeMismatchException <: Exception
 end
 show(io::IO, e::TypeMismatchException) =
     print(io, "stored type $(e.typename) does not match currently loaded type")
-
-# Wrapper for associative keys
-# We write this instead of the associative to avoid dependence on the
-# Julia hash function
-immutable AssociativeWrapper{K,V,T<:Associative}
-    keys::Vector{K}
-    values::Vector{V}
-end
 
 include("jld_types.jl")
 
@@ -317,7 +309,7 @@ function read(parent::Union(JldFile, JldGroup), name::ByteString)
     finally
         close(obj)
     end
-    val
+    readas(val)
 end
 read(parent::Union(JldFile,JldGroup), name::Symbol) = read(parent,bytestring(string(name)))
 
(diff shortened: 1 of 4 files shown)


timholy:
OK, this has been updated. Much simpler this way.

The biggest issue is whether I've added `readas` and `writeas` in every place that needs it.

mbauman:
This is great.  I've been thinking about this in the context of #212 and https://github.com/Keno/SIUnits.jl/pull/47 (type definitions changing in Base and other libraries), and I have a hunch you are, too.  It'd be nice to do things like rename and change types with this, but how we serialize types (#204) is going to make or break this.

Adding `readas` in [JLD.julia_type](https://github.com/timholy/HDF5.jl/blob/70af3a0e9b884f8f705800d5c39bbba80ef7035d/src/JLD.jl#L775) allows adding type parameters gracefully:

```julia
$ julia -q
julia> using HDF5, JLD
       type Foo{A} end
       save("test.jld", "x", Foo{1}())
       quit()

$ julia -q
julia> using HDF5, JLD
       type Foo{A,B} end
       JLD.readas{A}(::Foo{A}) = Foo{A,0}()
       JLD.readas{A}(::Type{Foo{A}}) = Foo{A,0}
       load("test.jld", "x")
Foo{1,0}()
```

Which is really awesome.  But it won't work for things like `Set{Foo{A}}`… but if parameter lists become serialized lists it automagically will.  I've not figured out how to deal with changing names yet.

simonster:
I think this may have to be hacked into `jld_types.jl` to work for some bits type cases. I'll look more closely later today.

timholy:
Thanks to both for looking.

Yes, @mbauman, it seemed sensible to start my getting back to HDF5 here, since it's possible it could help with a subset of recent challenges. Here's some mildly-informative goofing around:
```julia
julia> type MyType
           a::Int
       end

julia> t = MyType(3)
MyType(3)

julia> using HDF5, JLD

julia> @save "data.jld" t
```
Start a new session
```julia
julia> using HDF5, JLD

julia> t1 = load("data.jld", "t")
WARNING: type MyType not present in workspace; reconstructing
JLD.##MyType#32267(3)

julia> type MyType  # in the meantime, we've redefined the type
           a::Int
           b::Float32
       end

julia> eval(JLD, :(readas(x::$(typeof(t1))) = Main.MyType(x.a, 0.0f0)))

readas (generic function with 3 methods)

julia> load("data.jld", "t")
ERROR: stored type MyType does not match currently loaded type
 in jldatatype at /home/tim/.julia/v0.4/HDF5/src/jld_types.jl:646
 in read at /home/tim/.julia/v0.4/HDF5/src/JLD.jl:323
 in read at /home/tim/.julia/v0.4/HDF5/src/JLD.jl:308
 in anonymous at /home/tim/.julia/v0.4/HDF5/src/JLD.jl:990
 in jldopen at /home/tim/.julia/v0.4/HDF5/src/JLD.jl:229
 in load at /home/tim/.julia/v0.4/HDF5/src/JLD.jl:989
```
Obviously not there yet, but this direction seems worth chewing over a bit more.

timholy:
This issue was moved to JuliaLang/JLD.jl#3

Issue #192, opened 2014-12-31
Simple array not longer mmappable in HDF5 0.4.x?


rened:
The following snippet returns `true` in `0.3.0`, but returns `false` in `0.4.x` as well as the current master. Am I overlooking something? Thanks!
(Running Julia 0.3.3 on OSX 10.9)

```jl
using HDF5, JLD
if Pkg.installed("HDF5") < v"0.4.0"
	file = jldopen(tempname(), "w")
elseif Pkg.installed("HDF5") <= v"0.4.6"
	file = jldopen(tempname(), "w", mmaparrays=true)
else
	file = jldopen(tempname(), "w", compress=false, mmaparrays=true)
end
write(file, "a", [1:3])
@show ismmappable(file["a"])
```

rened:
Closed via #195

Issue #178, opened 2014-11-15
Enable Blosc compression of JLD files by default?


stevengj:
As discussed in #173 and #174, it is worth considering enabling `compress=true` by default when saving JLD files.  In all the benchmarks I've done so far, even with single-threaded Blosc this has a performance within a factor of two of uncompressed HDF5 (with the worst case being random data that does not benefit from compression), and is faster than uncompressed HDF5 for data that compresses well.

(This is dramatically better than zip/deflate compression, which can be orders of magnitude slower than uncompressed HDF5.)

Any other stress tests would be appreciated (just pass `compress=true` to `save` or `jldopen`).  Note that compression is not used for small datasets (< 8k).

cc: @jakebolewski, @dmbates, @kmsquire, @StefanKarpinski

stevengj:
Note that Blosc has been working on Windows for a while.

However, I don't see a clamor of need for compressed HDF5/JLD files, so maybe we should just leave it as an option.

davidssmith:
*clamors*

Doesn't Blosc improve read performance on large files?  If so, seems worth enabling.

stevengj:
A downside of enabling it by default is that it will be harder for other HDF5 applications to extract data from the file (assuming they don't link to Blosc too).

davidssmith:
Does it make any sense to enable by default on JLD but not for HDF5?

stevengj:
@davidssmith, definitely this would only be for JLD, and it is true that most people saving files for interchange with other programs would probably use the raw HDF5 interface.

timholy:
I'm fine with turning it on if (1) it doesn't substantially hurt performance for data that compress poorly, and (2) if people besides myself take responsibility for fixing any breakages :-).

stevengj:
@jakebolewski, could you re-run your benchmarks now that things should be working?

kleinhenz:
closing since it looks like the consensus is that we would only do this for JLD not HDF5.

PR #195, opened 2015-01-02
fix for #192 + allow to ensure mmappability on a per-write basis

rened:
This PR makes sure
* mmaparrays takes precendence over compress
* only uses the compact storage format if mmaparrays is false
* Unit test for #192 
* allows to ensure mmappability for individual writes to a file for which compress==true

So as long as mmaparrays==true during construction or mmap==true at write time, data is always stored in a mmappable way.

### DIFF
--- a/src/JLD.jl
+++ b/src/JLD.jl
@@ -68,10 +68,10 @@ type JldFile <: HDF5.DataFile
     function JldFile(plain::HDF5File, version::VersionNumber=version_current, toclose::Bool=true,
                      writeheader::Bool=false, mmaparrays::Bool=false,
                      compress::Bool=false)
-		if compress && mmaparrays
-			error("A JLD file cannot have compress=true and mmaparrays=true set at the same time")
-		end
-		f = new(plain, version, toclose, writeheader, mmaparrays, compress,
+        if compress && mmaparrays
+            error("A JLD file cannot have compress=true and mmaparrays=true set at the same time")
+        end
+        f = new(plain, version, toclose, writeheader, mmaparrays, compress,
                 Dict{HDF5Datatype,Type}(), Dict{Type,HDF5Datatype}(),
                 Dict{HDF5ReferenceObj,WeakRef}(), ByteString[])
         if toclose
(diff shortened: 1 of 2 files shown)


src/JLD.jl
```diff
@@ -92,6 +95,10 @@ iscompressed(f::JldFile) = f.compress
 iscompressed(g::JldGroup) = g.file.compress
 iscompressed(d::JldGroup) = d.file.compress
 
+ismmapped(f::JldFile) = f.mmaparrays
+ismmapped(g::JldGroup) = g.file.mmaparrays
+ismmapped(d::JldGroup) = d.file.mmaparrays
```
simonster: This is probably my fault, but this and `iscompressed` above have two methods for `JldGroup` when only one is needed.

test/jld.jl
```diff
@@ -255,6 +255,14 @@ typ{T}(x::Int64, ::Type{T}) = Base.box(Typ{T}, Base.unbox(Int64,x))
 abstract UnexportedT
 end
 
+# test mmapping of small arrays (Issue #192)
+fid = jldopen(fn, "w", mmaparrays=true)
+write(fid, "a", [1:3])
+assert(ismmappable(fid["a"]))
```
simonster: This would be better as `@test ismmappable(fid["a"])`, since that will provide a more informative error message on failure.

src/JLD.jl
```diff
@@ -68,7 +68,10 @@ type JldFile <: HDF5.DataFile
     function JldFile(plain::HDF5File, version::VersionNumber=version_current, toclose::Bool=true,
                      writeheader::Bool=false, mmaparrays::Bool=false,
                      compress::Bool=false)
-        f = new(plain, version, toclose, writeheader, mmaparrays, compress,
+		if compress && mmaparrays
```
simonster: This and below should be indented with spaces rather than tabs.

simonster:
This sounds good to me in principle. If we enable compression by default (#178) we may want to make `mmaparrays=true` override `compress=true`, or alternatively allow both `compress=true` and `mmaparrays=true` in the read-only case (where `compress=true` will have no effect).

rened:
Thanks @simonster, I've included your feedback! I've also set it up that `mmaparrays` overrules `compress`, that seems to be the most desirable behavior from a user perspective.

rened:
I have added the ability to ensure mmappability at write time for individual items when writing to a compressed file.
This, IMHO, gives the best user experience (especially once compression is default):
* per default all writes are compressed / stored in the most compact format (smallest file, fastest IO)
* individual items (which need to be mmappable) can be written ensuring mmappability
(instead of requiring the entire file to be saved uncompressed)

timholy:
Very nice, many thanks!

Issue #186, opened 2014-12-03
[PkgEval] HDF5 may have a testing issue on Julia 0.3 (2014-12-03)


IainNZ:
[PackageEvaluator.jl](https://github.com/IainNZ/PackageEvaluator.jl) is a script that runs nightly. It attempts to load all Julia packages and run their tests (if available) on both the stable version of Julia (0.3) and the nightly build of the unstable version (0.4). The results of this script are used to generate a [package listing](http://pkg.julialang.org/) enhanced with testing results.

#### On Julia 0.3
* On **2014-12-02** the testing status was `Tests pass.`
* On **2014-12-03** the testing status changed to `Package doesn't load.`

`Tests pass.` means that PackageEvaluator found the tests for your package, executed them, and they all passed.

`Package doesn't load.` means that PackageEvaluator did not find tests for your package. Additionally, trying to load your package with `using` failed.

*This issue was filed because your testing status became worse. No additional issues will be filed if your package remains in this state, and no issue will be filed if it improves. If you'd like to opt-out of these status-change messages, reply to this message saying you'd like to and @IainNZ will add an exception. If you'd like to discuss [PackageEvaluator.jl](https://github.com/IainNZ/PackageEvaluator.jl) please file an issue at the repository. For example, your package may be untestable on the test machine due to a dependency - an exception can be added.*

*Test log*:
```
>>> 'Pkg.add("HDF5")' log
INFO: Installing BinDeps v0.3.6
INFO: Installing HDF5 v0.4.6
INFO: Installing SHA v0.0.3
INFO: Installing URIParser v0.0.3
INFO: Building HDF5
INFO: Package database updated

>>> 'using HDF5' log
HDF5-DIAG: Error detected in HDF5 (1.8.14) thread 140103333197632:
  #000: H5P.c line 299 in H5Pcreate(): not a property list class
    major: Invalid arguments to routine
    minor: Inappropriate type
HDF5-DIAG: Error detected in HDF5 (1.8.14) thread 140103333197632:
  #000: H5Pstrcpl.c line 159 in H5Pset_char_encoding(): can't find object for ID
    major: Object atom
    minor: Unable to find atom information (already closed?)
  #001: H5Pint.c line 3381 in H5P_object_verify(): property list is not a member of the class
    major: Property lists
    minor: Unable to register new atom
  #002: H5Pint.c line 3331 in H5P_isa_class(): not a property list
    major: Invalid arguments to routine
    minor: Inappropriate type
[...]

PR #200, opened 2015-01-07
Fix whitespace, and add gc_enable at the end of the test

timholy:
@simonster, was this missing a `gc_enable()` at the end of the test? (From what I could see, it was leaving it in disabled state.)

Motivated by (fruitlessly) looking into #186.


simonster:
Yes, I believe you are correct.

Issue #198, opened 2015-01-04
 typeassert error preventing load of saved jld file after 0.3.4 updates


disleyland:
Hi,  my first post.

I just upped to Julie 0.3.4 and updated packages.  but when i tried to load my jld file, i got this error:

ERROR: type: jlconvert: in typeassert, expected Dict{Symbol,Int64}, got Dict{Symbol,Union(Real,AbstractArray{Real,1})}

Do i have to downgrade to retrieve data, save as csv, reload, resave, or is there an alternative way to recover data?

timholy:
Oh, drat. Not being able to read old .jld files is precisely what we don't want happening.

Let's try to fix the error without forcing you to convert to a different format. Two questions:
- Is there any more information to that error message? Any more detail you can provide would be extremely helpful; this is probably not yet enough information to go on.
- If the .jld file is not too big and you don't feel that it has sensitive information, feel free to email it to me and I'll try it directly.

disleyland:
HI Tim,

The full message is:

while loading In[2], in expression starting on line 56

 in jlconvert at /Users/richard/.julia/v0.3/HDF5/src/jld_types.jl:397
 in read_scalar at /Users/richard/.julia/v0.3/HDF5/src/JLD.jl:356
 in read at /Users/richard/.julia/v0.3/HDF5/src/JLD.jl:328
 in read_ref at /Users/richard/.julia/v0.3/HDF5/src/JLD.jl:470
 in jlconvert at /Users/richard/.julia/v0.3/HDF5/src/jld_types.jl:397
 in read_scalar at /Users/richard/.julia/v0.3/HDF5/src/JLD.jl:356
 in read at /Users/richard/.julia/v0.3/HDF5/src/JLD.jl:328
 in read at /Users/richard/.julia/v0.3/HDF5/src/JLD.jl:313
 in anonymous at /Users/richard/.julia/v0.3/HDF5/src/JLD.jl:972
 in jldopen at /Users/richard/.julia/v0.3/HDF5/src/JLD.jl:234
 in load at /Users/richard/.julia/v0.3/HDF5/src/JLD.jl:971

[...]

PR #201, opened 2015-01-07
Modernize .travis.yml and fix beginswith->startswith deprecation

timholy:
Not sure if I've got the syntax correct, but we'll see. This is more complicated than is typical, because:
- I want to submit to Coveralls only from the nightly build, since I want to use features like `--inline=no`
- I am running the tests for two packages (HDF5 and JLDArchives) in a single session, to make sure the coverage files include the tests of both

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,38 +1,16 @@
-language: cpp
+language: julia
 os:
     - linux
     - osx
-compiler: 
-    - gcc
+julia:
+    - release
+    - nightly
 notifications:
     email: false
-env:
-  matrix:
-    - JULIAVERSION="juliareleases"
-    - JULIAVERSION="julianightlies"
-before_install:
-    - if [ `uname` = "Linux" ]; then
-        sudo add-apt-repository ppa:staticfloat/julia-deps -y;
-        sudo add-apt-repository ppa:staticfloat/${JULIAVERSION} -y;
-        sudo apt-get update -qq -y;
-        sudo apt-get install libpcre3-dev julia -y;
-      elif [ `uname` = "Darwin" ]; then
-        if [ "$JULIAVERSION" = "julianightlies" ]; then
-          wget -O julia.dmg "http://status.julialang.org/download/osx10.7+";
-        else
-          wget -O julia.dmg "http://status.julialang.org/stable/osx10.7+";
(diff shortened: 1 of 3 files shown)


timholy:
Looks like the syntax isn't correct. CC @ninjin, @tkelman, any tips here?

tkelman:
I'll have to check those docs I wrote, but I think you want to change `$JULIAVERSION` to `$TRAVIS_JULIA_VERSION`

timholy:
It's also complaining
```
/home/travis/build.sh: line 41: [: =: unary operator expected
```
Is using `=` for comparisons correct?

tkelman:
Pretty sure that's just a bash symptom of trying to do `if [ $JULIAVERSION = "nightly" ]` when `JULIAVERSION` is not defined, since it has no quotes around it.

timholy:
Indeed, that did the trick. Thanks!

(And wow, those tests run much more quickly with the modern script.)

tkelman:
> And wow, those tests run much more quickly with the modern script

That part was a big surprise to me. Here I thought apt-get was a reasonably fast way to get a binary onto a machine but apparently not. You don't have an education account at Travis do you? I think those might route to the new docker-based workers if you don't use `sudo` at all. Or you might need to explicitly ask for that with `sudo: false` in your `.travis.yml` file, not sure. Supposedly those are even faster.

timholy:
Whew, looks like I finally ran out of ways to make syntax errors.

Turning on these various settings made a big difference in the (new) coverage percentages: 59% with the old travis script and 72% with the new. Most of that probably comes from combining the results of HDF5 & JLDArchive tests, but looking at the results, there's some real benefit to turning off the inlining (functions I _knew_ were tested, indeed are).

tkelman:
yeah that makes sense, I misread your Julia pull request the first time around

romainFr:
It seems that the deprecation fix isn't compatible with 0.3.4 (startswith is not defined there), but 0.3.4 users are likely to want to use the checkout version of HDF5 because of #198.

simonster:
@romainFr I believe this should work if you update Compat.jl.

timholy:
yes, just do `Pkg.update()`.

PR #207, opened 2015-01-12
Add Pacman as a provider.

MichaelHatherly:
I got a nice surprise today when Images.jl installed imagemagick for me. Unfortunately HDF5.jl didn't install it's required lib. I believe this change is all that's needed for that to happen.

I also noticed that I've gone a stripped a whitespace in the file. If you'd rather I didn't I can undo that bit of the change.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,10 +1,11 @@
-using BinDeps 
+using BinDeps
 
 @BinDeps.setup
 
 @linux_only begin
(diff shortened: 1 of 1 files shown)


timholy:
This will be nice.

I have a hard time imagining that the travis failure in building Blosc is due to this PR, so pinging @stevengj.

stevengj:
Could it have been a temporary glitch?  Seems to be a downloading problem, and I can't reproduce on my Linux system.

timholy:
Thanks!

MichaelHatherly:
Thanks @timholy and @stevengj.

PR #213, opened 2015-01-27
Throw error when saving unsupported type parameters

mbauman:
Only allow bitstypes that show as AST literals and make sure that they read
back exactly as they are saved. Use show (instead of print) to preserve as many
Julian type distinctions as we can e.g., 1 vs 0x01. A different implementation
will be required to support custom immutables or things as simple as Int16(1).

### DIFF
--- a/src/JLD.jl
+++ b/src/JLD.jl
@@ -815,7 +815,19 @@ function full_typename(io::IO, file::JldFile, jltype::(Type...))
     end
     print(io, ')')
 end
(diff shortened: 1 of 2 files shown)


mbauman:
I think this should be good to go once travis green lights it.  I'm seeing the dreaded `<?::<?::<?::<?::<?::<?::>>>>>>` when running the tests on a 2-day old julia-0.4 and master HDF5, but it works in my interactive tests.

Issue #212, opened 2015-01-26
Cannot load variables in v0.4 that load fine in v0.3


milktrader:
This is breaking tests on Quandl for current nightlies and when I tried loading in REPL I likewise got this error

``` julia
julia> load(Pkg.dir("Quandl/test/response.jld"))["missing"]
WARNING: type JLD.AssociativeWrapper{Core.String,Core.String,Base.Dict{Core.String,Core.String}} not present in workspace; reconstructing
WARNING: type Core.String not present in workspace; interpreting array as Array{Any}
WARNING: type Core.String not present in workspace; interpreting array as Array{Any}
ERROR: `convert` has no method matching convert(::Type{Dict{AbstractString,AbstractString}}, ::##JLD.AssociativeWrapper{Core.String,Core.String,Base.Dict{Core.String,Core.String}}#12109)
 in jlconvert at /Users/Administrator/.julia/v0.4/HDF5/src/jld_types.jl:406
 in read_scalar at /Users/Administrator/.julia/v0.4/HDF5/src/JLD.jl:359
 in read at /Users/Administrator/.julia/v0.4/HDF5/src/JLD.jl:331
 in read at /Users/Administrator/.julia/v0.4/HDF5/src/JLD.jl:316
 in anonymous at /Users/Administrator/.julia/v0.4/HDF5/src/JLD.jl:978
 in jldopen at /Users/Administrator/.julia/v0.4/HDF5/src/JLD.jl:237
 in load at /Users/Administrator/.julia/v0.4/HDF5/src/JLD.jl:977
```

milktrader:
https://github.com/milktrader/Quandl.jl/issues/74

mbauman:
Thanks for the report.  Looks like we'll need a switch for `AbstractString` like that for `UInt` (https://github.com/timholy/HDF5.jl/blob/efbb3f636bbb477b48ae60dceee80ef0d5e7ec9c/src/JLD.jl#L14-L18).  I probably won't have time to fully test a solution for a few days, but I can do this if nobody beats me to it.

timholy:
@mbauman, how about this for a deal: I'll solve this one, but if you haven't dealt with https://github.com/timholy/HDF5.jl/pull/203#issuecomment-69121466 I'd love to have that done sooner rather than later. Related: #211, although we should figure out how to implement that one.

mbauman:
Deal. :)

musm:
stale

Issue #209, opened 2015-01-16
JLD interoperability between 32 and 64 bits


GunnarFarneback:
What are the ambitions with respect to interoperability between 32 and 64 bit Julia for the JLD format?

Saving a substring from 32 bit Julia 0.3

```
julia> using HDF5, JLD

julia> s = SubString{ASCIIString}("abc", 2, 3)
"bc"

julia> save("foo.jld", "s", s)
```

and then loading it in 64 bit Julia fails:

```
julia> using HDF5, JLD

julia> load("foo.jld")
ERROR: stored type Base.SubString{Core.ASCIIString} does not match currently loaded type
 in jldatatype at /home/gunnar/.julia/HDF5/src/jld_types.jl:637
 in read at /home/gunnar/.julia/HDF5/src/JLD.jl:299
 in read at /home/gunnar/.julia/HDF5/src/JLD.jl:284
 in anonymous at /home/gunnar/.julia/HDF5/src/JLD.jl:932
 in jldopen at /home/gunnar/.julia/HDF5/src/JLD.jl:205
 in load at /home/gunnar/.julia/HDF5/src/JLD.jl:931
```

Nothing terribly surprising about it since the SubString type contains fields of type Int, but rather annoying since substrings tend to crop up anytime you do string processing.

This is of course not limited to substrings but happens to any type with explicit Int fields, both those you define yourself and those you get from Base or packages. Is there anything constructive that can be done about it or should the problem just be documented?

GunnarFarneback:
I see now that there's related discussion in #202.

timholy:
This is important. I am pretty swamped right now, though, so I confess I haven't really gotten serious about a plan for a fix.

kleinhenz:
closing since this looks like a JLD issue.

Issue #202, opened 2015-01-07
sparse jld file saved in win not work in linux


paulanalyst:
data=load("D_z_win.jld","D")
ERROR: .... on the end of bug
File D_z_win.jld was created at win 7/64 in this same way like this sparse (I,J,V)
On the win machine is julia 3.0

samsung2@samsung2:~$ julia
_
_ _ ()_ | A fresh approach to technical computing
() | () () | Documentation: http://docs.julialang.org
_ _ _| | __ _ | Type "help()" for help.
| | | | | | |/ ` | |
| | || | | | (| | | Version 0.3.3 (2014-10-21 20:18 UTC)
/ |\'|||_'| | Official http://julialang.org release
|_/ | i686-linux-gnu

julia> I=vec(readcsv("I.txt"));

julia> J=vec(readcsv("J.txt"));

julia> V=vec(readcsv("V.txt"));

julia> D=sparse(int(I),int(J),int(V));

julia> using HDF5,JLD

julia> save("D.jld","D",D)

julia> data=load("D.jld","D")
10000000x30000 sparse matrix with 1000000 Int32 entries:
[170590 , 1] = 1
[275820 , 1] = 1
[3131540 , 1] = 1
[3182060 , 1] = 1
[3286470 , 1] = 1
[3398480 , 1] = 1
[3690470 , 1] = 1
⋮
[3215620 , 30000] = 1
[3509960 , 30000] = 1
[4311630 , 30000] = 1
[4984860 , 30000] = 1
[5425410 , 30000] = 1
[6618640 , 30000] = 1
[6698030 , 30000] = 1
[7282230 , 30000] = 1

julia> data=load("D_z_win.jld","D")
ERROR: stored type Base.SparseMatrix.SparseMatrixCSC{Core.Float64,Core.Int64} does not match currently loaded type
in jldatatype at /home/samsung2/.julia/v0.3/HDF5/src/jld_types.jl:637
in read at /home/samsung2/.julia/v0.3/HDF5/src/JLD.jl:331
in read at /home/samsung2/.julia/v0.3/HDF5/src/JLD.jl:316
in anonymous at /home/samsung2/.julia/v0.3/HDF5/src/JLD.jl:983
in jldopen at /home/samsung2/.julia/v0.3/HDF5/src/JLD.jl:237
in load at /home/samsung2/.julia/v0.3/HDF5/src/JLD.jl:982

Paul

simonster:
[...]

PR #214, opened 2015-01-28
Record creator information in JLD files

timholy:
I don't have an immediate plan to use this information, but it seems likely to be useful in several contexts. Might as well start recording it.

Possibly relevant for #202, #209, #212.


ggggggggg:
Looks like this belongs in `JLD.jl` since the packages have split, this pull request should be closed.

timholy:
Yeah, you're right. Thanks.

PR #217, opened 2015-02-28
Fix JuliaLang/julia#10289

simonster:
Don't try to store singleton objects on Julia versions that don't use memory to store them. This could cause a segfault when a singleton object was to be stored at the end of an object, since we'd write one byte too many. It didn't cause problems in other cases because we'd overwrite whatever we'd stored in that field when storing the next field.


simonster:
Travis failure is in the JLDArchives/JLD00 code (because Top is no longer a type).

timholy:
Thanks for tackling this and so many other issues!

PR #220, opened 2015-03-01
Allow writing tuples of tuples and fix #219

### LINKED ISSUE #219: Represent tuples with reconstructed types as tuples
Right now we reconstruct tuples that contain reconstructed types as immutables. I believe this should always be possible, but it would be better if we don't have to create a new type for them.

simonster:


simonster:
It seems the three new "missed" lines are actually [these](https://github.com/timholy/HDF5.jl/blob/sjk/219/src/jld_types.jl#L708-L711), which I expanded from a single line to multiple lines. They are definitely hit.

PR #228, opened 2015-04-17
WIP: adapt method definitions for tuple types

rened:
This PR allows to use `using HDF5` with the changes from https://github.com/JuliaLang/julia/commit/ef062117bab19959a6905b1bb6d7963b42dddea4
The package now loads in 0.4 master and loads + passes unit tests on 0.3 (when using https://github.com/JuliaLang/Compat.jl/pull/63), but the unit tests in 0.4 yield this error:
```
> ~/local/devdevjulia/julia test/runtests.jl
ERROR: LoadError: LoadError: MethodError: `one` has no method matching one(::Type{Any})
 in _mapreduce at reduce.jl:135
 in prod at reduce.jl:219
 in _setindex! at /Users/rene/.julia/v0.4/HDF5/src/plain.jl:1583
 in setindex! at /Users/rene/.julia/v0.4/HDF5/src/plain.jl:1618
 in include at ./boot.jl:250
 in include_from_node1 at ./loading.jl:129
 in include at ./boot.jl:250
 in include_from_node1 at loading.jl:129
 in process_options at ./client.jl:299
 in _start at ./client.jl:398
while loading /Users/rene/.julia/v0.4/HDF5/test/plain.jl, in expression starting on line 84
while loading /Users/rene/.julia/v0.4/HDF5/test/runtests.jl, in expression starting on line 3
```
It is caused by the array being passed to `prod` in https://github.com/rened/HDF5.jl/blob/master/src/plain.jl#L1583 being empty.

@timholy or @simonster, could you please take over from here? I don't know how to solve this properly.

PR #232, opened 2015-04-24
Fix tuple types for 0.4

rened:
Continuing here from #228 to allow for collaborative editing.

### DIFF
--- a/src/jld_types.jl
+++ b/src/jld_types.jl
@@ -382,7 +382,7 @@ function h5type(parent::JldFile, T::ANY, commit::Bool)
         id = HDF5.h5t_create(HDF5.H5T_COMPOUND, typeinfo.size)
         for i = 1:length(typeinfo.offsets)
             fielddtype = typeinfo.dtypes[i]
-            HDF5.h5t_insert(id, mangle_name(fielddtype, T.names[i]), typeinfo.offsets[i], fielddtype)
+            HDF5.h5t_insert(id, mangle_name(fielddtype, HDF5.getnames(T)[i]), typeinfo.offsets[i], fielddtype)
         end
     end
 
@@ -410,11 +410,11 @@ function _gen_jlconvert_type(typeinfo::JldTypeInfo, T::ANY)
             push!(args, quote
                 ref = unsafe_load(convert(Ptr{HDF5ReferenceObj}, ptr)+$h5offset)
                 if ref != HDF5.HDF5ReferenceObj_NULL
-                    out.$(T.names[i]) = convert($(T.types[i]), read_ref(file, ref))
+                    out.$(HDF5.getnames(T)[i]) = convert($(T.types[i]), read_ref(file, ref))
                 end
             end)
         else
-            push!(args, :(out.$(T.names[i]) = jlconvert($(T.types[i]), file, ptr+$h5offset)))
+            push!(args, :(out.$(HDF5.getnames(T)[i]) = jlconvert($(T.types[i]), file, ptr+$h5offset)))
         end
     end
     @eval function jlconvert(::Type{$T}, file::JldFile, ptr::Ptr)
@@ -491,7 +491,7 @@ const DONT_STORE_SINGLETON_IMMUTABLES = VERSION >= v"0.4.0-dev+385"
 function gen_jlconvert(typeinfo::JldTypeInfo, T::ANY)
     haskey(JLCONVERT_DEFINED, T) && return
 
-    if isempty(T.names)
+    if isempty(HDF5.getnames(T))
         if T.size == 0
             @eval begin
                 jlconvert(::Type{$T}, ::JldFile, ::Ptr) = $T()
@@ -524,7 +524,7 @@ end
 
 # Whether this datatype should be stored as opaque
 isopaque(t::@compat Tuple{Vararg{Type}}) = isa(t, ())
-isopaque(t::DataType) = isempty(t.names)
+isopaque(t::DataType) = isempty(HDF5.getnames(t))
 
 # The size of this datatype in the HDF5 file (if opaque)
 opaquesize(t::@compat Tuple{Vararg{DataType}}) = 1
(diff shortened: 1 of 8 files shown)


fundamental:
@rened if you need a hand getting this to work let me know where to start modifying as this issue is a blocker for me as I need to get at my data to start debugging/replicating a gc bug somewhere deep inside the base julia.

rened:
@fundamental Thanks! I am waiting to see whether Jeff's recent fix now works for @simonster

src/jld_types.jl
```diff
@@ -292,15 +328,16 @@ h5fieldtype{T,N}(parent::JldFile, ::Type{Array{T,N}}, ::Bool) = JLD_REF_TYPE
...
 else
-    h5fieldtype(parent::JldFile, T::(Type...), ::Bool) = JLD_REF_TYPE
+    h5fieldtype(parent::JldFile, T::TupleType, ::Bool) = JLD_REF_TYPE
 end
 
-function h5type(parent::JldFile, T::(Type...), commit::Bool)
-    !isa(T, (Union(Tuple, DataType)...)) && unknown_type_err(T)
-    T = T::(Union(Tuple, DataType)...)
+function h5type(parent::JldFile, T::TupleType, commit::Bool)
+    @assert isa(T, TupleType)
```
simonster: This is supposed to be a no-op, right?

simonster:
Jeff's fix helped a bit, but I think the underlying issue is actually a dispatch bug that I need to track down...

src/jld_types.jl
```diff
@@ -523,11 +560,12 @@ end
 ## Common functions for all non-special types (including gen_h5convert)
 
 # Whether this datatype should be stored as opaque
-isopaque(t::(Type...)) = isa(t, ())
-isopaque(t::DataType) = isempty(t.names)
+isopaque(t::TupleType) = t == EMPTY_TUPLE_TYPE
+# isopaque(t::DataType) = isempty(fieldnames(t))
+isopaque(t::DataType) = isa(t, TupleType) ? t == EMPTY_TUPLE_TYPE : isempty(fieldnames(t))
```
simonster: This is how I noticed JuliaLang/julia#10978, but I think `isopaque(t::TupleType)` should be called if `isa(t, TupleType)`, and not this method.

simonster:
Filed https://github.com/JuliaLang/julia/issues/10995 with test case

simonster:
Field https://github.com/JuliaLang/julia/issues/10998. This one I think I can work around.

sebastiang:
Wish I'd looked earlier to see this work going on! Was hacking away at it a bit myself. Let me know if you need any help. I just got to the `SimpleVector` issue myself, but perhaps you have a workaround? It caused me to file https://github.com/JuliaLang/julia/issues/10999

simonster:
The SimpleVector issue is not really a bug, just an implementation change. It was among the easier things to handle.

The PR here currently passes the test suite (although not the JLDArchives tests yet). However, you probably don't want to use it yet because I can't guarantee that any files you save with it will be readable by when this is in a state to be merged to master. Anything not involving tuples or types with fields that are tuples is probably okay, but anything else might not work. I also need to add more backwards compatibility tests to JLDArchives to see how badly this breaks reading older files, and fix that as best as I can.

sebastiang:
Understood! I have an aversion to magical reflection based serialization in any language, so was just planning on using the raw HDF5 interfaces mostly anyway.

timholy:
Thanks a bunch for this, everyone! I'll also link https://github.com/JuliaLang/julia/issues/10999#issuecomment-96158871 as a general issue of growing concern to me (but without me yet having carved out the time to find a general solution solution).

simonster:
I added test files to JLDArchives that have our entire test suite saved with 0.3 and master. Those tests all pass (files written with both versions can be read by both versions), except for the SubArray tests, which fail because the structure of the type has changed. (Maybe we should reconstruct the type in this case too instead of throwing an error.)

This doesn't mean everything works, because covering all our code paths isn't actually sufficient to make sure that all old files can be read by newer Julia, but at least the situation is not too bad.

simonster:
I can't figure out what's going on with the release build on OS X, which is intermittently failing with a completely empty log (??), but I suspect that's not related to the package at all. Other than that I think this is ready to merge.

src/jld_types.jl
timholy: Do we still need this?

src/jld_types.jl
```diff
@@ -631,6 +659,9 @@ function jldatatype(parent::JldFile, dtype::HDF5Datatype)
         if T == UnsupportedType
             warn("type $typename not present in workspace; reconstructing")
             T = reconstruct_type(parent, dtype, typename)
+        elseif T == Type
+            # https://github.com/JuliaLang/julia/issues/10998
```
timholy: Likewise here?

src/plain.jl
```diff
@@ -431,10 +431,8 @@ end
 type EmptyArray{T}; end
 
 # Stub types to encode fixed-size arrays for H5T_ARRAY
-immutable DimSize{N}; end  # Int-wrapper (can't use tuple of Int as param)
-immutable FixedArray{T,D<:(DimSize...)}; end
-dimsize{N}(::Type{DimSize{N}}) = N
-size{T,D}(::Type{FixedArray{T,D}}) = map(dimsize, D)::(Int...)
+immutable FixedArray{T,D}; end
+size{T,D}(::Type{FixedArray{T,D}}) = D
```
timholy: I had forgotten that this capability had gotten backported to 0.3. Nice!

timholy:
It all looks great. Thanks to both of you! Merge at will.

src/jld_types.jl
simonster: The tests fail without it, so the answer seems to be yes, although I'm not sure why.

src/jld_types.jl
simonster: Probably not. I'll check.

rened:
Thanks to you two for tackling the hard parts!

timholy:
I didn't do anything, you two deserve the thanks!

Issue #236, opened 2015-05-19
Error running dump() on a JLD (HDF5) file containing a DataFrame


josefsachsconning:
From [julia-users thread](https://groups.google.com/forum/?fromgroups=#!topic/julia-users/Kj6p4T2Kk4k)

```
Am I doing something wrong?  I get the same error in 0.3.8.
I notice that there is a method with signature
read(::HDF5.HDF5Dataset, ::Type{Array{HDF5.HDF5Compound,N}})
but not
read(::HDF5.HDF5Dataset, ::Type{HDF5.HDF5Compound})

               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "help()" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.4.0-dev+4825 (2015-05-14 00:12 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit ff35d98 (4 days old master)
|__/                   |  x86_64-w64-mingw32

julia> using DataFrames

julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
4x2 DataFrames.DataFrame
| Row | A | B   |
|-----|---|-----|
| 1   | 1 | "M" |
| 2   | 2 | "F" |
| 3   | 3 | "F" |
| 4   | 4 | "M" |

julia> using HDF5, JLD

julia> file = jldopen("df.jld", "w")
Julia data file version 0.1.0: df.jld

julia> file["df"] = df
4x2 DataFrames.DataFrame
| Row | A | B   |
|-----|---|-----|
| 1   | 1 | "M" |
| 2   | 2 | "F" |
| 3   | 3 | "F" |
| 4   | 4 | "M" |

julia> close(file)

julia> file = jldopen("df.jld", "r")
Julia data file version 0.1.0: df.jld

julia> dump(file)
JLD.JldFile
  plain: HDF5.HDF5File len 4
    _refs: HDF5.HDF5Group len 16
      00000001: HDF5Dataset (2,) : [HDF5.HDF5ReferenceObj(0x00000000000015d8),HDF5.HDF5ReferenceObj(0x0000000000002e70)]
      00000002: HDF5Dataset () : ERROR: MethodError: `read` has no method matching read(::HDF5.HDF5Dataset, ::Type{HDF5.HDF5Compound})
Closest candidates are:
[...]

PR #237, opened 2015-05-19
Adding Union to method signature related to issue #236

AndyGreenwell:
Addressing issue #236, updating the method signature of read in line 1306 to be a Union Array{HDF5Compound} and HDF5Compound, and adding corresponding tests to test/jld_dataframe.jl


timholy:
Thanks, @AndyGreenwell!

PR #239, opened 2015-06-02
Fix ntuple deprecation warnings

timholy:


timholy:
Hmm. I pushed 76e0265faec79ca5d11b8fb0678169c132fd322e directly (without going through a PR), because it worked on my machine. But this is worrisome:
https://travis-ci.org/timholy/HDF5.jl/builds
Note that one was using `Julia Version 0.4.0-dev+5114` and another `Julia Version 0.4.0-dev+5163`, so it's something that has been consistent over a couple of days' worth of master.

I can revert that commit, but the strange part is that I can't replicate this at all on my machine, even with an up-to-the minute build of julia.

simonster:
I can reproduce a hang, apparently in type inference in the JLDArchives tests, when running `julia -e 'reload("HDF5/test/runtests.jl"); workspace(); import LastMain.Compat, LastMain.HDF5; reload("HDF5/src/JLD.jl"); reload("JLDArchives/test/runtests.jl")'`, which is what Travis is trying to do. Just running the JLD or JLDArchives tests does not result in a hang.

simonster:
Opened JuliaLang/julia#11948

PR #247, opened 2015-07-30
Import Base.== to avoid deprecation warnings on 0.4-dev

samuelpowell:


timholy:
Thanks for all of these, @samuelpowell!

PR #248, opened 2015-08-03
fix mmap for 0.4

rened:
Silences a deprecation warning on 0.4.

I put the version-checking logic directly in here, as importing all of mmap's functionality into Compat seems to be overkill.

### DIFF
--- a/src/plain.jl
+++ b/src/plain.jl
@@ -1421,7 +1421,7 @@ function readmmap{T}(obj::HDF5Dataset, ::Type{Array{T}})
     if offset == reinterpret(Hsize, convert(Hssize, -1))
         error("Error mmapping array")
     end
-    mmap_array(T, dims, fdio(fd), convert(FileOffset, offset))
+    Mmap.mmap(fdio(fd), Array{T,length(dims)}, dims, offset)
 end
 
 function readmmap(obj::HDF5Dataset)


timholy:
One wouldn't have to add the functionality, one would simply need to make
```jl
@compat(Mmap.mmap(io, Array{T,N}, dims, offset))
```
be translated into
```jl
mmap_array(T, dims, io, offset)
```
on julia 0.3. See https://github.com/JuliaLang/Compat.jl/blob/2458d81963e14e44433f73bacdda50c5e3435f25/src/Compat.jl#L275-L335

rened:
Yes, but there are several variants for `Mmap.mmap`, dims vs length parameter, IO vs filename, therefor I was hesitant. Shall I make those available through Compat and rework this PR?

rened:
Appveyor bails while building HDF5, while downloading `zlib1`

timholy:
Yeah, seems like an opensuse problem (hopefully temporary).

timholy:
I suspect we only need a Compat entry for the variant that maps directly to `mmap_array`. It's just a tool for people to make minimal changes to code and avoid deprecation warnings. When they later edit the code and get rid of the `@compat`, they can take the opportunity to exploit some of the nice new features of `Mmap.mmap`.

I'd prefer to do this via Compat, if you're willing to tackle that. If you're hesitant, though, I won't insist.

timholy:
Interestingly, I can reach those files fine, so it seems like a DNS problem for AppVeyor.

rened:
Fixed the PR, now relies on https://github.com/JuliaLang/Compat.jl/pull/123

timholy:
You are a god.

tkelman:
The opensuse mirrors are occasionally flaky, but with mmap functionality, please don't tag until it's been tested on Windows.

tkelman:
Oh, you already did in https://github.com/JuliaLang/METADATA.jl/pull/3010. Strikes me as really risky if all it's fixing is a deprecation warning.

timholy:
It passed tests in https://github.com/JuliaLang/Compat.jl/pull/123, but now I see that didn't test Windows. Hmm, sorry about that. And the DNS problems at AppVeyor seem to be continuing, so we still don't know the answer.

timholy:
Can someone who has a local install on Windows already test? We could revert https://github.com/JuliaLang/METADATA.jl/pull/3010 if need be.

tkelman:
I'm waiting on Pkg.update to do its thing.

tkelman:
Alright, apologies for my paranoia, it looks like HDF5 0.5.2 and JLD 0.5.1 are passing tests on win64 Julia 0.3. I can see that it was good to get https://github.com/JuliaLang/JLD.jl/pull/8 tagged for 0.4.

rened:
cool! thanks for keeping a close eye on the windows side of things, much appreciated!

PR #249, opened 2015-08-05
Generalize read(::DataFile, name::T...) to T=ByteString

timholy:
This avoids an ambiguity warning that (I think) appears only when precompiling.
But it seems like the right thing to do anyway, since the definition in
plain.jl also allows ByteString.

### DIFF
--- a/src/datafile.jl
+++ b/src/datafile.jl
@@ -28,7 +28,7 @@ macro write(fid, sym)
 end
 
 # Read a list of variables, read(parent, "A", "B", "x", ...)
-read(parent::DataFile, name::ASCIIString...) =
+read(parent::DataFile, name::ByteString...) =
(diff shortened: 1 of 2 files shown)


simonster:
Not sure what AppVeyor's problem is but this seems good to me. HDF5 definitely supports UTF8 in names.

Should we split this module out into a separate package? Or should I copy/paste these methods to JLD2 (which has no HDF5 dependency)?

timholy:
I'm fine with either choice.

Also, FYI: see the not-yet-(re)announced [FileIO.jl](https://github.com/JuliaIO/FileIO.jl). I've got a branch that ports this to FileIO. It was very seamless (only a few lines), but I haven't pushed yet because I was going to check that I could handle `ImageMagick` inside of `Images` the same way.

PR #252, opened 2015-08-11
remove redundant register_blosc() call

stevengj:
I noticed that `register_blosc()` was being called twice, once in the body of the module (wrong) and once in `__init__()` (right).  This removes the former.


timholy:
Thanks!

PR #263, opened 2015-09-19
Update README.md

jpfairbanks:
update `Sys.DL_LOAD_PATH` to `Libdl.DL_LOAD_PATH`

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,10 @@ how to do it:
   Julia to crash when freeing memory allocated by libhdf5.
 
 If you've installed the library but discover that Julia is not finding
-it, you can add the path to Julia's `Sys.DL_LOAD_PATH` variable, e.g.,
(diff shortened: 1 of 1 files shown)


jpfairbanks:
The line "Inserting this command into your `.juliarc.jl` file will cause this to
 happen automatically each time you start Julia." implies that this is necessary for the library to work. Is that true?

ggggggggg:
In the case where `using HDF5` doesn't work because it can't find the library, and adding something to `DL_LOAD_PATH` helps, putting that in the `.juliarc.jl` file would be convenient.  I used to use this workaround, but no longer need to, probably thanks to `Homebrew.jl`.

timholy:
On julia 0.3 it's still `Sys` unless you're `using Compat`. Can you modify this to acknowledge that fact?

timholy:
At this point 0.3 is pretty old so let's just go with it.

PR #270, opened 2015-10-11
'package' in place of 'module' and 'repository'

catawbasam:
It isn't entirely clear from the current text that `JLD.jl` lives in a separate package.


timholy:
Thanks!

PR #277, opened 2015-12-07
fix readme appveyor badge link

tkelman:
~~Apparently I lost commit access here?~~


timholy:
If this is still correct & relevant, please do merge.

tkelman:
Yup. AppVeyor doesn't really have organization accounts last I checked. So the url is under whoever's account turned it on.

PR #280, opened 2016-01-10
convenience functions for read and writing attributes

bjarthur:
i just hoisted the lower-level means of accessing a dataset's attributes to the level of h5read/write.  complete with docs and tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,16 @@ h5open("test.h5", "w") do file
 end
 ```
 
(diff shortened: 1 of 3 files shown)


timholy:
Note failure on julia 0.3. (Probably need Compat.jl for the Dict syntax.)

bjarthur:
fixed.

timholy:
Thanks!

PR #284, opened 2016-02-05
Update README.md

JoshChristie:
Remove deprecated functions (h5writeattr, h5readattr) from docs.


timholy:
Actually, it's just that the docs reflect `master`, but the latest package-manager release (0.5.7) doesn't  include these functions.

I'll tag a new release.

timholy:
0.5.8 is now out.

PR #303, opened 2016-05-15
tweak Compat.String usage

StefanKarpinski:


src/plain.jl
```diff
@@ -269,17 +269,17 @@ type UTF8Char<:CharType
     c::UInt8
 end
 length(c::UTF8Char) = 1
-chartype(::Type{Compat.ASCIIString}) = ASCIIChar
```
yuyichao: This is clearly API breaking

timholy:
I'm reopening because I missed the build script, at least.

yuyichao:
I don't think there's anything missing in the build script. The depwarn there all comes from BinDeps (which is fixed but not tagged so you'll see them on the CI).

yuyichao:
The only change this PR made in the build script is a no op and it doesn't fix the depwarn in the build script https://travis-ci.org/JuliaIO/HDF5.jl/jobs/130395576#L527

src/plain.jl
timholy: @StefanKarpinski, if you think this works you might want to try running the JLD and JLDArchives tests.

The problem @yuyichao is noting is that these are still separate types on julia-0.3 and julia-0.4. Changing it to `Type{String}` seems like it will only match the non-leaftype `ByteString` and throw a `MethodError` for a concretely-type `ASCIIString` or `UTF8String`.

timholy:
Oh, right. @StefanKarpinski, if there are any changes you think should be made, by all means feel free to revise this. Or close it if you think it's OK as-is, given comments above.

PR #311, opened 2016-06-16
fix deprecations

abhijithch:
Fixes most of the deprecation warnings, requires latest `Compat` to be tagged.


timholy:
A huge help, many thanks!

PR #314, opened 2016-07-06
Tell Homebrew.jl to tap homebrew/science if necessary

iamed2:
HDF5 is no longer in `staticfloat/juliadeps`.
When Homebrew.jl installs its own brew (i.e. on CI) it won't have tapped `homebrew/science` so it won't find hdf5.
This change will let it find hdf5 in `homebrew/science`.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,5 +2,5 @@ julia 0.3.4
 BinDeps
 Blosc
 Compat 0.8.0
-@osx Homebrew
+@osx Homebrew 0.3.1
 @windows WinRPM
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -18,7 +18,7 @@ end
 @osx_only begin
     using Homebrew
     hdf5 = library_dependency("libhdf5")
-    provides(Homebrew.HB, ["hdf5", "homebrew/science/hdf5"], hdf5, os = :Darwin )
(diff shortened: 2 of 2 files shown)


iamed2:
Never mind, that's not how Homebrew.jl works. That tap would have to be manually tapped first for this to function correctly.

simonster:
@staticfloat Perhaps you know the right solution here?

staticfloat:
You should just list `"homebrew/science/hdf5"`, not `["hdf5", "homebrew/science/hdf5"]`.  By manually specifying the tap, it will automatically tap `homebrew/science` for you.

staticfloat:
Hmmmm.  Well that should have worked.  Some code path isn't behaving properly.

iamed2:
Looks like Homebrew.jl used to tap when `info` was called, and now it does not. With the BinDeps integration, `info` is used to determine whether a package is available. Since the tap isn't tapped, that fails, so it never tries to install.

staticfloat:
Fixed on the latest `master`, a new release of `Homebrew.jl` will be out in a day or so.  Thank you for your bug reports!

iamed2:
Thank you @staticfloat !

iamed2:
Rebuild and merge? This should work now.

PR #325, opened 2016-09-09
Use isassigned(::Array) instead of isdefined(::Array)

yuyichao:
For https://github.com/JuliaLang/julia/pull/18362


yuyichao:
.......Forgot to change the PR title after pushing the second commit......

PR #331, opened 2016-10-01
Update example code in README.md

r-chris:
Without an explicit `Array{Int64}` instantiation the following error is produced (in julia 0.5.0): `ERROR: MethodError: no method matching write(::HDF5.HDF5File, ::String, ::Array{UnitRange{Int64},1})`

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ end
 Convenience functions for attributes attached to datasets are also provided:
 
 ```julia
-  A=[1:10]
+  A=Array{Int64}(1:10)
   h5write("bar.h5", "foo", A)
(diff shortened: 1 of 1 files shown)


andyferris:
The motivation behind the PR is pretty uncontroversial... I'm a little surprised there hasn't been a single comment in weeks.

andyferris:
bump

README.md
```diff
@@ -115,7 +115,7 @@ end
 Convenience functions for attributes attached to datasets are also provided:
 
 ```julia
-  A=[1:10]
+  A=Array{Int64}(1:10)
```
musm: agreed

musm: Better to write it as `Vector{Int64}(1:10)`

musm:
Closer by https://github.com/JuliaIO/HDF5.jl/pull/373

Issue #338, opened 2016-10-21
Easy access all datasets


OmriRoames:
I find it very handy to get all datasets and/or their paths without reading the entire file. It gives the possibility to iterate through all the datasets and load them one at a time without blowing up the memory. 

Would a PR with that functionality be welcome?

PR #339, opened 2016-10-24
Get datasets

OmriRoames:
A utility method for getting easy access to all datasets in the file to allow iteration over all datasets (issue #338 338).


src/plain.jl
```diff
@@ -2300,6 +2300,29 @@ function hiding_errors(f)
...
 end
 
+"""
+    get_datasets(file::HDF5File) -> datasets::Vector{HDF5Dataset}
+
+get all datasets in the file without loading the data
+"""
+function get_datasets(file::HDF5File)
+    list = Vector{HDF5Dataset}()
+    get_datasets!(file, list)
```
andyferris: It's typical Julia convention to have the mutated variable appear first (e.g. `get_datasets!(list, file)`).

(not that it matters for this PR since the function is not exported)

andyferris:
bump

src/plain.jl
```diff
@@ -2300,6 +2300,29 @@ function hiding_errors(f)
     res
 end
 
+"""
+    get_datasets(file::HDF5File) -> datasets::Vector{HDF5Dataset}
+
+get all datasets in the file without loading the data
+"""
+function get_datasets(file::HDF5File)
+    list = Vector{HDF5Dataset}()
```
musm: can we write this as `Vector{HDF5Dataset}(0)` it's more explicit, thanks!

OmriRoames:
re-based and made the requested change

musm:
fixed the test based upon my suggestion

musm:
@timholy and @simonster  does this look good with you?

simonster:
I'm really sold on the utility/necessity of this, but the implementation looks fine.

musm:
> I'm really sold on the utility/necessity of this, but the implementation looks fine.

did you mean to say you are 'not really sold' ?

simonster:
Yes, sorry, that is what I meant.

andyferris:
I guess the utility of this is that we can easily iterate and touch all the datasets without loading them fully into memory. 

For instance you can get the size/shape/number of dimensions for every data set without loading it into memory via `map(size, get_datasets(hdf5file))`. You could also apply general transformations like filtering, converting element type, or whatever, one dataset at time to avoid blowing memory.

Minimizing memory usage is essential for us when we do cloud computing on nodes with smallish amounts of memory per node.

musm:
@simonster  do you mind either approving this or closing this, depending on whether you think it's needed or not? thanks

musm:
@ggggggggg do you have any opinions here.
Personally since two people have found this useful this could warrant inclusion ?

ggggggggg:
It looks ok to me. It doesn't seem necessary, but it is small, tested and documented. I'd be in favor of it in an effort to increase contributors if nothing else.

musm:
Planning on merging this sans objections soon

PR #348, opened 2016-12-19
Add alias for support of IBM Power architecture

AndyGreenwell:
Execution of `apt-get install hdf5-tools` on a Power system running Ubuntu 16.04 does not seem to install un-versioned .so files.  The edit in this commit allows for `Pkg.build("HDF5")` to be executed successfully, and for HDF5.jl to work subsequently.


tkelman:
nightly failure is unrelated

PR #350, opened 2016-12-27
Added collect() to make a ReshapedArray into an Array.

beekor:
I'm new at this - suggestions welcome and appreciated.


timholy:
Thanks!

PR #352, opened 2016-12-30
don't access `.data` field of `String`

JeffBezanson:


src/plain.jl
```diff
@@ -1106,10 +1106,10 @@ datatype(dset::HDF5Attribute) = HDF5Datatype(h5a_get_type(checkvalid(dset).id),
 datatype{T<:HDF5BitsKind}(x::T) = HDF5Datatype(hdf5_type_id(T), false)
 datatype{T<:HDF5BitsKind}(::Type{T}) = HDF5Datatype(hdf5_type_id(T), false)
 datatype{T<:HDF5BitsKind}(A::Array{T}) = HDF5Datatype(hdf5_type_id(T), false)
-function datatype{S<:String}(str::S)
```
tkelman: is the parameterization needed on 0.4 where this is Compat.String?

PR #359, opened 2017-01-19
Rm plain.jl

musm:


tkelman:
much better this way, but could use a rebase to just the last 3 commits

musm:
anything else to do here?

tkelman:
Guess not. All existing PR's would have to be rebased, but maybe that's worth the cleanup?

musm:
True. I  took a look and there are maybe ~3 PRs that would require a rebase, the others are trivial. It looks like many of the PRs are stale from October and before. Your call on what is best.

andyferris:
We can rebase our ones from September/October - go ahead and merge. :)

By the way, they didn't become stale - those PRs were simply ignored by the maintainers... :(

musm:
I can help rebase some of the PRs if that helps.

PR #363, opened 2017-02-04
Tweak .gitignore

musm:
Just moved the gitignore specific to deps to the deps gitignore. also removed the gitignore on the .mat files since we don't have any .mat files in the repo


musm:
Does the current .gitignore need
`*~
`
that seems superfluous? 
what about the `deps/downloads` folder ?

PR #361, opened 2017-01-28
Drop 0.4

musm:
Will help to simplify string handling. Since 0.6 is soon looming might as well.

Close https://github.com/JuliaIO/HDF5.jl/issues/356

Helps with https://github.com/JuliaIO/HDF5.jl/pull/340  cc @andyferris

PR #364, opened 2017-02-04
Use Base.Test for testing

musm:
Replaces `@assert` with `@test`


musm:
Needs https://github.com/JuliaIO/HDF5.jl/pull/361, since using testsets

Issue #347, opened 2016-12-12
Reading in HDF5Compound types


tamasgal:
I am trying to read data from an HDF5 file, which was created by pytables and uses HDF5Compound types. I stumbled upon a few posts where I read that read reading compound types works well but could not find a single example. Well, I found one reading in an array with complex numbers using `reinterpret` but I did not manage to make it work with my data.

The reading procedure is clear:

```
hits = h5open("the_dataset.h5", "r") do file
    read(file, "/hits")
end
```

And this shows already:

```HDF5.HDF5Compound(UInt8[0x1d,0x00,0x00,0x00,0x00,0x00,0x00,0xf8,0x7f,0x00  …  0x07,0xe8,0xfa,0x02,0x1c,0x00,0xae,0x0c,0x00,0x00],Type[UInt8,Float64,Float64,Float64,UInt32,UInt32,UInt32,Float64,Float64,Float64,Int32,Int32,UInt8,UInt8,UInt32],String["channel_id","dir_x","dir_y","dir_z","dom_id","id","pmt_id","pos_x","pos_y","pos_z","t0","time","tot","triggered","event_id"],UInt64[0x0000000000000000,0x0000000000000001,0x0000000000000009,0x0000000000000011,0x0000000000000019,0x000000000000001d,0x0000000000000021,0x0000000000000025,0x000000000000002d,0x0000000000000035,0x000000000000003d,0x0000000000000041,0x0000000000000045,0x0000000000000046,0x0000000000000047])```

But then, no idea how to access the arrays inside it.

This is the output of `hits.membertype`:

```
15-element Array{Type,1}:
 UInt8  
 Float64
 Float64
 Float64
 UInt32 
 UInt32 
 UInt32 
 Float64
 Float64
 Float64
 Int32  
 Int32  
 UInt8  
 UInt8  
 UInt32
```

I tried to parse the data using`reinterpret()` but I suspect that even it would read in the whole file at once, which is problematic, since the usual file size is about 20GB.

However, this is how I thought it could work:

`reinterpret(hits.membertype, hits.data)`

But I get:

```
LoadError: MethodError: no method matching reinterpret(::Array{Type,1}, ::Array{UInt8,1})
Closest candidates are:
  reinterpret{T,S}(!Matched::Type{T}, ::Array{S,1}) at array.jl:73
  reinterpret{T,S}(!Matched::Type{T}, ::Array{S,N}) at array.jl:79
[...]

PR #365, opened 2017-02-06
(WIP) Read in compound dsets with strings using IOBuffer

grahamrow:
Hi All,

I was keen on getting something working with Strings in compound datasets. My original thought was to use a fixed length vector from the FixedSizeArrays.jl package, but a quicker fix was to run the data through an IOBuffer with a bit of extra handling for string types. I opted to store the member lengths for the fields in addition to the offsets in the HDF5Compound Type, which allows for a fairly straightforward conversion to Julia strings. This procedure is a bit redundant since the UInt8 buffer gets filled and then immediately reinterpreted through the IOBuffer, so it would be better to have a type into which the data might be read directly.

For now, I have changed the return type of the buffer to an Array{Any} in order to accommodate mixed types. I will probably utilize a DataFrame for my own code, but don't want to force that dependency on anyone else.

Please let me know your thoughts and I can potentially move onto to adding a unit test. I'm VERY new to Julia so please forgive any gross misunderstandings of Julian conventions.

Thanks,
Graham

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -426,10 +426,11 @@ hash(x::HDF5ReferenceObj, h::UInt) = hash(x.r, h)
 # Compound types
 # These are "raw" and not mapped to any Julia type
 type HDF5Compound
-    data::Array{UInt8}
+    data::Array
     membertype::Vector{Type}
     membername::Vector{Compat.ASCIIString}
     memberoffset::Vector{UInt}
+    membersize::Vector{UInt8}
 end
 
 # Opaque types
@@ -1378,26 +1379,43 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
     h5object(obj_id, parent)
 end
 
+# Helper for reading compound types
+function read_row(io::IO, membertype, membersize)
+    row = []
+    for (dtype, dsize) in zip(membertype, membersize)
+        if dtype == String
+            push!(row, strip(String(read(io, UInt8, dsize)),'\0'))
+        else
+            push!(row, read(io, dtype))
+        end
+    end
+    return row
+end
+
 # Read compound type
 function read(obj::HDF5Dataset, ::Union{Type{Array{HDF5Compound}},Type{HDF5Compound}})
     t = datatype(obj)
-    local sz = 0; local n; local membername; local membertype; local memberoffset; local memberfiletype
+    local sz = 0; local n;
+    local membername; local membertype;
+    local memberoffset; local memberfiletype; local membersize;
     try
         n = h5t_get_nmembers(t.id)
         memberfiletype = Vector{HDF5Datatype}(n)
         membertype = Vector{Type}(n)
         membername = Vector{Compat.ASCIIString}(n)
         memberoffset = Vector{UInt64}(n)
+        membersize = Vector{UInt8}(n)
         for i = 1:n
             filetype = HDF5Datatype(h5t_get_member_type(t.id, i-1))
             T = hdf5_to_julia_eltype(filetype)
-            if T <: String
-                error("Not yet supported")  # need to handle the vlen issues
-    #             T = Ptr{UInt8}
-            end
+    #         if T <: String
+    #             error("Not yet supported")  # need to handle the vlen issues
+    # #             T = Ptr{UInt8}
+    #         end
             memberfiletype[i] = filetype
             membertype[i] = T
             memberoffset[i] = sz
+            membersize[i] = sizeof(filetype)
             sz += sizeof(filetype)
             membername[i] = h5t_get_member_name(t.id, i-1)
         end
@@ -1412,7 +1430,15 @@ function read(obj::HDF5Dataset, ::Union{Type{Array{HDF5Compound}},Type{HDF5Compo
     # Read the raw data
     buf = Vector{UInt8}(length(obj)*sz)
     h5d_read(obj.id, memtype_id, H5S_ALL, H5S_ALL, H5P_DEFAULT, buf)
-    HDF5Compound(buf, membertype, membername, memberoffset)
+
+    # Convert to the appropriate data format using iobuffer
+    iobuff = IOBuffer(buf)
+    data = []
+    while !eof(iobuff)
+        push!(data, read_row(iobuff, membertype, membersize))
+    end
+
+    HDF5Compound(data, membertype, membername, memberoffset, membersize)
 end
 
 # Read OPAQUE datasets and attributes


grahamrow:
Shameless pings: @simonster you had some helpful commentary about compound types in #347, @musm I saw you had some other PRs open for string-related phenomena.

Any thoughts on how to proceed?

src/HDF5.jl
```diff
@@ -1378,26 +1379,43 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
     h5object(obj_id, parent)
 end
 
+# Helper for reading compound types
+function read_row(io::IO, membertype, membersize)
+    row = []
+    for (dtype, dsize) in zip(membertype, membersize)
+        if dtype == String
+            push!(row, strip(String(read(io, UInt8, dsize)),'\0'))
```
grahamrow: Not sure how universal this \0 packing is — numpy sticks these chars into its fixed length strings.

grahamrow:
Added unit test. Now working with FixedArray types. Working in Julia 0.4/0.5 — 0.6 is failing due to other reasons.

src/HDF5.jl
```diff
@@ -426,10 +426,11 @@ hash(x::HDF5ReferenceObj, h::UInt) = hash(x.r, h)
 # Compound types
 # These are "raw" and not mapped to any Julia type
 type HDF5Compound
-    data::Array{UInt8}
+    data::Array
```
blakejohnson: This should be `Array{Any}` to match your intention of storing appropriately typed data in here.

src/HDF5.jl
```diff
@@ -1378,26 +1379,46 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
...
             filetype = HDF5Datatype(h5t_get_member_type(t.id, i-1))
             T = hdf5_to_julia_eltype(filetype)
-            if T <: String
-                error("Not yet supported")  # need to handle the vlen issues
-    #             T = Ptr{UInt8}
-            end
+    #         if T <: String
+    #             error("Not yet supported")  # need to handle the vlen issues
+    # #             T = Ptr{UInt8}
+    #         end
```
blakejohnson: You can just delete this block of code.

src/HDF5.jl
```diff
@@ -1378,26 +1379,46 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
...
 end
 
+# Helper for reading compound types
+function read_row(io::IO, membertype, membersize)
+    row = []
+    for (dtype, dsize) in zip(membertype, membersize)
+        if dtype == String
+            push!(row, strip(String(read(io, UInt8, dsize)),'\0'))
+        # This is pretty awkward...
+        elseif dtype.name == HDF5.FixedArray.name
```
blakejohnson: Maybe you want `dtype <: HDF5.FixedArray` ?

src/HDF5.jl
```diff
@@ -1378,26 +1379,46 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
     h5object(obj_id, parent)
 end
 
+# Helper for reading compound types
+function read_row(io::IO, membertype, membersize)
+    row = []
+    for (dtype, dsize) in zip(membertype, membersize)
+        if dtype == String
```
blakejohnson: `dtype === String`

src/HDF5.jl
```diff
@@ -1378,26 +1379,46 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
...
+            push!(row, strip(String(read(io, UInt8, dsize)),'\0'))
+        # This is pretty awkward...
+        elseif dtype.name == HDF5.FixedArray.name
+            push!(row, read(io, eltype(dtype), size(dtype)[1]))
```
blakejohnson: There are multi-dimensional `dtypes`? (I'm looking at `size(dtype)[1]`)

src/HDF5.jl
musm: Just to double check you want an Array here and not a Vector?

blakejohnson: Good point, `Vector{Any}` would be better.

grahamrow:
Thanks for the review. Updated to use Vector{Any} and no longer expose any of the offset/type/size information since we have explicitly executed the type conversion.

src/HDF5.jl
```diff
@@ -1378,12 +1378,11 @@ end
 
 # Helper for reading compound types
 function read_row(io::IO, membertype, membersize)
-    row = []
+    row = Vector{Any}()
```
musm: Also can use `Any[]`, I think tkelman mentioned that there have been some discussions of deprecated this type of constructor. May be best to use `Any[]` to be safe

src/HDF5.jl
```diff
@@ -1378,26 +1376,41 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
     h5object(obj_id, parent)
 end
 
+# Helper for reading compound types
+function read_row(io::IO, membertype, membersize)
+    row = Vector{Any}()
```
musm: I think `@tkelman` mentioned that there have been some discussions of deprecated this type of constructor. May be best to use `Any[]` to be safe

grahamrow: Thanks! The latter is a bit easier on the eyes at any rate.

blakejohnson: @musm I think you are referring to the deprecation `Array(T, size)` in favor of `Array{T}(size)` in julia 0.6. `Vector{Any}()` should be fine, though `Any[]` is more concise.

src/HDF5.jl
```diff
@@ -1412,7 +1425,14 @@ function read(obj::HDF5Dataset, ::Union{Type{Array{HDF5Compound}},Type{HDF5Compo
     # Read the raw data
     buf = Vector{UInt8}(length(obj)*sz)
     h5d_read(obj.id, memtype_id, H5S_ALL, H5S_ALL, H5P_DEFAULT, buf)
-    HDF5Compound(buf, membertype, membername, memberoffset)
+
+    # Convert to the appropriate data format using iobuffer
+    iobuff = IOBuffer(buf)
+    data = []
```
musm: should this also be `Any[]` ?

blakejohnson: `[]` is equivalent to `Any[]`, but as you say, it doesn't hurt to be explicit.

src/HDF5.jl
```diff
@@ -426,10 +426,8 @@ hash(x::HDF5ReferenceObj, h::UInt) = hash(x.r, h)
 # Compound types
 # These are "raw" and not mapped to any Julia type
```
blakejohnson: I guess we need to update the comment now that the data is mapped to Julia types.

src/HDF5.jl
```diff
@@ -1378,26 +1376,41 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
     h5object(obj_id, parent)
 end
 
+# Helper for reading compound types
+function read_row(io::IO, membertype, membersize)
+    row = Any[]
+    for (dtype, dsize) in zip(membertype, membersize)
+        if dtype === String
+            push!(row, strip(String(read(io, UInt8, dsize)),'\0'))
```
blakejohnson: There is a method `unpad` defined elsewhere in this file that deals with different string padding modes. Can we use that in combination with `h5t_get_strpad` to handle more general cases?

src/HDF5.jl
```diff
@@ -1378,26 +1376,41 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
...
+        elseif dtype <: HDF5.FixedArray
+            push!(row, read(io, eltype(dtype), size(dtype)[1]))
```
blakejohnson: I still don't understand the `[1]` after `size(dtype)`. The definition:
```
size{T,D}(::Type{FixedArray{T,D}}) = D
```
ensures that you can only get back a single value when taking size of a `FixedArray`, so the extra indexing is unnecessary.

src/HDF5.jl
```diff
@@ -1412,7 +1425,14 @@ function read(obj::HDF5Dataset, ::Union{Type{Array{HDF5Compound}},Type{HDF5Compo
...
     h5d_read(obj.id, memtype_id, H5S_ALL, H5S_ALL, H5P_DEFAULT, buf)
-    HDF5Compound(buf, membertype, membername, memberoffset)
+
+    # Convert to the appropriate data format using iobuffer
+    iobuff = IOBuffer(buf)
+    data = Any[]
+    while !eof(iobuff)
+        push!(data, read_row(iobuff, membertype, membersize))
+    end
+    HDF5Compound(data, membername)
```
blakejohnson: I think the return type should depend upon whether the input storage format is `Array{HDF5Compound}` or just `HDF5Compound`, where we return `Vector{HDF5Compound}` in the former case and a single `HDF5Compound` element in the latter.

src/HDF5.jl
blakejohnson: Ah, I see now. `D` can be a tuple so the indexing is necessary.

src/HDF5.jl
```diff
@@ -1378,26 +1376,41 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
...
+    for (dtype, dsize) in zip(membertype, membersize)
+        if dtype === String
+            push!(row, strip(String(read(io, UInt8, dsize)),'\0'))
+        elseif dtype <: HDF5.FixedArray
+            push!(row, read(io, eltype(dtype), size(dtype)[1]))
+        else
+            push!(row, read(io, dtype))
+        end
+    end
+    return row
```
blakejohnson: It would be nice if this whole method could use the set of `read` methods defined immediately above this in the source, but it is not immediately obvious how to do that. At the very least, then, the last block should check that `dtype <: HDF5BitsKind` and then have a final `else` block throw an error.

blakejohnson:
From my perspective, this is ready to go. The PR deserves a new title, since what this actually does is improve handling of `HDF5Compound` types to convert `String`, `HDF5BitsKind`, and `FixedArray` elements to appropriate native types, rather than requiring the user to do so manually.

This significantly changes the `HDF5Compound` type to effectively be a poor man's named tuple. This PR probably will break any code relying on the current layout of this type.

blakejohnson:
Bump? Note that the build error is unrelated.

Issue #324, opened 2016-08-26
Pkg.build fails with BinDeps.SynchronousStepCollection


GlenHertz:
On Julia 0.5.0-rc3 on RHEL6.4 I tried to `Pkg.add("HDF5")` and it fails with build errors:

```
LoadError: File AbstractString["libhdf5.la"] was not created successfully (Tried to run BinDeps.SynchronousStepCollection(`...
...
while loading /remote/us01home32/hertz/.julia/v0.5/HDF5/deps/build.jl, in expression starting on line 27
```

The line in question in build.jl:27 is:

`@compat @BinDeps.install Dict(:libhdf5 => :libhdf5)`

From `~/.julia/v0.5/HDF5/deps/builds/libhdf5` I checked `libhdf5.la`:

```
% du src/libhdf5.la
4   src/libhdf5.la

```

So it looks like the file isn't getting built correctly (since it is only 4 kB)...even though `make` completes successfully.   

I tried `Pkg.rm("HDF5")` to start from scratch but it did not help.  Any ideas?

tkelman:
paste or gist the entire build output

ulfworsoe:
Building on CentOS I had a similar problem. The missing file `libhdf5.la` is correctly built, but `BinDeps` does not find it. Manually making a symlink to the library in a place where BinDeps looks seems to fix it. Specifically, following made the build process succeed:
- Do `Pkg.add("HDF5")`, let it run until it fails.
- Manually make symlinks 
```
.../HDF5/deps/builds/libhdf5/lib/libhdf5.la -> deps/builds/libhdf5/src/libhdf5.la
.../HDF5/deps/builds/libhdf5/libhdf5.la -> deps/builds/libhdf5/src/libhdf5.la
```
- Run `Pkg.build("HDF5")`

GlenHertz:
@tkelman I've uploaded my complete build output for HDF5 [here](https://gist.github.com/GlenHertz/830dd7c372bb698ef0781d5f4962d87f).

@ulfworsoe Thanks for the instructions, that fixes my issues.

rohitvarkey:
Just ran into this issue while trying to install `LightGraphs` on my university cluster. 

@ulfworsoe's instructions fixes it though.

sbromberger:
yup, this bit me just now as well on a Linux build. I will try the workaround.

GordStephen:
[...]

PR #367, opened 2017-02-08
Fix libtarget when building from source

carlobaldassi:
The build was broken without this change (tested on RHEL).


carlobaldassi:
(Appveyor failure is unrelated.)

simonster:
@eschnett Can you verify that this fix is correct?

musm:
for me both this pr and master work on fedora 25

iamed2:
Also occurs on Amazon Linux

musm:
Ok I'm going to merge this since it seems to be fixing the issue  for several people ref https://github.com/JuliaIO/HDF5.jl/issues/324#issuecomment-299235750

PR #371, opened 2017-03-09
Add scalar reference reading for attributes

grahamrow:
Wasn't able to read a single reference object from an attribute (only arrays) — quick fix.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1115,6 +1115,7 @@ function datatype{S<:String}(str::Array{S})
     HDF5Datatype(type_id)
 end
 datatype(R::Array{HDF5ReferenceObj}) = HDF5Datatype(H5T_STD_REF_OBJ, false)
+datatype(R::HDF5ReferenceObj) = HDF5Datatype(H5T_STD_REF_OBJ, false)
 datatype{T<:HDF5BitsKind}(A::HDF5Vlen{T}) = HDF5Datatype(h5t_vlen_create(hdf5_type_id(T)))
 function datatype{C<:CharType}(str::HDF5Vlen{C})
     type_id = h5t_copy(hdf5_type_id(C))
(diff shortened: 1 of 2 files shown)


src/HDF5.jl
```diff
@@ -1258,6 +1258,13 @@ function read{A<:FixedArray}(obj::DatasetOrAttribute, ::Type{Array{A}})
     ret
 end
 
+# Read a scalar reference objected from an attribute
+function read{T<:HDF5ReferenceObj}(obj::HDF5Attribute, ::Type{T})
+    refs = Array{T}(1)
```
musm: Would this not be better of using `Ref{T}()`?

simonster:
It would be good to have a test for this, if it's easy to make one (i.e., if we can write a reference to an attribute).

src/HDF5.jl
grahamrow: True, but more lifting would be required since `getindex` doesn't support Ref at the moment.

simonster: You can use `ref[]` to get the contents of a `Ref`.

grahamrow:
I was working on a test, for which one requires the ability to write a single reference to an attribute. It seems like one must override `a_create` and `a_write` for `HDF5ReferenceObjs`, but I'm having trouble figuring out the correct dataspace for a reference object.

grahamrow:
@simonster It seems like this is working now. Let me know if anything else needs tweaking.

musm:
I rebased this for you @grahamrow

musm:
Since a test has been added, planing on merging this Saturday

grahamrow:
Sounds good, thanks for the comments.

simonster:
I think #380 is a better way of handling this.

musm:
close in favor of https://github.com/JuliaIO/HDF5.jl/pull/380 ?

PR #376, opened 2017-03-14
Update libhdf5 source URL

GordStephen:
Update source url to match the download link on https://support.hdfgroup.org/HDF5/release/obtainsrc5110.html


simonster:
Thanks.

PR #380, opened 2017-03-17
Treat HDF5ReferenceObj similarly to other scalars

simonster:
Makes existing code for HDF5BitsKind apply to HDF5ReferenceObj, which implements the functionality in #371 (as well as support for vlens of references) with a net code deletion due to removal of duplicate code. cc @grahamrow. Apologies for not noticing that this could be done sooner.


musm:
bump?

PR #384, opened 2017-03-31
Remove plain from export

musm:
There is no `plain` function anywhere, is there a reason this is exported, cc @simonster ?


simonster:
I think this can go. It's a relic of a time before #39.

PR #386, opened 2017-04-01
Use generator syntax

musm:


musm:
Out of curiosity the try .... finally idiom is used quite often in function is it not possible to just remove these, or are there some edge cases. I'd like to remove them for perf reasons and since try blocks can't be debugged using Gallium.

simonster:
`try...finally` is used to make sure things are actually closed if an error occurs. Have you benchmarked this to see if the performance impact is actually meaningful? I wouldn't think it would matter in this context.

PR #387, opened 2017-04-02
Conversion to Int in VersionNumber

musm:
Take advantage of conversion to Int via the VersionNumber function.


musm:
unrelated download failure trying travis again

PR #388, opened 2017-04-03
Use Refs instead of Vectors

musm:


musm:
Is there an advantage to having this is global const refs as opposed to local variables in the corresponding functions?

musm:
bump

simonster:
In principle, it would be better if these were local variables, but at the moment that would incur allocation and GC overhead that I don't think we want. IIRC the `HDF5ReferenceObj` constructor is used extensively by JLD.

musm:
Thank you.  Yeah I observed a noticeable decrease in performance, when I tried local variables.

PR #394, opened 2017-04-28
Add (single reader multiple writer) swmr support

### LINKED ISSUE #392: Support SWMR for HDF5 v1.10
HDF5 offers a new single-writer / multiple-reader (SWMR) feature:

* https://support.hdfgroup.org/HDF5/docNewFeatures/NewFeaturesSwmrDocs.html
* https://support.hdfgroup.org/HDF5/Tutor/swmr.html

Besides the fact that it allows for use cases like reading (and analysing) data during data aquisition, the SWMR mode is also interesting because it increases robustness (file recovery after crashes).

It would be great if HDF5.jl would support SWMR (when HDF5 v1.10 is installed).

ggggggggg:
I took a stab at adding SWMR support. I modified `h5open` to allow 'w,swmr', 'r,swmr', and 'r+,swmr', as well as added functions. I also removed some unused boolean flags from the lower level `h5open` method.

I'm not sure how to handle the library versioning. SWMR requires libversion 1.10 or greater. I assume we'd prefer to maintain compatibility with older versions, so I made the SWMR tests run only if the version is high enough, and made `h5open` error with the 'X,swmr' options if the version is too low.

I'd like to wrap the `H5DOappend` function, but `ccall` tells me it can't find it. I think thats because in `H5DOpublic.h` the definition is wrapped in a `#ifdef __cplusplus` and I guess that wasn't enabled when the library was built?

I added docstrings to the functions I worked on.

Fixes #392


ggggggggg:
appveyor failure is unrelated

ggggggggg:
I removed `h5open(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Bool, ff::Bool,
        cpl::HDF5Properties=DEFAULT_PROPERTIES, apl::HDF5Properties=DEFAULT_PROPERTIES)` because it hard to understand once "swmr" options were added, and neither `ff` or `rd` were used.. I doubt it was widely used, but since it is breaking, I should probably deprecate it. Anybody know how to do that? There is an open issue for documenting `Base.@deprecate` so I don't think there is a good place to read about it.

ggggggggg:
Any feedback?

src/HDF5.jl
```diff
@@ -563,26 +572,27 @@ function h5open(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Bool
     HDF5File(fid, filename)
 end
 
+"""h5open(filename::AbstractString, mode::AbstractString="r")
```
musm: I think it's better to format the docstring as in base  also see typos
```julia
"""
    h5open(filename::AbstractString, mode::AbstractString="r")

Open or create an HDF5 file where `mode` is one of "r", "r+", "w", "r,swmr", or "w,swmr".
"""

```

src/HDF5.jl
```diff
@@ -1153,15 +1163,29 @@ dataspace(n::Void) = HDF5Dataspace(h5s_create(H5S_NULL))
 dataspace(sz::Dims; max_dims::Union{Dims, Tuple{}}=()) = _dataspace(sz, max_dims)
 dataspace(sz1::Int, sz2::Int, sz3::Int...; max_dims::Union{Dims, Tuple{}}=()) = _dataspace(tuple(sz1, sz2, sz3...), max_dims)
 
-# Get the array dimensions from a dataspace or a dataset
-# Returns both dims and maxdims
+
 get_dims(dspace::HDF5Dataspace) = h5s_get_simple_extent_dims(dspace.id)
+"""get_dims(dset::HDF5Dataset)
+Get the array dimensions from a dataset
```
musm: `Get the array dimensions from a dataset and return a tuple of dims and maxdims.`

src/HDF5.jl
```diff
@@ -1153,15 +1163,29 @@ dataspace(n::Void) = HDF5Dataspace(h5s_create(H5S_NULL))
...
+Returns a tuple of dims and maxdims
+"""
 get_dims(dset::HDF5Dataset) = get_dims(dataspace(checkvalid(dset)))
 
-# change the current dimensions of a dataset, new_dims fit in max_dims = get_dims(dset)[2]
-# reduction is possible, and leads to loss of truncated data
+"""set_dims!(dset::HDF5Dataset, new_dims::Dims)
+ change the current dimensions of a dataset, new_dims fit in max_dims = get_dims(dset)[2]
+reduction is possible, and leads to loss of truncated data
+"""
```
musm: `Change the current dimensions of a dataset to `new_dims` to fit in `max_dims = get_dims(dset)[2]` . Reduction is possible and leads to loss of truncated data.`

ggggggggg:
I made your changes, and generally improved the docstrings and their formatting.

src/HDF5.jl
```diff
@@ -535,25 +536,33 @@ heuristic_chunk(x) = Int[]
 
 ### High-level interface ###
 # Open or create an HDF5 file
-function h5open(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Bool, ff::Bool,
+function h5open(filename::AbstractString, mode::String,
```
musm: Is there a reason why this is `::String` and not `::AbstractString` 
also I think the docstring below needs to be here

ggggggggg: I think you're right that the signature could be AbstractString. I think the docstring should be in only one place. I put it on `function h5open end` because I thought that was the preferred way to document a function as a whole rather than just a method, but the output of `?h5open` is the same if I just move the docstring here.

ggggggggg: The manual is pretty clear that I should use AbstractString "If you define a function expecting a string argument, you should declare the type as AbstractString in order to accept any string type." https://docs.julialang.org/en/stable/manual/strings/

musm: ignore ~~also I think the docstring below needs to be here~~ I totally missed the second line of this function defining additional arguments.

I'm not 100% sure on the doc string and where it's best to place it

ggggggggg: I think you just explained better why I did it, I just put in a minimal docstring that covered the most common usage, but didn't cover all the optional arguments in any one method.

ggggggggg: I was reluctant to document the methods more fully because I feel like the pv argument should be refactored to accept `Pair`s instead of relying on alternation of keys and values.

ggggggggg:
updated (a while ago, but since last comment)

ggggggggg:
I decided that a keyword arg is cleaner for `swmr` than doubling the number of file modes. I also added a docstring for `h5open(f::Function, args...;kwargs...)`.

musm:
Rebased this
MacOS nightly
```julia
create by swmr mode, then close and open again: Test Failed
  Expression: nbigger >= 9
   Evaluated: 8 >= 9
Stacktrace:
 [1] (::#remote_test#61)(::HDF5.HDF5File) at /Users/travis/.julia/v0.7/HDF5/test/swmr.jl:83
 [2] #h5open#5(::Array{Any,1}, ::Function, ::##49#66, ::String, ::Vararg{String,N} where N) at /Users/travis/.julia/v0.7/HDF5/src/HDF5.jl:628
 [3] (::HDF5.#kw##h5open)(::Array{Any,1}, ::HDF5.#h5open, ::Function, ::String, ::String) at ./<missing>:0
 [4] macro expansion at /Users/travis/.julia/v0.7/HDF5/test/swmr.jl:108 [inlined]
 [5] macro expansion at ./test.jl:856 [inlined]
 [6] macro expansion at /Users/travis/.julia/v0.7/HDF5/test/swmr.jl:103 [inlined]
 [7] macro expansion at ./test.jl:856 [inlined]
 [8] macro expansion at /Users/travis/.julia/v0.7/HDF5/test/swmr.jl:15 [inlined]
 [9] anonymous at ./<missing>:?
```

ggggggggg:
I added some explanatory text to help others understand the test, and I added a delay after notification of a write before attempting to read. The delay seems to have helped quite a bit. I make 10 write, and look for the data size to increase on 10 reads. Previously I only saw 9 data size increases on a regular basis, but with the delay I now see 10 regularly. So hopefully this is more reliable.

musm:
> I decided that a keyword arg is cleaner for swmr than doubling the number of file modes. I also added a docstring for h5open(f::Function, args...;kwargs...).

agreed

ggggggggg:
I got the same test failure again. Clearly it is not reliable enough yet.Any advice would be appreciated. I guess I could just keep calling refresh until I see the dataset get larger, or I could not be so stringent about seeing it get larger after every write.
```
  Expression: nbigger == 10
   Evaluated: 9 == 10
```

oschulz:
Fixes #392.

musm:
Is the sleep time too stringent for CI, sleep(0.01) ?

test/swmr.jl
```diff
@@ -0,0 +1,120 @@
...
+  addprocs(1)
+end
+@everywhere using HDF5
+using Base.Test
+
+
+
+
+@testset "swmr" begin
+fname = tempname()
```
musm: It would be good to delete this after testing is complete.

test/swmr.jl
```diff
@@ -0,0 +1,120 @@
...
+if HDF5.libversion >= v"1.10.0"
+
+if nprocs() == 1
+  addprocs(1)
+end
+@everywhere using HDF5
+using Base.Test
+
+
+
```
musm: extra white space?

ggggggggg:
travis failure is only on osx nightly, and is unrelated to this

musm:
I'm wondering how to get updated libhdf5 in WinRPM  since it's version 1.8  ....

ggggggggg:
I think the tests are under control now. I made it just keep checking and waiting longer until it sees the new data. I triggered CI a few times, and it's worked 5 times in a row. The most recent passed fully (except coveralls which I think is mad that I added docs), and the previous runs passed on all but the nightly (and the failure was unrelated).

I figured out how to deprecate the `h5open` signature with many Bools, so now I think I've accomplished everything I aimed for. Ready for more review or merge.

ggggggggg:
I removed the deprecation of an `h5open` signature. I realized it would need some coordination with `JLD` and it's just simpler to keep the issues separate.

Appveyor failure is only on nightly and looks unrelated.

musm:
cc @timholy

musm:
well we should first try to get the HDF5 binaries on WinRPM updated to versions that support swmr so that can also be tested

test/swmr.jl
```diff
@@ -0,0 +1,119 @@
+# following https://support.hdfgroup.org/HDF5/doc/RM/RM_H5F.html#File-StartSwmrWrite
+# and https://support.hdfgroup.org/HDF5/docNewFeatures/SWMR/HDF5_SWMR_Users_Guide.pdf
+
+if HDF5.libversion >= v"1.10.0"
+
+if nprocs() == 1
+  addprocs(1)
```
musm: minor but these should all be 4 space indented

ggggggggg:
Thanks for the help! Glad to see this merged.

musm:
thank you very much @ggggggggg  for taking a big role in improving this package !

oschulz:
Thanks for implementing SWMR support, @ggggggggg!

musm:
Hi @ggggggggg  do you have any ideas on the failures on mac here
https://travis-ci.org/JuliaIO/HDF5.jl/jobs/296102381
seems related to SWMR

ggggggggg:
There are two failures
1) failing to create the file with `Error creating file /var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/juliaeAhUTJ` and `#003: H5FDsec2.c line 940 in H5FD_sec2_lock(): unable to lock file, errno = 35, error message = 'Resource temporarily unavailable'` at
https://github.com/JuliaIO/HDF5.jl/blob/9a916c4661fa6ad1bfa6c894a2d533a6d6063e52/test/swmr.jl#L108
This one I don't understand at all, it seems like it is simply failing to create the file. I could wrap the relevant code in `retry`? Should I not do it in the temp directory?

2) failing to write to an hdf5 dataset with `MethodError: no method matching setindex!(::HDF5.HDF5Dataset, ::Array{Int64,1}, ::Array{Int64,1})` at
https://github.com/JuliaIO/HDF5.jl/blob/9a916c4661fa6ad1bfa6c894a2d533a6d6063e52/test/swmr.jl#L42-L44
#443 changed line 42 from `inds = (1:10) + (i - 1) * 10` to `inds = (1:10) .+ (i - 1) * 10`. Previously `inds` was a `UnitRange`, now it is a `Vector`. It's not clear to me why this works on any platforms given that `setindex!(::HDF5.HDF5Dataset, ::Array{Int64,1}, ::Array{Int64,1})` doesn't exist.

musm:
Thanks, I don't have access to a Mac so I am not sure why it would fail to create the file? Maybe we could try a PR with a couple of retrys to see if it works?

Submitted a PR with your suggestions for point 2. https://github.com/JuliaIO/HDF5.jl/pull/448

musm:
Maybe we should add setindex with Array{Int64,1} to match base?

ggggggggg:
Maybe, I think it might be really slow given the overhead on hdf5 calls.

Issue #366, opened 2017-02-06
Can't dump closed file handle


ggggggggg:
It seems like this should work

```
julia> h5 = h5open(tempname(),"w")
HDF5 data file: /var/folders/_0/25kp6h7x25v6vyjv2yjlcnkm000wrm/T/juliaBebRVv

julia> close(h5)

julia> dump(h5)
HDF5-DIAG: Error detected in HDF5 (1.8.18) thread 0:
  #000: H5Gdeprec.c line 844 in H5Gget_num_objs(): not a location ID
    major: Invalid arguments to routine
    minor: Inappropriate type
  #001: H5Gloc.c line 253 in H5G_loc(): invalid object ID
    major: Invalid arguments to routine
    minor: Bad value
ERROR: Error getting group length
 in h5g_get_num_objs at /Users/user/.julia/v0.5/HDF5/src/HDF5.jl:2104 [inlined]
 in length(::HDF5.HDF5File) at /Users/user/.julia/v0.5/HDF5/src/HDF5.jl:959
 in dump(::IOContext{Base.TTY}, ::HDF5.HDF5File, ::Int64, ::String) at /Users/user/.julia/v0.5/HDF5/src/HDF5.jl:1081
 in #dump#323 at ./show.jl:1226 [inlined]
 in (::Base.#kw##dump)(::Array{Any,1}, ::Base.#dump, ::IOContext{Base.TTY}, ::HDF5.HDF5File) at ./<missing>:0
 in dump(::HDF5.HDF5File) at ./show.jl:1227
```

musm:
This work for me. Try checking out master and try again.

ggggggggg:
I still see the same behavior on Master. The following is after I ran `Pkg.checkout("HDF5") in a previous julia terminal.

```julia
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.0 (2016-09-19 18:14 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin13.4.0

julia> using HDF5

julia> f = h5open(tempname(),"w")
[...]

PR #395, opened 2017-04-30
Remove dump and update show methods

musm:
suggestions welcome  ref https://github.com/JuliaIO/HDF5.jl/issues/366

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -6,7 +6,7 @@ using Compat
 using Compat: unsafe_convert, String
 
 ## Add methods to...
-import Base: ==, close, convert, done, dump, eltype, endof, flush, getindex,
+import Base: ==, close, convert, done, eltype, endof, flush, getindex,
              isempty, isvalid, length, names, ndims, next, parent, read,
              setindex!, show, size, sizeof, start, write
 
@@ -290,6 +290,8 @@ hdf5_type_id{C<:CharType}(::Type{C})  = H5T_C_S1
 # that occur when passing a freshly-created file to some other
 # application).
 
+const indent_width = 4
+
 # This defines an "unformatted" HDF5 data file. Formatted files are defined in separate modules.
 type HDF5File <: DataFile
     id::Hid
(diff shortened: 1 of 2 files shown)


yuyichao:
I think this needs an doc update and possibly deprecation too (though I guess there shouldn't be an error with the old use of dump either)

ggggggggg:
Looks pretty good, maybe a few tests?

src/HDF5.jl
```diff
@@ -304,7 +304,22 @@ type HDF5File <: DataFile
...
 convert(::Type{Hid}, f::HDF5File) = f.id
-show(io::IO, fid::HDF5File) = isvalid(fid) ? print(io, "HDF5 data file: ", fid.filename) : print(io, "Closed HFD5 data file: ", fid.filename)
+function show(io::IO, fid::HDF5File)
+    if isvalid(fid)
+        if get(io, :compact, false)
+            print(io, "HDF5 data file: ", fid.filename)
+        else
+            print(io, "HDF5 data file: ", fid.filename, " (length ", length(fid), ")")
+            for k in names(fid)
+                print(io, "\n  ")
```
musm: perhaps it's not useful to print the names ?

musm:
How do you deal with the deprecation in this case?  dump is currently imported and a method is added to the base function, which we want to avoid.

musm:
Or should I just document the change and if so where ? Readme file?

ggggggggg:
The more I think about it, the more I think printing them all out by default may not be very useful. For example, I often have 200+ groups in "/" in my hdf5 files, at that point I'd rather it print nothing. Maybe the functionality should just be renamed or removed instead of mapped onto part of `Base`? (and of course not error)

musm:
good point I have removed the verbose printing

musm:
What about renaming the old `dump` function to Base's `showall`?

musm:
@ggggggggg do you have any opinions on the old `dump` function? 
It seems somewhat useful to get a quick glance at the structure of the hdf5 file, but I'm not sure if it's needed or if so what it should be called etc.

ggggggggg:
I never used the previous `dump`, so I won't complain if it goes away. It does seem potentially useful, and I don't see any problem with keeping it.  The docs for `showall` say "Similar to show, except shows all elements of arrays." which doesn't seem to apply to anything other than array, which doesn't seem great for Base. `show_structure` seems better to me. It's a bit like a filesystem, but I don't think `walkdir` makes sense either.

musm:
@ggggggggg `show_structure` sounds like a good name for this function

musm:
Ok upon further thought I don't think the previous dump function is that useful nor robust. Deprecating it as in this PR seems the best strategy. 

I plan on merging this PR later this week sans objections.

PR #396, opened 2017-05-06
validate libhdf5 version in build.jl

ggggggggg:
This uses the validate functionality in BinDeps to validate the libhdf5 version number in build.jl. It should rebuild if it finds and out of date library. I'm not quite sure what the correct choice of `MINVERSION` is. It seems like this could be useful if people starting having problems with the fact that HDF5-1.8 can't read all files created by HDF5-1.10. I'm not convinced it's necessary, but I figured it would be good to put it out there.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -4,7 +4,7 @@ using BinDeps
 
 # https://support.hdfgroup.org/HDF5/ lists "Current Releases"
 # make sure we have one of those
-MINVERSION = v"1.8.0"
+const MINVERSION = v"1.8.0"
 function h5_get_libversion(n, h)
     f = Libdl.dlsym_e(h, "H5get_libversion")
     majnum, minnum, relnum = Ref{Cuint}(), Ref{Cuint}(), Ref{Cuint}()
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -2,22 +2,36 @@ using BinDeps
...
+# make sure we have one of those
+const MINVERSION = v"1.8.0"
+function h5_get_libversion(n, h)
+    f = Libdl.dlsym_e(h, "H5get_libversion")
+    majnum, minnum, relnum = Ref{Cuint}(), Ref{Cuint}(), Ref{Cuint}()
+    status = ccall(f,Cint, (Ptr{Cuint}, Ptr{Cuint}, Ptr{Cuint}), majnum, minnum, relnum)
+    status < 0 && error("Error getting HDF5 library version")
+    VersionNumber(majnum[], minnum[], relnum[])
+end
+validate(n,h) = h5_get_libversion(n,h) > MINVERSION
```
musm: shouldn't this be `≥` ?

ggggggggg:
fixes #287

musm:
Sans any objections I plan on merging this friday

musm:
Does https://github.com/ggggggggg/HDF5.jl/blob/75b9cbccf497389ea8a52cd1262649c75b0bbb5b/deps/build.jl#L38

Need to be tweaked?

ggggggggg:
Apparently 1.10.1 is out, but both pass the validate function so i can be a
separate PR easily.

musm:
sounds good. Yeah I'm not sure  that line does anything other than for source builds and as a fallback, but e.g. it won't work on windows.

ggggggggg:
It's not clear to me how BinDeps chooses which provider to try, or if it
tries a 2nd provider if the first fails. So it may never get used.

musm:
@ggggggggg thanks you !!

musm:
btw I'm wondering of a good way of deduplicating the validate function here and in src/HDF5.jl which are identical
perhaps we should write out the version in the deps.jl version?

ggggggggg:
I think writing the version number in deps.jl seems like the best option to deduplicate the code. Does bindeps has an intended way to write auxillary info like that.

PR #397, opened 2017-05-06
Clean up code, put exports at the top and do not reexport already exp…

musm:
…orted Base function and also only call `init_libhdf5()` once (inside `__init__`). In addition *turn off automatic error printing* unless in DEBUG mode (it's quite verbose and very unjulian)

In summary
- Move exports to the top
- Do not reexport already exported Base functions that are imported
- Remove extraneous call of `init_libhdf5()`
- Turn off automatic error printing unless DEBUG env variable set
- Do not use `@show` before an error branch

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -22,6 +22,6 @@ if is_apple()
 end
 
 provides(Sources, URI("https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.10/hdf5-1.10.0-patch1/src/hdf5-1.10.0-patch1.tar.gz"), hdf5)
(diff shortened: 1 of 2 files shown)


simonster:
I agree that automatic error printing is unjulian, but if we're going to turn it off, we should capture the error messages and include them in the error we throw. Otherwise it's going to be very challenging for people to figure out why their code isn't working.

musm:
I'm going to then remove that part of this PR since I don't know how to caputre the error.

musm:
Planning on merging this tomorrow. Please suggest a way to capture the error message if you know a good way to do that.

PR #401, opened 2017-05-09
fix doc about outdated types

ggggggggg:
Remove language about type `PlainHDF5File`.


musm:
sans objections I plan on merging this wensday

PR #411, opened 2017-05-28
Clean up and fix creating and testing of files

musm:


test/readremote.jl
```diff
@@ -3,186 +3,183 @@ using Base.Test
...
+     3   5   7;
+     6   7   8;
+     9   9   9;
      12  11  10]]
 
 
-    const savedir = joinpath(tempdir(), "h5")
-    if !isdir(savedir)
-        mkdir(savedir)
+const savedir = mktempdir()
```
tkelman: this is going to download these files repeatedly

musm: how so?

tkelman: every time it runs `mktempdir()` will be different

musm: Right, but that is only between separate runs of `Pkg.test`. This is no different than `savedir = joinpath(tempdir(), "h5")`  other than giving it a unique name.

I should add an `rm` for this test temp folder as well, however.

tkelman: Since these are downloading specific known files instead of creating some throwaway local file, it may be better to save them somewhere consistent. Giving it a unique name will result in repeated downloads, a predictable name will only download them once.

musm: I guess we can download the files in the test folder

PR #418, opened 2017-06-16
Clean up testing 

musm:
supersedes https://github.com/JuliaIO/HDF5.jl/pull/411


test/external.jl
```diff
@@ -45,4 +48,7 @@ close(source_file)
 # @test read(group2["1"])==1
 # @test read(group2["1.1"])==1.1
 
+rm(fn1)
+# rm(fn2)
```
musm: @ggggggggg trying to rm(fn2) here results is a `unlink: resource busy or locked (EBUSY)` error even though the handle has been explicitly closed, any ideas here?

musm:

test/external.jl
musm: probably related to the lib not properly detecting the closed handle

test/external.jl
```diff
@@ -3,46 +3,52 @@ using Base.Test
...
+# these tests succeed
+# reopening the target_file crashes due to "file close degree doesn't match"
+# target_file = h5open(target_file.filename, "r")
+# group2 = target_file["target_group"]["new_group"]
+# @test read(group2["abc"])=="abc"
+# @test read(group2["1"])==1
+# @test read(group2["1.1"])==1.1
+
+rm(fn1)
+# rm(fn2)
```
tkelman: what was preventing this from working? does the file need to manually be closed first?

musm: The problem is that even though it is manually closed, rm claims it is locked....

tkelman: might need to force finalize, or set to nothing and run gc a few times, or explicitly un-mmap more carefully. not a big deal, but good to get resources less leaky

musm: I think this is probably a bug in HDF5.jl, where a resource is not properly freed... Probably will need a simple MWE of the bug, but I haven't found something that exposes in a reduced case.

PR #419, opened 2017-06-17
Add missing return nothing to create_external

musm:


musm:

PR #420, opened 2017-06-18
Use unsafe_convert from Base

musm:
Remove Compat for unsafe_convert method


musm:
32-bit windows errors are unrelated

Issue #404, opened 2017-05-10
Update code for dropped 0.4 with regards to Strings


musm:
Code needs an update to reflect the changes in 0.5 in regards to strings

PR #421, opened 2017-06-18
Drop Compat.String requirements

musm:
close https://github.com/JuliaIO/HDF5.jl/issues/404


musm:
@ggggggggg 
Pretty sure this is a drop in replacement, but good to have a second set of eyes on this

PR #424, opened 2017-07-02
Use the BinDeps BSDPkg provider for building on FreeBSD

ararslan:
The provider uses the FreeBSD Ports Collection to install the HDF5 library system-wide, as is done with Apt et al. on Linux. The Autotools-based `BuildProcess` fails on FreeBSD.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -34,6 +34,10 @@ if is_apple()
     provides(Homebrew.HB, "homebrew/science/hdf5", hdf5, os=:Darwin)
 end
 
+if Sys.KERNEL === :FreeBSD
+    provides(BSDPkg, "hdf5", hdf5, os=:FreeBSD)
+end
+
 provides(Sources, URI("https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.10/hdf5-1.10.0-patch1/src/hdf5-1.10.0-patch1.tar.gz"), hdf5)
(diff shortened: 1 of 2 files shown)


deps/build.jl
```diff
@@ -34,6 +34,10 @@ if is_apple()
     provides(Homebrew.HB, "homebrew/science/hdf5", hdf5, os=:Darwin)
 end
 
+if Sys.KERNEL === :FreeBSD
```
musm: why not use `is_bsd()`?

ararslan: `is_bsd()` is true for any BSD-based system, including macOS, OpenBSD, DragonFly, etc. The `BSDPkg` provider only applies to FreeBSD.

musm: Oh right. I'm guessing it will probably make sense to  introduce a `Sys.isfreebsd()` function at some point after your PR.

musm: Actually, do we really need the `if` statement here... I think this can just be put in the `is_linux()` branch and let `BinDeps` find a provider.

ararslan: We actually don't need any of these branches, I was just matching style in this PR.

ararslan: FreeBSD isn't `is_linux` though

musm: Maybe it's best to just remove the `is_linux` branch and put these all together.

PR #425, opened 2017-07-02
Fix #379

### LINKED ISSUE #379: Not yet supported
Hi, I would like to read the timestamps of the attached file and get an "Not yet supported" error. Is there a way to just extract the column with the timestamp? Will this be supported in future?
Thanks, Lukas

[example.h5.zip](https://github.com/JuliaIO/HDF5.jl/files/850928/example.h5.zip)

tknopp:
Fixes #379 by simple increasing the membersize field. I am actually not sure what is correct here. Maybe directly 64bit?

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1437,7 +1437,7 @@ function read{N}(obj::HDF5Dataset, T::Union{Type{Array{HDF5Compound{N}}},Type{HD
         membertype = Vector{Type}(N)
         membername = Vector{String}(N)
         memberoffset = Vector{UInt64}(N)
-        membersize = Vector{UInt8}(N)
+        membersize = Vector{UInt16}(N)
         for i = 1:N
             filetype = HDF5Datatype(h5t_get_member_type(t.id, i-1))
             memberfiletype[i] = filetype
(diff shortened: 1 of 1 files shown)


ararslan:
I looked at the HDF5 documentation on [compound types](https://support.hdfgroup.org/HDF5/Tutor/compound.html) a bit and it doesn't specify a limit on the size of members, so I wonder what the rationale was for using `UInt8` here. Apparently this `membersize` vector was introduced in https://github.com/JuliaIO/HDF5.jl/pull/365, so maybe @grahamrow has some insight.

simonster:
`H5Tget_size` returns a `Csize_t`, but given that the HDF5 file format encodes the size of the type with 4 bytes, a `UInt32` is sufficient.

tknopp:
Updated according to @simonster suggestion.

I no can read the dataset from https://sourceforge.net/projects/ismrmrd/files/data/ although unfortunately, the returned data is still not what it is supposed to be. h5py seems to load it correctly. But I actually cannot understand howto correctly interpret the data. Seems to be an immutable embedded into an HDF5 file.

tknopp:
moved discussion on specific dataset and followup problems to #426.

musm:
@tknopp  care to add the failing example as a test here?  FYI 32-bit windows failure is unrelated

tknopp:
My test data can still not be loaded (see #426) so that is not suitable for a test case.

PR #428, opened 2017-07-17
replace unsafe pointer operations with their gc-safe Ref equivalents

vtjnash:
Avoids using `pointer` so that Julia can natively track the usage of the objects and ensure it is gc-rooted.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -5,13 +5,12 @@ module HDF5
 using Compat
 
 using Base: unsafe_convert
+using Compat: StringVector
 
 import Base:
     close, convert, done, eltype, endof, flush, getindex, ==,
     isempty, isvalid, length, names, ndims, next, parent, read,
     setindex!, show, size, sizeof, start, write, isopen
-
-import Compat: StringVector
 
 export
     # types
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -6,10 +6,12 @@ using Compat
 
 using Base: unsafe_convert
 
-import Base: 
+import Base:
     close, convert, done, eltype, endof, flush, getindex, ==,
     isempty, isvalid, length, names, ndims, next, parent, read,
     setindex!, show, size, sizeof, start, write, isopen
+
+import Compat: StringVector
 
 export
     # types
@@ -1330,7 +1332,7 @@ function read{S<:String}(obj::DatasetOrAttribute, ::Type{S})
     objtype = datatype(obj)
     try
         if h5t_is_variable_str(objtype.id)
-            buf = Ptr{UInt8}[C_NULL]
+            buf = Cstring[C_NULL]
             memtype_id = h5t_copy(H5T_C_S1)
             h5t_set_size(memtype_id, H5T_VARIABLE)
             h5t_set_cset(memtype_id, h5t_get_cset(datatype(obj)))
@@ -1371,10 +1373,10 @@ function read{S<:String}(obj::DatasetOrAttribute, ::Type{Array{S}})
         ret = Array{S}(sz)
         if isvar
             # Variable-length
-            buf = Vector{Ptr{UInt8}}(len)
+            buf = Vector{Cstring}(len)
             h5t_set_size(memtype_id, H5T_VARIABLE)
             readarray(obj, memtype_id, buf)
-            # FIXME? Who owns the memory for each string? Will Julia free it?
+            # FIXME? Who owns the memory for each string? Julia v0.6+ won't free it.
             for i = 1:len
                 ret[i] = unsafe_string(buf[i])
             end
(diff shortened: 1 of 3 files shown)


src/HDF5.jl
```diff
@@ -1371,10 +1373,10 @@ function read{S<:String}(obj::DatasetOrAttribute, ::Type{Array{S}})
         ret = Array{S}(sz)
         if isvar
             # Variable-length
-            buf = Vector{Ptr{UInt8}}(len)
+            buf = Vector{Cstring}(len)
             h5t_set_size(memtype_id, H5T_VARIABLE)
             readarray(obj, memtype_id, buf)
-            # FIXME? Who owns the memory for each string? Will Julia free it?
+            # FIXME? Who owns the memory for each string? Julia v0.6+ won't free it.
```
musm: Is this comment still valid ?

src/HDF5.jl
```diff
@@ -6,11 +6,13 @@ using Compat
 
 using Base: unsafe_convert
 
-import Base: 
+import Base:
     close, convert, done, eltype, endof, flush, getindex, ==,
     isempty, isvalid, length, names, ndims, next, parent, read,
     setindex!, show, size, sizeof, start, write, isopen
 
+import Compat: StringVector
```
musm: Should be `using` since `StringVector` not extended.

src/HDF5.jl
```diff
@@ -1926,11 +1927,11 @@ function h5d_write{T<:HDF5Scalar}(dataset_id::Hid, memtype_id::Hid, x::T)
...
-    p = Array{Ptr{UInt8}}(size(strs))
-    for i = 1:len
-        p[i] = !isassigned(strs, i) || isempty(strs[i]) ? pointer(EMPTY_STRING) : pointer(strs[i])
+    nonnull_str = copy(strs)
+    for i = 1:length(nonnull_str)
+        isassigned(nonnull_str, i) || (nonnull_str[i] = "")
     end
+    p = Ref{Cstring}(nonnull_str)
     h5d_write(dataset_id, memtype_id, H5S_ALL, H5S_ALL, H5P_DEFAULT, p)
 end
```
musm: if `str = ["\0","a","b"]`
then
```julia

julia> p = Ref{Cstring}(nonnull_str)
ERROR: ArgumentError: embedded NULs are not allowed in C strings: "\0"
```
Is this expected?

src/HDF5.jl
```diff
@@ -1926,11 +1927,11 @@ function h5d_write{T<:HDF5Scalar}(dataset_id::Hid, memtype_id::Hid, x::T)
...
 function h5d_write{S<:String}(dataset_id::Hid, memtype_id::Hid, strs::Array{S})
-    len = length(strs)
-    p = Array{Ptr{UInt8}}(size(strs))
-    for i = 1:len
-        p[i] = !isassigned(strs, i) || isempty(strs[i]) ? pointer(EMPTY_STRING) : pointer(strs[i])
+    nonnull_str = copy(strs)
+    for i = 1:length(nonnull_str)
+        isassigned(nonnull_str, i) || (nonnull_str[i] = "")
     end
+    p = Ref{Cstring}(nonnull_str)
```
musm: if `str = ["\0","a","b"]`
then
```julia

julia> p = Ref{Cstring}(nonnull_str)
ERROR: ArgumentError: embedded NULs are not allowed in C strings: "\0"
```
Is this expected?

vtjnash: Yes, HDF5 appears to only support C-type strings (the string "\0" will get truncated to ""). The error message seems pretty self-explanitory to me? 🙂

musm: Sorry I wasn't clear. I meant to say that with this PR the following doesn't work (i.e. the "\0" doesn't get truncated to "")

```julia
using HDF5
fn = tempname()
f = h5open(fn, "w")
test = ["s","dfs","Hi there","\0"]
write(f, "test", test)
ERROR: ArgumentError: embedded NULs are not allowed in C strings: "\0"
```
whereas on master it truncates the last element of `test` to `""`

vtjnash: Right, the last element should be `""`. Why would you ever want it to silently discard / truncate your data? :)

musm: Sry just wanted to confirm the behavior change.

vtjnash: Np, happy to confirm

musm: Isn't there still an inconsistency here 
For arrays with `"\0"` it will throw an ArgumentError, however if you just try to write  a string with `"\0"` it will still truncate it to `""` on this PR.

```julia
using HDF5
fn = tempname()
f = h5open(fn, "w")
write(f, "test", "\0")
```

vtjnash: File another bug?

src/HDF5.jl
musm: Just to double check, do you know whether this comment still applies with your fix? If not no problem just wanted to know if this `#FIXME` can be removed or not. Thanks.

vtjnash: I only answered one of the questions there. I suspect that implies the answer the the first question (is that this is a memory leak and we should be manually freeing the memory here after the call to `unsafe_string`), but perhaps instead HDF5 owns the memory (as a pointer into the mmap'd file?) and attempting to free it would immediately cause it to crash.

vtjnash: Checking the docs, it appears there may supposed to be a call to `H5Dvlen_reclaim` before the call to `h5t_close` (and this function should perhaps be wrapped in try/catch so those get called even if there's an error with reading part of the data.

PR #432, opened 2017-07-18
Remove some excess conditionals in the build script

ararslan:
Non-functional change

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -26,7 +26,9 @@ provides(BSDPkg, "hdf5", hdf5, os=:FreeBSD)
 if is_windows()
     using WinRPM
     provides(WinRPM.RPM, "hdf5", hdf5, os=:Windows)
-elseif is_apple()
+end
+
+if is_apple()
     using Homebrew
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -18,26 +18,19 @@ hdf5 = library_dependency("libhdf5",
...
+provides(Yum, "hdf5", hdf5, os=:Linux)
+provides(BSDPkg, "hdf5", hdf5, os=:FreeBSD)
 
 if is_windows()
     using WinRPM
     provides(WinRPM.RPM, "hdf5", hdf5, os=:Windows)
-end
-
-if is_apple()
+elseif is_apple()
```
musm: Can we leave this as it was , otherwise LGTM
```julia
if is_windows()
    using WinRPM
    provides(WinRPM.RPM, "hdf5", hdf5, os=:Windows)
end

if is_apple()
    using Homebrew
    provides(Homebrew.HB, "hdf5", hdf5, os=:Darwin)
end
```

PR #436, opened 2017-07-29
Call h5d_vlen_reclaim to release resource

musm:
release the memory that the HDF5 library allocates to store the string variable

Not sure how to test this

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1401,7 +1401,7 @@ function read{S<:String}(obj::DatasetOrAttribute, ::Type{Array{S}})
         close(dspace)
         h5t_close(memtype_id)
     end
-    return ret
+        return ret
     end
 end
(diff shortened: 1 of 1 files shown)


src/HDF5.jl
```diff
@@ -1395,7 +1395,13 @@ function read{S<:String}(obj::DatasetOrAttribute, ::Type{Array{S}})
             end
         end
     end
-    h5t_close(memtype_id)
+    try
+        dspace = dataspace(obj)
```
simonster: I don't know if we care about performance, but if we do: When we call `size(obj)` we're also constructing a dataspace object, and we don't need to do it twice.

musm: I see, thanks for pointing this out. I'm not sure if we should worry about performance at this point.

musm: I don't see an obvious way of returning the dataspace object that  `size` uses  so that It could be used instead of creating another `dataspace` object here.

musm:
Will merge tomorrow sans feedback

PR #442, opened 2017-08-31
add support for CartesianRange indexing 

jingpengw:
CartesianRange is widely used to handle ranges, adding support of CartesianRange to index the arrays inside dataset could make some implementation easier.

These functions were moved from BigArrays.jl
https://github.com/seung-lab/BigArrays.jl/issues/21

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -710,6 +710,12 @@ function h5read(filename, name::String, indices::Tuple{Vararg{Union{Range{Int},I
     dat
 end
 
+function h5read{N}(filename, name::String, cr::CartesianRange{CartesianIndex{N}})
+    # transfer to unit range
+    ur = ( map((x,y)->x:y, cr.start, cr.stop)...)
+    h5read(filename, name, ur)
+end 
+
 function h5writeattr(filename, name::String, data::Dict)
     fid = h5open(filename, "r+")
     try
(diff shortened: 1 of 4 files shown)


src/HDF5.jl
```diff
@@ -958,6 +970,15 @@ function setindex!(p::HDF5Properties, val, name::String)
     funcset(p, val...)
     return p
 end
+
+# cartesian range indexing 
+function setindex!{N}(dset::HDF5Dataset, val, cr::CartesianRange{CartesianIndex{N}})
+    # transfer to unit range
+    ur = map((x,y)->x:y, cr.start.I, cr.stop.I)
+    @show ur
```
musm: why `@show ur`  here ?

musm:
This functionality seems useful although I am not so sure that the implementation is optimal.

I wonder if anyone else has any comments

src/HDF5.jl
```diff
@@ -710,6 +710,12 @@ function h5read(filename, name::String, indices::Tuple{Vararg{Union{Range{Int},I
     dat
 end
 
+function h5read{N}(filename, name::String, cr::CartesianRange{CartesianIndex{N}})
+    # transfer to unit range
+    ur = map((x,y)->x:y, cr.start.I, cr.stop.I)
```
timholy: Better to use `first(cr)` and `last(cr)` than `cr.start.I` and `cr.stop.I`; in 0.7, the representation of CartesianRange has [changed](https://github.com/JuliaLang/julia/pull/20974). Indeed, in 0.7 this can just be `ur = cr.indices`.

test/cartesian_range.jl
```diff
@@ -0,0 +1,20 @@
...
+
+filename = "$(tempname()).h5"
+
+h5write(filename, "main", Array(1:10))
+@test h5read(filename, "main", CartesianRange((3:5,))) == [3:5;]
+f = h5open(filename,"r+")
+dset = f["main"]
+# setindex of cartesian range
+dset[CartesianRange((3:5,))] = [4:6;]
+@test dset[CartesianRange((3:5,))] == [4:6;]
```
ggggggggg: The tests should cover a case with more than one dimension. Also, why is the correct answer not `[3:5;]` as you would get with a normal range?

musm: because on  line 14 he rewrites elements 3:5 to 4:6 (which is testing setindex!)

musm:
addressed review in https://github.com/JuliaIO/HDF5.jl/pull/459#pullrequestreview-80685512

PR #443, opened 2017-09-13
Fix most of depwarns on 0.7

yuyichao:


test/readremote.jl
```diff
@@ -4,29 +4,29 @@ using Base.Test
 @testset "readremote" begin
 
 # check that we can read the official HDF5 example files
-const urlbase = "https://support.hdfgroup.org/ftp/HDF5/examples/files/exbyapi/"
+urlbase = "https://support.hdfgroup.org/ftp/HDF5/examples/files/exbyapi/"
```
musm: why get rid of the `const` ?

PR #445, opened 2017-10-05
Fix heuristic chunking of zero-dimensional arrays

jipolanco:
Previously, when calling `heuristic_chunk` with a zero-dimensional array
`x` (i.e. a scalar wrapped by an array, for example `x = ones()`),
heuristic chunking failed with an error, since `shape = size(x)` was an
empty Tuple, and thus `[shape...]` was an `Array{Any,1}` instead of the
expected `Array{Int,1}`.

In practice, this disables chunking of zero-dimensional arrays (chunking
scalars doesn't make much sense anyway!).

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -566,6 +566,7 @@ function heuristic_chunk(T, shape)
     end
     return chunk
 end
+heuristic_chunk(T, ::Tuple{}) = Int[]
 heuristic_chunk{T}(A::AbstractArray{T}) = heuristic_chunk(T, size(A))
 heuristic_chunk(x) = Int[]
 # (strings are saved as scalars, and hence cannot be chunked)
--- a/test/plain.jl
+++ b/test/plain.jl
@@ -13,6 +13,8 @@ f["Float64"] = 3.2
 f["Int16"] = Int16(4)
 # compression of empty array (issue #246)
 f["compressedempty", "shuffle", (), "compress", 4] = Int64[]
+# compression of zero-dimensional array (pull request #445)
+f["compressed_zerodim", "shuffle", (), "compress", 4] = fill(Int32(42), ())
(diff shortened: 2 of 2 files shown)


musm:
Could we have some tests?

musm:
> In practice, this disables chunking of zero-dimensional arrays (chunking

Isn't that case already covered in https://github.com/JuliaIO/HDF5.jl/pull/445/files#diff-02840edcdbd42c5b41e89b3f700b9a28R552

jipolanco:
Not really. In the code you highlighted, the value of `sz` is 1 and not 0 when the input array `A` is zero-dimensional. This is because `shape(A)` returns the empty tuple `()`, and `prod(())` returns 1 (which doesn't necessarily make much sense...).

Extended example:
```julia
A = zeros()
shape = size(A)
@assert shape === ()
sz = prod(shape)
@assert sz == 1
```

I'll provide some tests soon.

musm:
Ok so that line should be changed to
`isempty(shape) && return Int[]` 
to cover that case

jipolanco:
But then it wouldn't work with empty arrays of dimensions >= 1, which would break the current behaviour of that line.

For instance:
```julia
x = zeros(0, 0, 0)  # empty 3D array
sz = size(x)  # (0, 0, 0)
@assert prod(sz) == 0
@assert !isempty(sz)
```

jipolanco:
An alternative would be to add a variant of `heuristic_chunk` for empty tuples:

```julia
heuristic_chunk(::, ::Tuple{}) = Int[]
```

musm:
Yeah that seems resonable

jipolanco:
Ok, then I'll soon update the PR with that alternative and a test.

musm:
Great thank you

jipolanco:
I just updated the fix (I modified the original commit using a force push), and added tests.

Issue #434, opened 2017-07-19
HDF5 silently truncates "\0" 


musm:
```julia
using HDF5
fn = tempname()
f = h5open(fn, "w")
write(f, "test", "\0")
read(f,"test") == "" # true
```
we should probably just throw an argument error in this case instead  of silently truncating. 

This is also very bad since it will truncate, e.g., "SDF\0nooo" to "SDF"

PR #450, opened 2017-11-21
Do not silently truncate strings with embeded Nulls

musm:
close https://github.com/JuliaIO/HDF5.jl/issues/434

pinging @ggggggggg and @simonster for a review. 
The current behavior is nasty since it silently truncates, e.g,  "SDF\0nooo" to "SDF"

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1906,7 +1906,6 @@ end
 ### Convenience wrappers ###
 # These supply default values where possible
 # See also the "special handling" section below
-const EMPTY_STRING = UInt8[0x00]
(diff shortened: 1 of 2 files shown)


ggggggggg:
I'm not very familiar with c strings, but it looks like it will do what you said. Maybe test with setindex syntax as well, eg `f["trunc3"]="\0"`

simonster:
I think it would be better just to write this as a `ccall` to `H5Dwrite` where the string is typed as `Cstring`, which should do the same check and avoid copying if the string is already null-terminated.

musm:
@simonster  Good call, updated  to instead just do a  `unsafe_convert` so that we can reuse the `h5d_write` instead of a  direct call to  `ccall` in the method.

musm:
Nvm, it was not possible to do what I mentioned in the previous post. I explicitly added the `ccall` in the method, should be g2g now.

PR #451, opened 2017-11-22
Remove isassigned for writing string arrays

musm:
Right now saving an uninitialized string array, initializes the values to be `""` . I can't think of a good reason to allow this or why anyone would have to save an uninitialized string array. I think this could hide a potential issue/bug, when a user attempts to save a string array (that is believed to have proper references) and HDF5 silently initializes the array to `""`. 

Example (current behavior)
```julia
strs = similar(rand(3),String) # strs = String[#undef, #undef, #undef]
julia> h5write(f,"d",strs)
julia> h5read(f,"d")
3-element Array{String,1}:
 ""
 ""
 ""
```

**vs this pr which instead throws**
```
strs = similar(rand(3),String) # strs = String[#undef, #undef, #undef]
julia> h5write(f,"d",strs)
ERROR: UndefRefError: access to undefined reference
```

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1922,7 +1922,7 @@ function h5d_write(dataset_id::Hid, memtype_id::Hid, x::T) where {T<:HDF5Scalar}
     h5d_write(dataset_id, memtype_id, H5S_ALL, H5S_ALL, H5P_DEFAULT, tmp)
 end
 function h5d_write(dataset_id::Hid, memtype_id::Hid, strs::Array{S}) where {S<:String}
-    p = Ref{Cstring}(nonnull_str)
+    p = Ref{Cstring}(strs)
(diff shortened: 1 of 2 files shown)


musm:
If you try to write uninitialized values, the the current behavior sets them to `""`
Should we just ban this
```julia
strs = similar(rand(3),String) # strs = String[#undef, #undef, #undef]
julia> h5write(f,"d",strs)
julia> h5read(f,"d")
3-element Array{String,1}:
 ""
 ""
 ""
```


vs this pr which instead throws
```
ERROR: UndefRefError: access to undefined reference
```

musm:
Regardless of the decision we should add a test

musm:
Any thoughts on this?

musm:
I will merge in a few days sans any further objections

PR #456, opened 2017-11-23
Enable direct chunk writing

adambrewster:
Direct chunk writing speeds up creation of large chunked and compressed files in parallel environments by allowing workers to compress the chunks in parallel.  It's a cumbersome interface, but can provide significant speedup in data preparation steps, for example, where other file formats are converted to HDF5.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -36,6 +36,12 @@ else
     error("HDF5 not properly installed. Please run Pkg.build(\"HDF5\")")
 end
 
+let d = dirname(libhdf5), b = basename(libhdf5)
+    base,ext = split(b, ".", limit=2)
+    hl = joinpath(d, "libhdf5_hl.$ext")
+    global const libhdf5_hl = isfile(hl) ? hl : ""
+end
+
 function init_libhdf5()
     status = ccall((:H5open, libhdf5), Cint, ())
     status < 0 && error("Can't initialize the HDF5 library")
@@ -2062,7 +2068,12 @@ for (jlname, h5name, outtype, argtypes, argsyms, msg) in
     )
 
     ex_dec = funcdecexpr(jlname, length(argtypes), argsyms)
-    ex_ccall = ccallexpr(libhdf5, h5name, outtype, argtypes, argsyms)
+    library = startswith(string(h5name), "H5DO") ? libhdf5_hl : libhdf5
+    if library == ""
+        # If the high-level library is not installed, then skip these functions
+        continue
+    end
+    ex_ccall = ccallexpr(library, h5name, outtype, argtypes, argsyms)
     ex_body = quote
         status = $ex_ccall
         if status < 0
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -2030,7 +2030,7 @@ for (jlname, h5name, outtype, argtypes, argsyms, msg) in
      (:h5d_close, :H5Dclose, Herr, (Hid,), (:dataset_id,), "Error closing dataset"),
      (:h5d_flush, :H5Dflush, Herr, (Hid,), (:dataset_id,), "Error flushing dataset"),
      (:h5d_oappend, :H5DOappend, Herr, (Hid, Hid, Cuint, Hsize, Hid, Ptr{Void}) , (:dset_id, :dxpl_id, :index, :num_elem, :memtype, :buffer), "error appending"),
-     (:h5do_write_chunk, :H5DOwrite_chunk, Herr, (Hid, Hid, Int32, Ptr{Hsize}, Hsize, Ptr{Void}), (:dset_id, :dxpl_id, :filter_mask, :offset, :bufsize, :buf), "Error writing chunk"),
+     (:h5do_write_chunk, :H5DOwrite_chunk, Herr, (Hid, Hid, Int32, Ptr{Hsize}, Csize_t, Ptr{Void}), (:dset_id, :dxpl_id, :filter_mask, :offset, :bufsize, :buf), "Error writing chunk"),
      (:h5d_refresh, :H5Drefresh, Herr, (Hid,), (:dataset_id,), "Error refreshing dataset"),
      (:h5d_set_extent, :H5Dset_extent, Herr, (Hid, Ptr{Hsize}), (:dataset_id, :new_dims), "Error extending dataset dimensions"),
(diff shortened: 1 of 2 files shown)


adambrewster:
Tests are to be failing because the function is in the high level library.  It works for me because I hacked deps/deps.jl to point to libhdf5_hl.so instead of libhdf5.so.  I don't know if it's appropriate to make that change as a part of this or if I need to find a way to load both libraries.

musm:
Is it possible to just get away with `libhdf5_hl` , I wouldn't think so. We would probably need to load both libraries, but `libhdf5_hl` should always be installed along with `libhdf5`, right?

adambrewster:
Just using hdf5_hl works if the runtime linker can find the correct version of hdf5.

From `objdump -x libhdf5_hl.so`:
> Dynamic Section:
>   NEEDED               libhdf5.so.7
>   NEEDED               libc.so.6
>   SONAME               libhdf5_hl.so.7

If I understand correctly, this means that when `libhdf5_hl.so` is `dlopen`ed, the linker will search for `libhdf5.so.7` in its usual places and make all of the symbols from both of libraries available.  Unfortunately, there's no DT_RUNPATH or DT_RPATH, so if I set `libhdf5 = "/opt/fancy/libhdf5_hl.so"` in `deps.jl`, then I would also have to set LD_LIBRARY_PATH or something to tell the linker to find the right version of libhdf5.  That's probably okay for most users, but it looks like there's been some effort into making things work for people who want to use parallel versions (`libhdf5_openmpi` or `libhdf5_mpich`) of the library, so I need to keep that working.

This change tests for the presence of libhdf5_hl in the same directory as the automatically discovered or user specified libhdf5.  If it exists, then the extra functions are created (note that it also affects H5DOappend, which is also in the high level library).

It looks like the tests are passing for v0.6 and the failures under nightly are unrelated to this change.

src/HDF5.jl
```diff
@@ -1798,6 +1804,20 @@ function d_create_external(parent::Union{HDF5File, HDF5Group}, name::String, fil
 end
 d_create_external(parent::Union{HDF5File, HDF5Group}, name::String, filepath::String, t::Type, sz::Dims) = d_create_external(parent, name, filepath, t, sz, 0)
 
+function do_write_chunk(dataset::HDF5Dataset, offset, chunk_bytes::Vector{UInt8}, filter_mask=0)
+    checkvalid(dataset)
+    offs = collect(Hsize, reverse(offset))-1
+    h5do_write_chunk(dataset, H5P_DEFAULT, UInt32(filter_mask), offs, length(chunk_bytes), chunk_bytes)
+end
+
+immutable HDF5ChunkStorage
```
musm: `struct`

src/HDF5.jl
```diff
@@ -36,6 +36,12 @@ else
     error("HDF5 not properly installed. Please run Pkg.build(\"HDF5\")")
 end
 
+let d = dirname(libhdf5), b = basename(libhdf5)
+    base,ext = split(b, ".", limit=2)
```
musm: `splitext(b)`

src/HDF5.jl
```diff
@@ -36,6 +36,12 @@ else
...
+    hl = joinpath(d, "libhdf5_hl.$ext")
+    global const libhdf5_hl = isfile(hl) ? hl : ""
+end
```
musm: why not something simpler (setting the sentinel value to `nothing` instead of `""`)
```julia
hdf5_hl = joinpath(dirname(libhdf5), "libhdf5_hl.$(Libdl.dlext)")
const libhdf5_hl = isfile(hdf5_hl ) ? hdf5_hl : nothing
```

src/HDF5.jl
```diff
@@ -2047,7 +2068,12 @@ for (jlname, h5name, outtype, argtypes, argsyms, msg) in
     )
 
     ex_dec = funcdecexpr(jlname, length(argtypes), argsyms)
-    ex_ccall = ccallexpr(libhdf5, h5name, outtype, argtypes, argsyms)
+    library = startswith(string(h5name), "H5DO") ? libhdf5_hl : libhdf5
+    if library == ""
```
musm: with the above change this would then be `library === nothing`

ggggggggg:
There is a commented out test for `H5DOappend` [here](https://github.com/JuliaIO/HDF5.jl/blob/c04c6206a885474f7ad99878d89649c68401e780/test/swmr.jl#L34). You could uncommon it and see if it works, though that could easily be a later PR if you don't want to.

musm:
Also I am not sure if it is worth it but alternative strategy could be to add the high level library to bindeps `build.jl` as an optional dependency (not sure if it is possible to add optional dependencies)....

src/HDF5.jl
adambrewster:
```
julia> splitext("/usr/lib/libhdf5.so.10")
("/usr/lib/libhdf5.so", ".10")
```

src/HDF5.jl
adambrewster: Maybe it doesn't matter, but I think I wanted the compiler to know that
```
library = startswith(string(h5name), "H5DO") ? libhdf5_hl : libhdf5
```
to always be a string.

adambrewster:
I originally tried to add the dependency via bindeps, but couldn't find a way to get it to work well.  I think the problem was that I couldn't make it find corresponding libhdf5.so and libhdf5_hl.so libraries in case there are multiple versions of hdf5 installed (i.e. the same as just switching to  "libhdf5_hl").

I've never used H5DO_append.    That test fails with or without this change.

musm:
any ideas on the 32-bit win failures (they look related to this pr)

ggggggggg:
I've also never used `H5DO_append`, I'll try to get the test working after you resolve this.

adambrewster:
The last commit fixes the 32 bit windows failure (Hsize != Csize_t). Still failing on the nightly.

musm:
Thanks,  master failures are unrelated. Will merge tomorrow sans any objections

test/plain.jl
```diff
@@ -353,4 +353,18 @@ Wr = h5read(fn, "newgroup/W")
 close(f)
 rm(fn)
 
+# Test direct chunk writing
+h5open(fn, "w") do f
+  d = d_create(f, "dataset", datatype(Int), dataspace(4, 4), "chunk", (2, 2))
+  raw = HDF5ChunkStorage(d)
```
andreasnoack: This test should be conditional on the high-level library being installed

PR #457, opened 2017-11-27
update coverage  coveralls badge

musm:


musm:
looks like some of the webhooks are no longer working, someone else with permissions will need to fix/update those

PR #459, opened 2017-12-02
Add cartesian range indexing

musm:
supersedes https://github.com/JuliaIO/HDF5.jl/pull/442 and updated based on review comments


src/HDF5.jl
```diff
@@ -709,6 +709,16 @@ function h5read(filename, name::String, indices::Tuple{Vararg{Union{AbstractRang
     dat
 end
 
+function h5read(filename, name::String, cr::CartesianRange)
```
musm: @timholy  it seems that ` cr::CartesianRange` is sufficient in the function signature, is the  `cr::CartesianRange{CartesianIndex{N}}) where {N}...` restriction needed  (which actually doesn't work on 0.7)?

musm:
Actually I  think these should  be `CartesianIndex` not `CartesianRanges`. It's not like you can index regular arrays with a cartesianrange, so I don't know why you should be able to do the same with  hdf5 data.

musm:
Closing this since there doesn't seem to be strong evidence that the functionality is very useful or necessary, please let me know if it is indeed useful.

PR #462, opened 2018-02-15
Add support for MPIO and major rework HDF5 property handling

tbole:
I was quite astonished that the HDF5 wrapper did not yet have support for MPIO since this is quite important on clusters.

 - Add support for MPIO: this is implemented using a thin wrapper  
   without adding a dependency to MPI.jl. This requires at opening
   a file in MPIO mode, thus setting file access / file creation
   properties and setting the dataset transfer property directly when
   writing a dataset. The latter has been implementing by extending the
   dataset struct itself by a xfer property, which can be used when
   writing the dataset. The type wrapper provides primitive type handles
   of size 4 and 8 byte, which should account for MPICH/OpenMPI.
   The particular implementation needs to be provided by the user.
 - MPIO in HDF5 partially depends on MPI.jl supporting MPI_Info or some 
   other method to provide that data. There's another [pull request](https://github.com/JuliaParallel/MPI.jl/pull/198) of mine over
   at MPI.jl adding that support, which has been dangling around for some 
  months without comments.
   MPI sadly does not see to much support and enthusiasm in Julia for sure.
  Hopefully this merge request here will speed up things.

 - These changes required some massive rework of the various properties
   that HDF5 provides. HDF5.jl has been mostly operating using standard
   properties in nearly every case. Most routines have been extended to
   take a set of properties consisting of key value pairs. The property
   get/set dict has been extended by the property identifiers. This now
   permits creating the specific properties (dcpl, dapl, fapl, dxpl,...)
   just using the list of keys. These are now passed through most
   routines so we should be able to set nearly all properties that HDF5
   provides. In addition to that there are some minor bugfixes and
   additions.
 - `read(parent, name::String...)` is a special case, since in datafile.jl
   it is overloaded for reading multiple datasets. However this
   conflicts with function signature of the the property list implementation.
   The particular routine thus has been renamed `readp(parent, name, pv...)`.
 - The libhdf5 aliases are updated to preferably use parallel HDF5
   versions.
 - Some tests are added. By default the MPIO will be tested in serial
   mode. For parallel tests we need to execute mpirun.
 - Fixed a wrong prop setting in the h5p_get/set_driver/info setting.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -26,7 +26,7 @@ export
     h5open, h5read, h5rewrite, h5writeattr, h5readattr, h5write,
     has, iscontiguous, ishdf5, ismmappable, name,
     o_copy, o_delete, o_open, p_create,
-    readmmap, @read, @write, root, set_dims!, t_create, t_commit
+    readmmap, @read, readp, @write, root, set_dims!, t_create, t_commit
 
 include("datafile.jl")
 
@@ -75,6 +75,13 @@ const Hssize  = Int64
 const Htri    = Cint   # pseudo-boolean (negative if error)
 const Haddr   = UInt64
 
+# MPI communicators required by H5P
+abstract  type Hmpih end
+primitive type Hmpih32 <: Hmpih 32 end # MPICH C/Fortran, OpenMPI Fortran: 32 bit handles
+primitive type Hmpih64 <: Hmpih 64 end # OpenMPI C: pointers (mostly 64 bit)
+const mpihandles = Dict(4 => Hmpih32, 8 => Hmpih64)
+h5_mpihandle(handle) = reinterpret(mpihandles[sizeof(handle)],handle)
+
 # Function to extract exported library constants
 # Kudos to the library developers for making these available this way!
 const libhdf5handle = Libdl.dlopen(libhdf5)
@@ -120,6 +127,14 @@ const H5F_CLOSE_DEFAULT = 0
 const H5F_CLOSE_WEAK    = 1
 const H5F_CLOSE_SEMI    = 2
 const H5F_CLOSE_STRONG  = 3
+# file driver constants
+const H5FD_MPIO_INDEPENDENT    = 0
+const H5FD_MPIO_COLLECTIVE     = 1
+const H5FD_MPIO_CHUNK_DEFAULT  = 0
+const H5FD_MPIO_CHUNK_ONE_IO   = 1
+const H5FD_MPIO_CHUNK_MULTI_IO = 2
+const H5FD_MPIO_COLLECTIVE_IO  = 0
+const H5FD_MPIO_INDIVIDUAL_IO  = 1
 # object types (C enum H5Itype_t)
 const H5I_FILE         = 1
 const H5I_GROUP        = 2
@@ -359,9 +374,10 @@ end
 mutable struct HDF5Dataset
     id::Hid
     file::HDF5File
+    xfer::Hid
 
-    function HDF5Dataset(id, file)
-        dset = new(id, file)
+    function HDF5Dataset(id, file, xfer=H5P_DEFAULT)
+        dset = new(id, file, xfer)
         @compat finalizer(close, dset)
         dset
     end
@@ -369,7 +385,7 @@ end
 convert(::Type{Hid}, dset::HDF5Dataset) = dset.id
 function show(io::IO, dset::HDF5Dataset)
     if isvalid(dset)
-        print(io, "HDF5 dataset: ", name(dset), " (file: ", dset.file.filename, ")")
+        print(io, "HDF5 dataset: ", name(dset), " (file: ", dset.file.filename, "xfer_mode: ", dset.xfer ," )" )
     else
         print(io, "HFD5 dataset (invalid)")
     end
@@ -436,9 +452,10 @@ attrs(p::Union{HDF5File, HDF5Group, HDF5Dataset}) = HDF5Attributes(p)
 mutable struct HDF5Properties
     id::Hid
     toclose::Bool
+    class::Hid
 
-    function HDF5Properties(id, toclose::Bool=true)
-        p = new(id, toclose)
+    function HDF5Properties(id, toclose::Bool=true, class::Hid=H5P_DEFAULT)
+        p = new(id, toclose, class)
         if toclose
             @compat finalizer(close, p)
         end
@@ -551,6 +568,18 @@ end
 # Blosc compression:
 include("blosc_filter.jl")
 
+function checkprops(pv...)
+    if !iseven(length(pv))
+        error("Properties and values must come in pairs")
+    end
+    for i=1:2:length(pv)
+        if !isa(pv[i], AbstractString)
+            error("Keyword argument $i of key/value pairs should be a String, but it's a ", typeof(pv[i]))
+        end
+    end
+    nothing
+end
+
 # heuristic chunk layout (return empty array to disable chunking)
 function heuristic_chunk(T, shape)
     Ts = sizeof(T)
@@ -620,34 +649,31 @@ end
 Open or create an HDF5 file where `mode` is one of:
  - "r"  read only
  - "r+" read and write
+ - "cw" read and write, create file if not existing, to not truncate
  - "w"  read and write, create a new file (destroys any existing contents)
 
 Pass `swmr=true` to enable (Single Writer Multiple Reader) SWMR write access for "w" and
 "r+", or SWMR read access for "r".
 """
 function h5open(filename::AbstractString, mode::AbstractString="r", pv...; swmr=false)
+    checkprops(pv...)
     # pv is interpreted as pairs of arguments
     # the first of a pair is a key of hdf5_prop_get_set
     # the second of a pair is a property value
-    apl = p_create(H5P_FILE_ACCESS) # access property list
+    fapl = p_create(H5P_FILE_ACCESS, false, pv...) # file access property list
     # With garbage collection, the other modes don't make sense
     # (Set this first, so that the user-passed properties can overwrite this.)
-    apl["fclose_degree"] = H5F_CLOSE_STRONG
-    for i = 1:2:length(pv)
-        thisname = pv[i]
-        if !isa(thisname, String)
-            error("Argument ", i+2, " should be a String, but it's a ", typeof(thisname))
-        end
-        apl[thisname] = pv[i+1]
-    end
+    fapl["fclose_degree"] = H5F_CLOSE_STRONG
+    fcpl = p_create(H5P_FILE_CREATE, false, pv...) # file create property list
     modes =
         mode == "r"  ? (true,  false, false, false, false) :
         mode == "r+" ? (true,  true,  false, false, true ) :
+        mode == "cw" ? (false, true,  true,  false, true ) :
         mode == "w"  ? (false, true,  true,  true,  false) :
         # mode == "w+" ? (true,  true,  true,  true,  false) :
         # mode == "a"  ? (true,  true,  true,  true,  true ) :
         error("invalid open mode: ", mode)
-    h5open(filename, modes..., DEFAULT_PROPERTIES, apl; swmr=swmr)
+    h5open(filename, modes..., fcpl, fapl; swmr=swmr)
 end
 
 """
@@ -685,8 +711,8 @@ function h5rewrite(f::Function, filename::AbstractString, args...)
     end
 end
 
-function h5write(filename, name::String, data)
-    fid = h5open(filename, true, true, true, false, true)
+function h5write(filename, name::String, data, pv...)
+    fid = h5open(filename, "cw", pv...)
     try
         write(fid, name, data)
     finally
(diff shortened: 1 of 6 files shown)


tbole:
Fixed some more bugs for 0.6, which were probably not related to my changes. Nightlies are still broken.

musm:
Thanks! I am not sure who is best equipped to review these changes.  Most of the changes look fine to me. Perhaps @ggggggggg  also has some opinions?

Did you intend not to have the MPI tests  to run on CI?  You can add a `REQUIRE` file to the `test` folder with the `MPI`  package. (However it seems that the `MPI` package might first need your PR to be reviewed, before the HDF5 MPIO changes pass?)

src/HDF5.jl
```diff
@@ -618,34 +647,31 @@ end
 Open or create an HDF5 file where `mode` is one of:
  - "r"  read only
  - "r+" read and write
+ - "cw" read and write, create file if not existing, to not truncate
```
ggggggggg: do not truncate

src/HDF5.jl
```diff
@@ -931,7 +970,18 @@ end
...
 
 a_create(parent::Union{HDF5File, HDF5Object}, name::String, dtype::HDF5Datatype, dspace::HDF5Dataspace) = HDF5Attribute(h5a_create(checkvalid(parent).id, name, dtype.id, dspace.id), file(parent))
-p_create(class, toclose=false) = HDF5Properties(h5p_create(class), toclose)
+p_create(class, toclose=false) = HDF5Properties(h5p_create(class), toclose, class)
+function p_create(class, toclose, prop1::String, val1, pv...) 
+    p=p_create(class, toclose)
+    t = (prop1, val1, pv...)
+    checkprops(t...)
+    for i=1:2:length(t)
+        if hdf5_prop_get_set[t[i]][3] == class
```
ggggggggg: does this silently ignore non-applicable properties? should it error instead?

src/HDF5.jl
```diff
@@ -931,7 +970,18 @@ end
 t_commit(parent::Union{HDF5File, HDF5Group}, path::String, dtype::HDF5Datatype) = t_commit(parent, path, dtype, p_create(H5P_LINK_CREATE))
 
 a_create(parent::Union{HDF5File, HDF5Object}, name::String, dtype::HDF5Datatype, dspace::HDF5Dataspace) = HDF5Attribute(h5a_create(checkvalid(parent).id, name, dtype.id, dspace.id), file(parent))
-p_create(class, toclose=false) = HDF5Properties(h5p_create(class), toclose)
+p_create(class, toclose=false) = HDF5Properties(h5p_create(class), toclose, class)
+function p_create(class, toclose, prop1::String, val1, pv...)
```
ggggggggg: `p_create` appears to be a constructor for `HDF5Properties`, should we rename it `HDF5Properties`? Likewise for `a_create`

ggggggggg: The code and API here would be cleaner if we used `pv::Pair{String,<:Any}...`. Also `checkprops` would be much simpler. This is basically unrelated to your PR unless users need to provide properties for MPI use.

src/HDF5.jl
```diff
@@ -2230,6 +2287,25 @@ function h5l_get_info(link_loc_id::Hid, link_name::String, lapl_id::Hid)
     h5l_get_info(link_loc_id, link_name, info, lapl_id)
     info[]
 end
+
+"""
```
ggggggggg: Include the call signature and some info on how to use it, or where to lookup how to use it? If these aren't exported you could turn them into comments.

src/HDF5.jl
tbole: Yes, non-applicable props are ignored. The idea was to have a single list for all properties, throuhout the call tree. I don't know if it's a good idea to raise errors, since also default (empty) properties are created using this function. At the moment errors will only be raised, if the property key does not exist in the dict.

src/HDF5.jl
tbole: Would probably be a good idea, though I didn't implement that to not break the API with my changes. I don't really know to what extent other packages rely on that feature. If this is desired renaming is done quickly.

tbole: I thought of that as well, a Pait or even Dict based approach would be more cleaner definitely. But for the same reason as above, I used this approach to not break compatibility of existing routines like d_create, h5open, setindex! ,which are already using a simple parameter list.
This could eventually be changed once the improved keyword argument support in 0.7/1.0 is available.

tbole:
@musm As you already said I have postponed more complex tests until MPI.jl provides support for MPI_Info types, required by MPIO. The tests depend on the implementation in MPI.jl to provide the comm/info types while the actual implementation within HDF5.jl is generic. The current version is rather a proof of concept with the current implementation in MPI.jl based on my PR.
There has been some discussion going on, on how to implement optional features that depend on mutliple libs over at discourse using optional packages, but it seems these are still far in the future.

@ggggggggg  I mostly tried to stick with the present API style, so I do not break any third party libs. I can  of course add some documentation. I would add a list of the available features in `hdf5_prop_get_set` is important so people don't have to poke around in the sources, to find the keys.

src/HDF5.jl
tbole: Turned them to comments.

musm:
The last commit seems to cause some issues with 0.6. How about we leave the 0.7 fixes for a future PR and focus on the core changes in this PR?

I'm a little unsure about the API for `readp`. Does it warrant a new  exported function, could we  repurpose `read` with a keyword `parallel` or does that not make sense.

tbole:
Fixed that 0.7 problem. Just thought it was related to another issue I had.

The problem with `readp` (i.e. read with params) is not really MPI related, but concerns  concerning the parameter API. As already said, there's an interface clash between multiple dataset read
 
https://github.com/JuliaIO/HDF5.jl/blob/bc8a9645627cb0f844779dba5455df564823d557/src/datafile.jl#L31-L32

and passing parameters

https://github.com/JuliaIO/HDF5.jl/blob/16a9f644241a6b5a901fbb7f1899a3a51b28ddc9/src/HDF5.jl#L1291

Since the interfaces are ambiguous, this would mainly cause errors for multiple dataset read
```julia
read(h5file, "mydataset", "mykey", some-non-string-value) # works
read(h5file, "mydataset1") # works
read(h5file, "mydataset", "mykey", "some-string-value") # fails, ambiguous
read(h5file, "mydataset1", "mydataset2", "mydataset3") # fails, ambiguous
```

An alternative would be to introduce some dummy parameter in the interface like 

```julia
function read(parent::Union{HDF5File, HDF5Group}, withparams::Bool, name::String, pv...)
```

and not use `withparams` at all. Just thought introducing a new function was the cleaner solution. Any suggestions are welcome.

musm:
sans objections will merge  by weekend

src/HDF5.jl
```diff
@@ -1309,8 +1359,8 @@ end
 # Clean up string buffer according to padding mode
 function unpad(s::String, pad::Integer)
     if pad == H5T_STR_NULLTERM
-        v = search(s, '\0')
-        v == 0 ? s : s[1:v-1]
+        v = Compat.findfirst(equalto('\0'), s)
```
musm: Does this not need  Compat v0.55.2 as a minimum?

musm: Maybe just add the import `Compat: StingVector, findfirst` to the top of the file like `StringVector`? (not required, but I just thought  it might be cleaner, when these imports are dropped for 0.6)

musm:
Breaking improvements to the API can come in future PRs

However it might be annoying to add and export the function `readp` and then unexport it after fixing /improving the  the  interface.

src/HDF5.jl
tbole: I've bumped the Compat version to v0.56 (this will probably be required either way, once 0.7 arrives) and explicitly listed all uses of Compat.

tbole:
Ok, so I've removed `readp` and added a dummy parameter to the interface of read as suggested above.
Somewhat less clutter.

```julia
function read(parent::Union{HDF5File, HDF5Group}, name::String, withparams::Bool=true, pv...)
```

src/HDF5.jl
musm: I think I quoted the wrong minimum. It should be  v0.53.0https://github.com/JuliaLang/Compat.jl/commit/8cca897ff1bf251bc2fb6e806ec481cfc478de26    I updated the PR.  With minimums we need to be sure to put the minimums, so to not unnecessarily force and update on users (best practice).

Thanks again

musm:
Ok great. We can get this through. @ggggggggg and @tbole  for improving the interface a separate PR would be best, which will probably be  breaking; but I think it would be for the better and also it would be nice to have it before julia 1.0

src/HDF5.jl
```diff
@@ -73,6 +73,13 @@ const Hssize  = Int64
 const Htri    = Cint   # pseudo-boolean (negative if error)
 const Haddr   = UInt64
 
+# MPI communicators required by H5P
+abstract  type Hmpih end
+primitive type Hmpih32 <: Hmpih 32 end # MPICH C/Fortran, OpenMPI Fortran: 32 bit handles
+primitive type Hmpih64 <: Hmpih 64 end # OpenMPI C: pointers (mostly 64 bit)
+const mpihandles = Dict(4 => Hmpih32, 8 => Hmpih64)
+h5_mpihandle(handle) = reinterpret(mpihandles[sizeof(handle)],handle)
```
musm: should the input here be  `handle::Integer`  ?

musm:
There seems to be an unrelated download failure on 32 bit windows otherwise lgtm

src/HDF5.jl
tbole: This depends on the caller. In case of MPI.jl the handle will be a 4/8 byte `primitive type Ccomm` for the C communicator either wrapping a 4 byte integer handle (MPICH) or a 4/8 byte pointer (OpenMPI). I wanted to have this as generic as possible. This way users do not even have to use MPI.jl, but can just pass the pointers/handles without conversion.

tbole:
@musm A better interface would be nice, but since I don't know when 1.0 (or at least an alpha) will actually land and what nice features it will bring, I wasn't to keen on making big changes here. Though the interface could be really improved using the new features for Dicts and named tuples.

musm:
I am not sure either when it will be released (probably not for a few months), but I just thought I would mention that breaking the API to improve it  is something we can push for.

musm:
I've opened an issue over at `MPI.jl` to get a tag release. That way we can check the MPI tests on CI before merge.

test/mpio.jl
```diff
@@ -0,0 +1,68 @@
...
+using MPI
+using HDF5
+using Base.Test
+
+MPI.Init()
+info=MPI.Info()
+
+h = HDF5.mpihandles[sizeof(MPI.CComm)]
+if MPI.HAVE_MPI_COMM_C2F
+    ccomm=ccall(:MPI_Comm_f2c, h, (Cint,), MPI.COMM_WORLD.val)
```
musm: unrelated but does `MPI.jl` not provide an interface to this and below?

src/HDF5.jl
```diff
@@ -2,10 +2,10 @@ __precompile__()
 
 module HDF5
 
-using Compat
+using Compat: 
+    AbstractRange, @compat, equalto, StringVector, findfirst
```
musm: `@compat` isn't necessary here

src/HDF5.jl
```diff
@@ -1309,8 +1354,8 @@ end
 # Clean up string buffer according to padding mode
 function unpad(s::String, pad::Integer)
     if pad == H5T_STR_NULLTERM
-        v = search(s, '\0')
-        v == 0 ? s : s[1:v-1]
+        v = findfirst(equalto('\0'), s)
+        v == nothing ? s : s[1:v-1]
```
musm: typically better to use `v === nothing`  when comparing to `nothing`

src/HDF5.jl
```diff
@@ -931,7 +970,18 @@ end
 t_commit(parent::Union{HDF5File, HDF5Group}, path::String, dtype::HDF5Datatype) = t_commit(parent, path, dtype, p_create(H5P_LINK_CREATE))
 
 a_create(parent::Union{HDF5File, HDF5Object}, name::String, dtype::HDF5Datatype, dspace::HDF5Dataspace) = HDF5Attribute(h5a_create(checkvalid(parent).id, name, dtype.id, dspace.id), file(parent))
-p_create(class, toclose=false) = HDF5Properties(h5p_create(class), toclose)
+p_create(class, toclose=false) = HDF5Properties(h5p_create(class), toclose, class)
+function p_create(class, toclose, prop1::String, val1, pv...) 
+    p=p_create(class, toclose)
```
musm: spaces around `=` here and in the rest, would be nice for readability

src/HDF5.jl
```diff
@@ -549,6 +566,18 @@ end
 # Blosc compression:
 include("blosc_filter.jl")
 
+function checkprops(pv...)
+    if !iseven(length(pv))
+        error("Properties and values must come in pairs")
+    end
+    for i=1:2:length(pv)
+        if !isa(pv[i], AbstractString)
+            error("Keyword argument $i of key/value pairs should be a String, but it's a ", typeof(pv[i]))
```
musm: I think these would be better as   `throw(ArgumentError....`

musm:
rebase?

tbole:
Fixed all comments except for @compat. This is actually required by `datafile.jl` which is included below.

tbole:
I'll rebase it to the current changes then.

tbole:
Rebased, cleaned up and ready to go!

src/HDF5.jl
```diff
@@ -958,35 +1008,27 @@ function setindex!(p::HDF5Properties, val, name::String)
...
+    dcpl = p_create(H5P_DATASET_CREATE)
+    dxpl = p_create(H5P_DATASET_XFER, false, t...)
+    dapl = p_create(H5P_DATASET_ACCESS, false, t...)
+
+    need_chunks = any(t[i] in chunked_props for i=1:2:length(t))
+    have_chunks = any(t[i] == "chunk"       for i=1:2:length(t))
+    chunk = need_chunks ? heuristic_chunk(val) : Int[]
+    discard_chunks = need_chunks && isempty(chunk)
+    if need_chunks && !have_chunks && !discard_chunks
+        dcpl["chunk"] = chunk
```
musm: is it not possible to remove chunk on line 1019
and just put 
`dcpl["chunk"] = chunk`
I realize in the old code `chunk = heuristic_chunk(val)` always?

musm:
Thanks!

src/HDF5.jl
tbole: We cannot just set `dcpl`, because it could a) be already preset by the user and b) we need to ensure that chunk is not empty (`heuristic_chunk` can return empty chunks as well).

PR #463, opened 2018-03-16
support SharedArrays

bjarthur:
is there a reason not to support SharedArrays?  all this PR does is to use type `AbstractArray` instead of `Array` in select function signatures.  includes a test.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1131,7 +1131,7 @@ datatype(dset::HDF5Attribute) = HDF5Datatype(h5a_get_type(checkvalid(dset).id), 
 # Create a datatype from in-memory types
 datatype(x::T) where {T<:HDF5Scalar} = HDF5Datatype(hdf5_type_id(T), false)
 datatype(::Type{T}) where {T<:HDF5Scalar} = HDF5Datatype(hdf5_type_id(T), false)
-datatype(A::Array{T}) where {T<:HDF5Scalar} = HDF5Datatype(hdf5_type_id(T), false)
+datatype(A::AbstractArray{T}) where {T<:HDF5Scalar} = HDF5Datatype(hdf5_type_id(T), false)
 function datatype(str::S) where {S<:String}
     type_id = h5t_copy(hdf5_type_id(S))
     h5t_set_size(type_id, max(sizeof(str), 1))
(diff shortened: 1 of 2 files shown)


bjarthur:
thx for the approval!

musm:
lgmt. 

undefined and null tests are failing on windows any reason to suspect this PR? I don't think they were before

bjarthur:
egads.  well, the "undefined and null" testset is where i inadvertently placed the test for this PR.  do SharedArrays perhaps not work on windows?  i can't test until monday as i only have access to a windows box at work.

bjarthur:
the problem appears to have nothing to do with SharedArrays, but rather that `h5write` behaves differently on windows vs. OS X depending on whether the filename was generated with `tempdir()` or `tempname()`:

on windows:

```
julia> h5write(tempdir()*"asdfasdfasdf.tmp", "/data", rand(3,4,5))     # works!

julia> h5write(tempname(), "/data", rand(3,4,5))                       # doesn't work :(
ERROR: This does not appear to be an HDF5 file
Stacktrace:
 [1] #h5open#3(::Bool, ::Function, ::String, ::Bool, ::Bool, ::Bool, ::Bool, ::Bool, ::HDF5.HDF5Properties, ::HDF5.HDF5Properties) at C:\Users\arthurb\.julia\v0.6\HDF5\src\HDF5.jl:598
 [2] h5write(::String, ::String, ::SharedArray{UInt16,3}) at C:\Users\arthurb\.julia\v0.6\HDF5\src\HDF5.jl:687

julia> tempname()
"C:\\Users\\arthurb\\AppData\\Local\\Temp\\jl_79FF.tmp"
```

vs. on OS X:

```
julia> h5write(tempdir()*"asdfasdfasdf.tmp", "/data", rand(3,4,5))    # works!

julia> h5write(tempname(), "/data", rand(3,4,5))                      # works!

julia> tempname()
"/var/folders/x6/gf97mhtj7kd967vg252ygy4nmsxs4y/T/juliaE1apIA"
```

ideas why this would be the case?

musm:
Even simpler
```julia

julia> f1 = tempdir()*"adf.tmp"
"C:\\Users\\Mus\\AppData\\Local\\Temp\\adf.tmp"

julia> f2 = tempname()
"C:\\Users\\Mus\\AppData\\Local\\Temp\\jl_4E2F.tmp"

julia> ishdf5(f1)
true

julia> ishdf5(f2)
false

```

musm:
In the first case   windows does not create the file `f1` in the second case  windows *does* create the file `f2` and thus `ishdf5(f2)` fails

musm:
ref https://github.com/JuliaLang/julia/pull/25530

PR #464, opened 2018-03-19
Fix shared array test

musm:
ref https://github.com/JuliaIO/HDF5.jl/pull/463


bjarthur:
thanks!

bjarthur:
also, while working on my own fix for this, i noticed what i think is a mistake in the tests.  the file `fn` actually exists at the time for [this test](https://github.com/JuliaIO/HDF5.jl/blob/master/test/plain.jl#L357), when the test purports to test what happens if it doesn't.

musm:
agreed that looks like an error

musm:
mind opening a PR for that fix?

PR #465, opened 2018-03-20
fix 'file does not already exist' test

bjarthur:
i also took the liberty of moving the SharedArray test into the "plain" testset.


bjarthur:
failing only on nightlies, but that's okay, right?

PR #467, opened 2018-03-22
Call constructors directly instead of convert

musm:


src/blosc_filter.jl
```diff
@@ -46,10 +46,10 @@ function blosc_set_local(dcpl::Hid, htype::Hid, space::Hid)
                   dcpl, 32, blosc_chunkdims);
     chunksize = prod(resize!(blosc_chunkdims, ndims))
     if ndims < 0 || ndims > 32 || chunksize > Blosc.MAX_BUFFERSIZE
-        return convert(Herr, -1)
+        return Herr(-1)
     end
     htypesize = ccall((:H5Tget_size,libhdf5), Csize_t, (Hid,), htype)
-    htypesize == 0 && returnconvert(Herr, -1)
+    htypesize == 0 && return Herr(-1)
```
musm: bug fix here

PR #469, opened 2018-03-25
support AbstractArrays

bjarthur:


src/HDF5.jl
```diff
@@ -1614,7 +1614,7 @@ for (privatesym, fsym, ptype) in
         end
         # Scalar types
         ($fsym)(parent::$ptype, name::String, data::Union{T, AbstractArray{T}}, plists...) where {T<:ScalarOrString} =
-            ($privatesym)(parent, name, data, plists...)
+            ($privatesym)(parent, name, typeof(data)<:AbstractArray ? collect(data) : data, plists...)
```
musm: You can just use `data isa AbstractArray` or `isa(data,AbstractArray)` here

bjarthur:
do you know why travis is called twice?  once labeled pr the other labelled push.  Gadfly does this as well.  i think it's not necessary and would like to turn it off there.

musm:
> do you know why travis is called twice? once labeled pr the other labelled push. Gadfly does this as well. i think it's not necessary and would like to turn it off there.


Not sure about that

bjarthur:
curious too that the tests pass on the linux nightly but not mac or windows.

musm:
There is  a winrpm bug on 0.7 right now

musm:
I feel like this should also  be mentioned in the readme.  It's a little concerning to do the wholesale conversion automatically. Perhaps the conservative approach where the user is forced to call `collect` is better/safer.

musm:
anyone else have thoughts on this?

bjarthur:
i could go either way.  it's more Julian i guess to force the user to call collect, and would be less maintenance for HDF5.jl.  on the other hand, it is convenient.

would need to revert https://github.com/JuliaIO/HDF5.jl/pull/464 and https://github.com/JuliaIO/HDF5.jl/pull/463 too.  is that as easy as pushing the "revert" button on github?  never done that.

could also consider revoking my push privileges.  i'm busy enough with Gadfly so am happy to give up that responsibility, which seemingly i overstepped here.

musm:
I feel more comfortable, without wholesale conversion. I do however understand the convenience of the current approach. However, since HDF5 does not save type information I think it is reasonable to explicitly force the user to convert to an array, so they understand that HDF5 does not have the ability to retain type information, in which case they should use JLD/JLD2.

musm:
> would need to revert #464 and #463 too. is that as easy as pushing the "revert" button on github? never done that.

should be that simple

bjarthur:
i'm fine with that.  go for it!

musm:
we can open up a further issue to discuss and obtain other opinions in the meantime

PR #472, opened 2018-04-04
revert support AbstractArrays

musm:
Reverts JuliaIO/HDF5.jl#469
Reverts JuliaIO/HDF5.jl#464
Reverts JuliaIO/HDF5.jl#463


bjarthur:
thanks for doing this!  in hindsight i whole-heartedly agree that making the user call `collect` is more Julian.

PR #480, opened 2018-05-11
condition H5F_LIBVER_LATEST and friends on libversion

### LINKED ISSUE #474: H5F_LIBVER_ LATEST is hardcoded, but the value changed
The value of `H5F_LIBVER_LATEST` appears to have changed. Currently the values are hard coded, but it seems better to read them from the HDF5 library. Can anybody tell me how to do this, and I'd be happy to submit a PR. There are lots of examples of reading constants with the `HDF5.read_constant` function, but I don't understand it enough to adapt it to this case.

```c++
/* Library's format versions */
typedef enum H5F_libver_t {
    H5F_LIBVER_ERROR = -1,
    H5F_LIBVER_EARLIEST = 0,    /* Use the earliest possible format for storing objects */
    H5F_LIBVER_V18 = 1,         /* Use the latest v18 format for storing objects */
    H5F_LIBVER_V110 = 2,        /* Use the latest v10 format for storing objects */
    H5F_LIBVER_NBOUNDS
} H5F_libver_t;

#define H5F_LIBVER_LATEST   H5F_LIBVER_V110
```

This causes the SWMR tests to fail on my machine. I'm not sure why they aren't failing on travis on OSX, maybe it's installing a different version of the HDF5 library? I have `1.10.2`

ggggggggg:
Fix #474

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -252,8 +252,15 @@ const H5T_NATIVE_UINT64   = read_const(:H5T_NATIVE_UINT64_g)
 const H5T_NATIVE_FLOAT    = read_const(:H5T_NATIVE_FLOAT_g)
 const H5T_NATIVE_DOUBLE   = read_const(:H5T_NATIVE_DOUBLE_g)
 # Library versions
-const H5F_LIBVER_EARLIEST = 0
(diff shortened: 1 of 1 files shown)


musm:
thnx

PR #484, opened 2018-06-15
add exported function set_blosc_property

cnliao:
for users to select compressor/shuffle method.

function signature:
set_blosc_property(;shuffle=:none,compressor=:zstd)
compressor is one of :blosclz, :lz4, :lz4hc, :snappy, :zlib, :zstd
shuffle is one of :none, :normal, :bit

Users should call it before d_create, e.g.:
set_blosc_property(;shuffle=:none,compressor=:zstd)
dset = d_create(g, "B", datatype(Float64), dataspace(1000,100,10), "blosc", 5)

I need this function. Hopefully others will find it helpful.

### DIFF
--- a/src/blosc_filter.jl
+++ b/src/blosc_filter.jl
@@ -145,3 +145,11 @@ function h5p_set_blosc(p::HDF5Properties, level::Integer=5)
     status < 0 && error("Error setting blosc compression level")
     nothing
 end
+
+export set_blosc_property
+function set_blosc_property(;shuffle=:none,compressor=:zstd)
+    _set_blosc_values[6] = _hdf5_blosc_shuffle[shuffle]
+    _set_blosc_values[7] = _hdf5_blosc_compressor[compressor]
+end
+const _hdf5_blosc_shuffle = Dict(:none=>0, :normal=>1, :bit=>2)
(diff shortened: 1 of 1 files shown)


cnliao:
I just realized that the blosc_filter in HDF5.jl does not really respect cd_values like the blosc_filter.c it is based on.
Instead it calls Blosc.jl.
So to change the filter compressor you have to call Blosc.set_compressor.
But now the cd_values used to create the dataset does not reflect the compressor blosc is really using. This causes tools like h5ls to display wrong infomation.

So one can use:
function set_blosc_property(;shuffle=:none,compressor=:zstd)
    _set_blosc_values[6] = _hdf5_blosc_shuffle[shuffle]
    _set_blosc_values[7] = _hdf5_blosc_compressor[compressor]
    Blosc.set_compressor(string(compressor))
end

But this feels even more patchy. I am fine with it now. If it really makes you nervous maybe you should try:
1. Fix blosc_filter, set_compressor really should be called there;
2. Fix h5p_set_blosc so that it can set shuffle/compressor in addition to level;
3. Announce that d_create has a breaking change for the optional property "blosc".

Best

musm:
Closed by  https://github.com/JuliaIO/HDF5.jl/pull/880 

You can pass in these custom flags in the filter constructor, see: https://github.com/JuliaIO/HDF5.jl/blob/master/filters/H5Zblosc/src/H5Zblosc.jl#L137

PR #488, opened 2018-06-28
Relax test tolerances

musm:


musm:
This fixes the remote tests tolerances being too tight on 0.7 beta

PR #489, opened 2018-07-01
Fix Pkg3 usage

musm:


musm:
fixes master issues  win 32 master seems to have some unrelated issues and macos has homebrew issues.

Issue #483, opened 2018-06-06
Issue with using HDF5 and Pkg.Build("HDF5")


vd2309:
Hello!

I have this code to run:

```julia
**using HDF5
filename = "/Untitled/Users/philip/Desktop/test.h5"
img_A = h5read(filename, "img_A")
img_B = h5read(filename, "img_B")
ans_p_1step_prevfrm = h5read(filename, "ans_p_1step_prevfrm")
ans_p_1step = h5read(filename, "ans_p_1step")
ans_p_2step = h5read(filename, "ans_p_2step")
stack = h5read(filename, "refernece")
stack_hi = h5read(filename, "refernece_highpass");
img_phase = h5read(filename, "img_phase");
img_width, img_height = size(img_A)
zStkCnt = size(stack,3);**


when I run it it says:
**INFO: Precompiling module HDF5.
ERROR: LoadError: HDF5 not properly installed. Please run Pkg.build("HDF5")
Stacktrace:
 [1] error(::String) at ./error.jl:21
 [2] include_from_node1(::String) at ./loading.jl:576
 [3] include(::String) at ./sysimg.jl:14
 [4] anonymous at ./<missing>:2
while loading /Users/philip/.julia/v0.6/HDF5/src/HDF5.jl, in expression starting on line 38
ERROR: Failed to precompile HDF5 to /Users/philip/.julia/lib/v0.6/HDF5.ji.
Stacktrace:
 [1] compilecache(::String) at ./loading.jl:710
 [2] _require(::Symbol) at ./loading.jl:497
 [3] require(::Symbol) at ./loading.jl:405**


...so i include Pkg.build("HDF5") in the code I get this error:



**INFO: Building Homebrew
Already up-to-date.
INFO: Building CMakeWrapper
INFO: Building Blosc
INFO: Building HDF5
==> Installing gcc from staticfloat/juliatranslated
Error: The following formula:
  gcc
cannot be installed as a binary package and must be built from source.
Install the Command Line Tools:
  xcode-select --install

================================[ ERROR: HDF5 ]=================================

LoadError: failed process: Process(`/Users/philip/.julia/v0.6/Homebrew/deps/usr/bin/brew install --ignore-dependencies staticfloat/juliatranslated/gcc`, ProcessExited(1)) [1]
while loading /Users/philip/.julia/v0.6/HDF5/deps/build.jl, in expression starting on line 41

================================================================================

================================[ BUILD ERRORS ]================================

[...]

PR #492, opened 2018-07-23
Fix depwarns on julia 0.7

carlobaldassi:
With this, tests pass (locally) without deprecation warnings on both 0.6 and 0.7.
It also fixes a warning about using Mmap, which doesn't appear in HDF5 tests but appears when running JLD tests.

It also enables CI testing on 0.7.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -3,6 +3,7 @@ __precompile__()
 module HDF5
 
 using Compat
+using Compat.Mmap
 using Compat: findfirst
 
 using Base: unsafe_convert, StringVector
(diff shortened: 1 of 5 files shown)


carlobaldassi:
Travis failure on OS X (all versions) is #483, unrelated to this PR.
The Appveyor failures on Win32 (julia 0.7) are also unrelated, it's hanging on the "build LibCURL" step.

test/runtests.jl
```diff
@@ -1,6 +1,9 @@
 using HDF5
 using Compat.Test
 using Compat.Distributed
+@static if VERSION ≥ v"0.7.0-DEV.3637"
```
musm: Is the `@static` is necessary here since it's in global scope?

carlobaldassi: As far as I understand, `@static` means "evaluate the conditional at parse time". It shouldn't be related to the scope.

src/HDF5.jl
```diff
@@ -1969,7 +1982,7 @@ end
 ### Convenience wrappers ###
 # These supply default values where possible
 # See also the "special handling" section below
-h5a_write(attr_id::Hid, mem_type_id::Hid, buf::String) = h5a_write(attr_id, mem_type_id, Vector{UInt8}(buf))
+h5a_write(attr_id::Hid, mem_type_id::Hid, buf::String) = h5a_write(attr_id, mem_type_id, Vector{UInt8}(codeunits(buf)))
```
musm: isn't it better to use `unsafe_wrap` here instead to avoid the extra allocation?

carlobaldassi: Done.

PR #493, opened 2018-08-08
Workaround for mmap issues with Julia 0.7

tknopp:
This is a workaround for the alignment issue within Julia 0.7 and higher. I ask @Keno: Is this the official workaround? Will the `reinterpret` be fast in the future? @vtjnash indicated in https://github.com/JuliaLang/julia/issues/28424 that `reinterpret` is slow and in turn this workaround should not be baked into the `mmap` function directly (where it IMHO belongs).


vtjnash:
This is correct. This is as fast as it can possibly be while still using mmap instead of read.

PR #494, opened 2018-08-09
Implement julia 0.7 iteration protocol

carlobaldassi:
The old start/next/done protocol was deprecated during the beta phase.
(This also adds deps/build.log in .gitignore while at it.)

When this is merged it would be a good idea to tag a new release I guess.

### DIFF
--- a/deps/.gitignore
+++ b/deps/.gitignore
@@ -2,3 +2,4 @@ builds
 src
 deps.jl
 downloads
+build.log
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -9,9 +9,15 @@ using Compat: findfirst
 using Base: unsafe_convert, StringVector
 
 import Base:
-    close, convert, done, eltype, endof, flush, getindex, ==,
-    isempty, isvalid, length, names, ndims, next, parent, read,
-    setindex!, show, size, sizeof, start, write, isopen
+    close, convert, eltype, endof, flush, getindex, ==,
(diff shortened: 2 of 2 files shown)


src/HDF5.jl
```diff
@@ -1144,19 +1150,29 @@ function names(x::HDF5Attributes)
...
+    end
+    function next(parent::Union{HDF5File, HDF5Group}, iter)
+        iter[2] = h5object(h5o_open_by_idx(checkvalid(parent).id, ".", H5_INDEX_NAME, H5_ITER_INC, iter[1]-1, H5P_DEFAULT), parent)
+        iter[1] += 1
+        (iter[2], iter)
+    end
+else
+    function iterate(parent::Union{HDF5File, HDF5Group}, iter = (1,nothing))
+        n, prev_obj = iter
+        prev_obj ≢ nothing && close(prev_obj)
```
musm: this symbol is equivalent to `!(obj === nothing)`?

deps/.gitignore
```diff
@@ -2,3 +2,4 @@ builds
 src
 deps.jl
 downloads
+build.log
```
musm: why is this necessary?

musm:
great thanks. We should probably drop 0.6 support henceforth

src/HDF5.jl
garrison: yes (or, equivalently, `!==`)

PR #498, opened 2018-08-16
Update for 0.7 and drop earlier releases

musm:


andreasnoack:
I'd recommend not adding a project file yet since CIBot can't handle project files yet. Except for that, I think it would great to have this merged soon since the Mac/windows problem are unrelated to this PR.

musm:
@andreasnoack  It's not clear to me why we cannot add project files to this repo. Is there something special about HDF5 and CIBot? Any references or issue tracker. Thanks.

musm:
@andreasnoack let me know if the Project file will be a problem and I can open a further PR to remove it.

andreasnoack:
It is a problem but it has nothing to do with HDF5. CIBot still uses the old package manager and it fails when there are project files in the packages. Hence, we have been removing project files from many packages lately.

andreasnoack:
E.g. https://github.com/JuliaGraphics/Colors.jl/pull/323 and https://github.com/JuliaLang/Compat.jl/pull/624

musm:
thnx ref https://github.com/JuliaIO/HDF5.jl/pull/505

Issue #500, opened 2018-08-17
Errors when installing HDF5 v0.9.4


nhavt:
Hi there, 
When using the HDF5 v0.9.4 module, I have gotten errors which you can see in the attached file
[Error HDF5.txt](https://github.com/JuliaIO/HDF5.jl/files/2296628/Error.HDF5.txt)
. Could anyone please look at it and let me know how to fix this issue? I am using Julia 1.0.0 under Windows 10. Thanks in advance!
Nha

musm:
You need to update the packages

nhavt:
Hi there, thanks for your comment. However, I didn't get it. Do you mean by running the command Pkg.update()? Or could you please tell me a bit more? Thanks in advance! Nha

musm:
Yea, run` import Package; Package.update()`

musm:
Hmm I see u have the latest version. This error should not be there.

musm:
I think I see the problem

nhavt:
Even I ran both commands Pkg.update() and import HDF5, I still got the same problem :(

musm:
There's a missing deprecation, I have a PR to fix this. Please be patient until we tag a new release with the fix.

nhavt:
Ok, thanks very much for looking at this issue!

musm:
Actually there are downstream issues related to LibExpat.jl that prevent hdf5 building on windows on Julia 1.

daxinnan:
I also meet one problem when building HDF5
`(v0.7) pkg> add HDF5
 Resolving package versions...
 Installed WinRPM ────────── v0.4.0
 Installed Parameters ────── v0.9.2
 Installed LibExpat ──────── v0.4.2
 Installed HTTPClient ────── v0.2.1
 Installed BufferedStreams ─ v1.0.0
 Installed Homebrew ──────── v0.7.0
 Installed HDF5 ──────────── v0.9.4
 Installed LibCURL ───────── v0.3.1
 Installed CMakeWrapper ──── v0.2.2
 Installed CMake ─────────── v1.0.1
 Installed Libz ──────────── v1.0.0
 Installed BinDeps ───────── v0.8.10
 Installed Blosc ─────────── v0.5.1
  Updating `C:\Users\lxc\.julia\environments\v0.7\Project.toml`
  [f67ccb44] + HDF5 v0.9.4
  Updating `C:\Users\lxc\.julia\environments\v0.7\Manifest.toml`
[...]

PR #501, opened 2018-08-17
Fix endof deprecation

musm:
close https://github.com/JuliaIO/HDF5.jl/issues/500


musm:
passes on Linux, MacOS is not working right now.

musm:
Merging this quicker than usual since it is causing issues with users being able to use this package

Issue #508, opened 2018-09-02
H5Oget_info error with HDF library 1.10.3


BoundaryValueProblems:
Hello. It looks to me that HDF5.jl v0.10.0 is working well on Julia v0.7.0.
However, on Julia v0.6.4, I'm using HDF5.jl v0.9.4, and Pkg.test("HDF5") failed with the error:
```
julia> Pkg.test("HDF5")
INFO: Testing HDF5
HDF5 version 1.10.3
plain: Error During Test
  Got an exception of type ErrorException outside of a @test
  ccall: could not find function H5Oget_info in library /Users/xxx/.julia/v0.6/Homebrew/deps/usr/lib/libhdf5.dylib
```
Interesting enough, the HDF5 library used on v0.6.4 is 1.10.3 while that used on v0.7.0 is 1.10.2. I'm wondering if this is the source of this problem or not. (This might be the case since I ran ```Homebrew.update()``` on v0.6.4 while on v0.7.0, Homebrew.jl has problems and cannot be updated... Any help would be highly appreciated!

BoundaryValueProblems:
I just wanted to add some more info. I just recovered ~/.julia/v0.6/Homebrew one week ago using Time Machine, which was the previous version of HDF5 library 1.10.2_1. Then, ```Pkg.test("HDF5")``` passed!  So, this is a warning. If you try to update the Homebrew under Julia v0.6.x, then the HDF5 library would be updated to 1.10.3 and you may encounter the same problem as I had. I recall when I updated Homebrew today, the HDF5 library was configured and compiled instead of copying the prebuilt library, which might have caused this problem although it installed the HDF5 library without any errors.

BoundaryValueProblems:
Additional info:
* On HDF5 library v1.10.2, libhdf5.dylib contains the functions: H5Oget_info; H5Oget_info_by_idx; H5Oget_info_by_name.
* On HDF5 library v1.10.3, libhdf5.dylib contains the functions: H5Oget_info1; H5Oget_info2; H5Oget_info_by_idx1; H5Oget_info_by_idx2; H5Oget_info_by_name1; H5Oget_info_by_name2.
So, if we want to use the newer version of HDF5 library, then all the codes calling H5Oget_info must be replaced by H5Oget_info1 or 2 (I don't know yet what's the difference between them).

jeanpauphilet:
[...]

PR #510, opened 2018-09-04
 Correct H50get_info deprecation for earlier releases

musm:
close https://github.com/JuliaIO/HDF5.jl/issues/508

by retaining old function behavior in H50get_info

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,5 +12,7 @@ addons:
   apt:
     packages:
     - hdf5-tools
+before_script:
+    - if [ $TRAVIS_OS_NAME = osx ]; then brew install gcc; brew link --overwrite gcc; brew install hdf5; fi
 after_success:
   - julia -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder()); Codecov.submit(Codecov.process_folder())'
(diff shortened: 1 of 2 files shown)


musm:
merging sooner than usual since this is uncontroversial and causing headaches to users.

BoundaryValueProblems:
@musm , unfortunately, I won't be able to test this latest version since REQUIRE file specifies it as Julia v0.7. Many of us, due to the problem of Homebrew under Julia v0.7, the common situation, at least on macOS is the following:
* Julia v0.6.4 + HDF5.jl v0.9.4 + HDF5 library 1.10.3 (if Homebrew.update() was run);
* Julia v0.7/1.0 + HDF5.jl v0.10.1 + HDF5 library 1.10.2 (since Homebrew.update() does not work).

In other words, I could test HDF5.jl v0.10.1 + HDF5 library 1.10.2 combination, and Pkg.test passed. However, I still cannot test the combination of HDF5.jl v0.10.1 + HDF5 library 1.10.3 at this point due to the problem of Homebrew.jl under Julia v0.7/1.0.
Thanks!

BoundaryValueProblems:
@musm Thanks a lot! I just upgraded to HDF5.jl v0.9.5, and ran the Pkg.test("HDF5") on my MacBook Pro with the following setup:
* Julia v0.6.4 + HDF5.jl v0.9.5 + HDF5 library 1.10.3

 And the tests passed!!

PR #511, opened 2018-09-05
wrap a subset of H5TB

cnliao:
The HDF5 Table API defines a standard storage for HDF5 datasets that are intended to be interpreted as tables.
https://support.hdfgroup.org/HDF5/doc/HL/RM_H5TB.html


musm:
This seems to cause EXCEPTION_ACCESS_VIOLATION on 32-bit windows.

cnliao:
> This seems to cause EXCEPTION_ACCESS_VIOLATION on 32-bit windows.

The only thing I can come up with is the hard-coded record size in the tests, but it turns out fixing that does not help. It looks like an upstream bug. Unfortunately I don't have access to a 32-bit windows platform so it is unlikely that I will fix this.

musm:
resurected in https://github.com/JuliaIO/HDF5.jl/pull/777

test/h5tb.jl
```diff
@@ -0,0 +1,51 @@
...
+            HDF5.h5tb_read_records(f.id, name, 2, 3, recsize, offset, fieldsize, buf)
+            @test buf[1:6] == collect(23:28)
+
+            _nfield = Ref{HDF5.Hsize}(0)
+            _nrec = Ref{HDF5.Hsize}(0)
+            HDF5.h5tb_get_table_info(f.id, name, _nfield, _nrec)
+            @test _nfield[] == nfield
+            @test _nrec[] == 6
+
+            _colnamebuf = [fill(0xff, 10), fill(0xff,10)]
```
musm, line 39: Is there no way to determine the length of these strings from HDF5?

Issue #512, opened 2018-09-06
HDF5 + Mmap failed with Julia0.7/1.0


bluehope:
Simple HDF5 based Mmap example `HDF5test.jl`

```julia
using HDF5
hdf_cache_name = "test.hdf5"
# Write  HDF5 file
fid_hdf = h5open(hdf_cache_name,"w");
hdf5_A = d_create(fid_hdf,"A",datatype(Int64),dataspace(3,3));
A_rand = rand(Int64,3,3)
hdf5_A[:,:] = A_rand

flush(fid_hdf);
close(fid_hdf);
# Read HDF5 file & MMAP
fid_hdf = h5open(hdf_cache_name,"r");
A_mmaped = readmmap(fid_hdf["A"]);

println(A_mmaped .- A_mmaped)
```
However, this fails as bellow.
```julia
julia_1.0 HDF5test.jl
ERROR: LoadError: UndefVarError: Mmap not defined
Stacktrace:
 [1] readmmap(::HDF5Dataset, ::Type{Array{Int64,N} where N}) at /Users/bluehope/.julia/dev/HDF5/src/HDF5.jl:1613
 [2] readmmap(::HDF5Dataset) at /Users/bluehope/.julia/dev/HDF5/src/HDF5.jl:1621
 [3] top-level scope at none:0
 [4] include at ./boot.jl:317 [inlined]
 [5] include_relative(::Module, ::String) at ./loading.jl:1038
 [6] include(::Module, ::String) at ./sysimg.jl:29
 [7] exec_options(::Base.JLOptions) at ./client.jl:229
 [8] _start() at ./client.jl:421
in expression starting at /Users/bluehope/Dropbox/shared/DFTforge.jl/src/HDF5test.jl:13
```

I guess this could be fixed by adding [Mmap](https://docs.julialang.org/en/v1/stdlib/Mmap/index.html) 
```julia
using Mmap
```
at `HDF5.jl`.

PR #515, opened 2018-09-06
import Mmap

tknopp:
this otherwise breaks in Julia 1.0

fixes https://github.com/JuliaIO/HDF5.jl/issues/512


musm:
Could you add `Mmap` to `_Project.toml` under `[deps]` for future migration?

musm:
Thanks!  This could also  be a great opportunity to add test. Perhaps just  use https://github.com/JuliaIO/HDF5.jl/issues/512

musm:
Thank you and I have published a release

PR #520, opened 2018-10-05
Allow AbstractArray writing

nicoleepp:
On 1.0 calling `reinterpret` on arrays returns a `ReinterpretArray` which is a subtype of AbstractArray and fails with the current master of HDF5. This allows those arrays to be written.

### DIFF
--- a/test/plain.jl
+++ b/test/plain.jl
@@ -416,16 +416,17 @@ end # testset null and undefined
 # test writing reinterpreted data
 @testset "reinterpreted data" begin
 fn = tempname()
+try
+    h5open(fn, "w") do f
+        data = reinterpret(UInt8, [true, false, false])
+        write(f, "reinterpret array", data)
+    end
 
-h5open(fn, "w") do f
-    data = reinterpret(UInt8, [true, false, false])
-    write(f, "reinterpret array", data)
+    @test h5open(fn, "r") do f
+        read(f, "reinterpret array")
+    end == UInt8[0x01, 0x00, 0x00]
+finally
+    rm(fn)
 end
-
-@test h5open(fn, "r") do f
-    read(f, "reinterpret array")
-end == UInt8[0x01, 0x00, 0x00]
-
-rm(fn)
 
 end # writing reinterpreted data
--- a/test/plain.jl
+++ b/test/plain.jl
@@ -226,11 +226,11 @@ z = read(fr, "Float64", "Int16")
 @test typeof(z) == Tuple{Float64,Int16}
 # Test function syntax
 read(fr, "Float64") do x
-	@test x == 3.2
+    @test x == 3.2
 end
 read(fr, "Float64", "Int16") do x, y
-	@test x == 3.2
-	@test y == 4
+    @test x == 3.2
+    @test y == 4
 end
 # Test reading entire file at once
 z = read(fr)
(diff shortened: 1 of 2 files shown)


musm:
can you add a test for the reinterpret array setting, thanks.

nicoleepp:
Added a test that fails on master

nicoleepp:
bump

test/plain.jl
```diff
@@ -412,3 +412,20 @@ close(f)
...
+h5open(fn, "w") do f
+    data = reinterpret(UInt8, [true, false, false])
+    write(f, "reinterpret array", data)
+end
+
+@test h5open(fn, "r") do f
+    read(f, "reinterpret array")
+end == UInt8[0x01, 0x00, 0x00]
+
+rm(fn)
```
omus: Not necessary but it's nice to do:
```julia
fn = tempname()
try
    ...
finally
    rm(fn)
end
```
to ensure the file gets deleted even when an exception occurs

yuyichao:
LGTM.

nicoleepp:
Can this be tagged?

bjarthur:
um, note that a [PR of mine](https://github.com/JuliaIO/HDF5.jl/pull/469) which did near exactly the same thing as this PR was [eventually reverted](https://github.com/JuliaIO/HDF5.jl/pull/472) because we decided it was more julian to [require the user to called `collect`](https://github.com/JuliaIO/HDF5.jl/pull/469#issuecomment-377991051).

bjarthur:
this test fails with the master branch:

```
julia> using HDF5
[ Info: Recompiling stale cache file /Users/arthurb/.julia/compiled/v0.7/HDF5/L7Dga.ji for HDF5 [f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f]

julia> data = rand(UInt16,2,3);

julia> pdv_data = PermutedDimsArray(data,(2,1))
3×2 PermutedDimsArray(::Array{UInt16,2}, (2, 1)) with eltype UInt16:
 0x2d26  0x129a
 0x87b3  0x54d0
 0x0bd6  0xbcc9

julia> h5write("temp.h5", "/pdv_data", pdv_data)

julia> pdv_data_readback = h5read("temp.h5", "/pdv_data");

julia> pdv_data==pdv_data_readback 
false

julia> pdv_data_readback
3×2 Array{UInt16,2}:
 0x2d26  0x54d0
 0x129a  0x0bd6
 0x87b3  0xbcc9
```

yuyichao:
> did near exactly the same thing

They appear to be totally different since there's no copy here.

yuyichao:
> this test fails with the master branch:

HDF5 is **NOT** a generic format for saving julia object so round trip won't guarantee to work. Use `JLD` for that.

Of course in principle this should only be allowed for arrays with the same stride as `Array` and one can probably add a check for that.

bjarthur:
i edited the link to point [here](https://github.com/JuliaIO/HDF5.jl/pull/463/files).  near identical.

agreed that we need to throw an exception for AbstractArrays that fail silently.  @nicoleepp since i've written the test above, could you please add the code which checks the stride?

bjarthur:
i'd definitely say no tag until this issue is resolved.

nicoleepp:
https://github.com/JuliaIO/HDF5.jl/pull/523

PR #525, opened 2018-10-18
return a plain `Array` from mmap when alignment allows it

JeffBezanson:
This allows more tests to pass in JLD and JLDArchives.


musm:
do you think we could add a  test in this repo for this? thanks

musm:
What's the rationale for the `offset % Base.datatype_alignment(T) == 0` conditional?

tknopp:
@JeffBezanson: Isn't this making all code type unstable that uses the mmap feature from HDF5.jl?

PR #529, opened 2018-10-24
Remove manual homebrew commands

musm:


musm:
Seems to work and is cleaner

PR #538, opened 2018-11-22
Support reading AbstractArray

halleysfifthinc:
This is needed by JuliaIO/MAT.jl#86 to allow the use of `reinterpret`ed arrays as buffers. I'm not at all familiar with this code base, so I don't know if/what any tests for this should look like.


musm:
The tests for writing AbstractArray are https://github.com/JuliaIO/HDF5.jl/pull/520/files#diff-cb8a736be11ad8e4c992aa7427b04339R416

Is there a way to we can add a test for this just in case?

musm:
@yuyichao  is the check good enough for now?

PR #540, opened 2019-01-10
Initial support for reading BITFIELD type

### LINKED ISSUE #307: Reading HDF5 files containing boolean (H5T_BITFIELD) values
I'm trying to open files which contain `Table`s written via pytables and getting this error:

```
    ERROR: Class id 4 is not yet supported
     in hdf5_to_julia_eltype at /Users/tamasgal/.julia/v0.4/HDF5/src/plain.jl:1828
     in read at /Users/tamasgal/.julia/v0.4/HDF5/src/plain.jl:1380
     in read at /Users/tamasgal/.julia/v0.4/HDF5/src/plain.jl:1197
     in h5read at /Users/tamasgal/.julia/v0.4/HDF5/src/plain.jl:622
```

This only occurs for tables containing the type `tables.BoolCol`, corresponding to `H5T_BITFIELD`. Is it planned to support this in near future? Or is there nice workaround? Currently I need to reprocess the HDF5 files and convert the boolean column to Int8 to have `0` and `1`, which is of course currently just a dirty workaround?

Here is a dummy example of the Python code, if you want to recreate `foo.h5` and see the error:

```
# Julia code to read the file
using HDF5
data = h5read("foo.h5", "event_info")
```

```
# Python script to create foo.h5, which holds a table with a boolean column
import tables

class EventInfoDesc(tables.IsDescription):
    event_id = tables.UIntCol()
    mc_t = tables.Float64Col()
    overlays = tables.UInt8Col()
    triggered = tables.BoolCol()   # This causes trouble

h5file = tables.open_file('foo.h5', mode="w", title="Foo")
filters = tables.Filters(complevel=5)
event_info = h5file.create_table('/', 'event_info',
                                 EventInfoDesc, "Event Info",
                                 filters=filters)

# Writing some random rows
for i in range(10):
    event_info.row['event_id'] = i
    event_info.row['mc_t'] = i*2
    event_info.row['overlays'] = i*3
    event_info.row['triggered'] = True
    event_info.row.append()

event_info.flush()
h5file.close()
```

vilim:
In order to able to be able to read files written with the PyTables, it is necessary to read the BITFIELD type. 
This only enables reading of 1-bit width BITFIELD types, I don't fully understand their purpose to be able to implement the more general case. Resolves #307

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -237,6 +237,7 @@ const H5T_C_S1            = read_const(:H5T_C_S1_g)
 const H5T_STD_REF_OBJ     = read_const(:H5T_STD_REF_OBJ_g)
 const H5T_STD_REF_DSETREG = read_const(:H5T_STD_REF_DSETREG_g)
 # Native types
+const H5T_NATIVE_B8       = read_const(:H5T_NATIVE_B8_g)
 const H5T_NATIVE_INT8     = read_const(:H5T_NATIVE_INT8_g)
 const H5T_NATIVE_UINT8    = read_const(:H5T_NATIVE_UINT8_g)
 const H5T_NATIVE_INT16    = read_const(:H5T_NATIVE_INT16_g)
@@ -265,6 +266,7 @@ end
 const HDF5ReferenceObj_NULL = HDF5ReferenceObj(UInt64(0))
 
 ## Conversion between Julia types and HDF5 atomic types
+hdf5_type_id(::Type{Bool})       = H5T_NATIVE_B8
 hdf5_type_id(::Type{Int8})       = H5T_NATIVE_INT8
 hdf5_type_id(::Type{UInt8})      = H5T_NATIVE_UINT8
 hdf5_type_id(::Type{Int16})      = H5T_NATIVE_INT16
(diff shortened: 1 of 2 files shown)


musm:
Could we add some unit tests. Thanks

musm:
Thanks looks good. @ggggggggg  This adds Bool support.

@vilim  If we don't have any other feedback ping me again in 1-2 days, to merge.

vilim:
I guess it would be useful to add a cast somewhere from BitArray to Array[Bool], to support writing this type as well, but I am not sure where the best place for the cast would be.

vilim:
@musm thanks for having a look.
If there are no further suggestions for now, I guess this can be merged.

musm:
thanks @vilim  ping me in a week for a tag of HDF5. Let's let it sit on master for a bit of time before tagging.

tamasgal:
Haha, many thanks. I completely forgot about this since I am living with a workaround for a year or so ;)

Time to throw that out...

JeffBezanson:
This is a good change, but since it changes the mapping for the `Bool` type I believe it breaks reading files written with previous versions of this package. I'm trying to contend with this in JLD. Should some version checks be added?

Issue #89, opened 2014-04-30
Windows mmap test failures


tkelman:
Was trying to set up AppVeyor Windows CI for this package and uncovered this issue in the process:

```
  | | |_| | | | (_| |  |  Version 0.3.0-prerelease+2742 (2014-04-25 02:49 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit 4229d30* (5 days old master)
|__/                   |  x86_64-w64-mingw32

julia> Pkg.test("HDF5")
INFO: Testing HDF5
WARNING: read(s::IOStream,a::Array) is deprecated, use read!(s,a) instead.
 in depwarn at deprecated.jl:36
WARNING: Error reading A
ERROR: could not create file mapping
 in mmap_array at mmap.jl:134
while loading C:\Users\Tony\.julia\HDF5\test\jld.jl, in expression starting on line 136
while loading C:\Users\Tony\.julia\HDF5\test\runtests.jl, in expression starting on line 5
```

This is probably an issue in base Julia's mmap on Windows, but git blame tells me you wrote that section of code.

An AppVeyor build here https://ci.appveyor.com/project/tkelman/hdf5-jl/build/1.0.2/job/04fw9ibhdhlv8tup shows the same issue (except with a messed up backtrace pointing to `process_options at client.jl:282` for some reason?) with the most recent 32 bit Windows binary too.

simonbyrne:
Fixed by #545

Issue #357, opened 2017-01-17
Cannot create mmap file on windows


musm:
I keep getting the following error
`ERROR: LoadError: LoadError: could not create file mapping: Access is denied.`

musm:
Tested another pc, issue remains

simonster:
A while ago, @tkelman tried to figure out how to make `mmap` work on Windows, but he was unsuccessful (https://github.com/JuliaIO/HDF5.jl/issues/89).

Cody-G:
+1 for giving it another try...though I may know too little about Windows to help

musm:
Funny that you ask, I gave it another try today, but wasn't able to fix it (btw I believe this is an issue with base julia's mmap ccall, but I wasn't able to figure out what)

Cody-G:
Yes it's definitely an issue with mmap, not Julia.  I run into this often without using HDF5.jl

simonbyrne:
Closed by #545

PR #545, opened 2019-04-17
Fix mmap support for windows

tknopp:
This is a test to fix mmap support on windows.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1598,14 +1598,19 @@ function readmmap(obj::HDF5Dataset, ::Type{Array{T}}) where {T}
     if isempty(dims)
         return T[]
     end
-    local fd
-    prop = h5d_get_access_plist(checkvalid(obj).id)
-    try
-        ret = Ptr{Cint}[0]
-        h5f_get_vfd_handle(obj.file.id, prop, ret)
-        fd = unsafe_load(ret[1])
-    finally
-        HDF5.h5p_close(prop)
+    if !Sys.iswindows()
+        local fdint
+        prop = h5d_get_access_plist(checkvalid(obj).id)
+        try
+            ret = Ptr{Cint}[0]
+            h5f_get_vfd_handle(obj.file.id, prop, ret)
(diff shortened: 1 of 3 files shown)


tknopp:
Test is successful. This is a workaround for the mmap issue on windows. Since the file descriptor that we get from HDF5 is not correct, I instead simply open the file again. To get the permissions of the original file open, I needed to wrap the `get_intent` method. 

This closes #89 and #357.

tknopp:
appveyor on x86 has other issues. The other ones are green. 

Ping @musm for review.

tknopp:
@musm @timholy any objections to merging and making a release? I am requiring this in my package https://github.com/MagneticParticleImaging/MPIFiles.jl which is supposed to also work on windows.

musm:
I plan to take a closer look this tomorrow.

musm:
So is the issue `prop = h5d_get_access_plist(checkvalid(obj).id)` incorrect?

musm:
I don't really understand why 
`h5f_get_vfd_handle(obj.file.id, prop, ret)` 
gives the wrong result, do you?

musm:
In any case, since it wasn't working before I plan to merge this soon, since something that works even though not ideal,  is better than it not working at all.

@tknopp  It would be good to leave some breadcrumbs for future reference, if someone wants to delve more deeply into a proper fix.

tknopp:
> I don't really understand why
> `h5f_get_vfd_handle(obj.file.id, prop, ret)`
> gives the wrong result, do you?

No, if I would know that, I would have fixed it ;-)

My guess is that the file handle is not compatible with what Julia requires in the `fdio` function.

But I actually don't have access to Windows, so I am not really the best to work on the original issue.

I improved the documentation and this then should be good to go from my side. I would appreciate, if this could be integrated into the upcoming release.

musm:
Fantastic thank you very much @tknopp for working on this !

musm:
I plan a new release very soon with this.

PR #546, opened 2019-04-17
Try to Fix Mmap support for windows (alternative)

tknopp:
This is a test to check if appveyor likes this.


tknopp:
try was not successful

Issue #539, opened 2018-12-27
Can't build on Windows 10 


ahmetumutdurmus:
Sorry if this is a duplicate yet even though other people have had build problems and opened posts I haven't seen the same error I am getting so here it is. 

I am trying to build HDF5.jl on Windows 10 with Julia 1.0.0. From the package manager I write `build HDF5` command and get the following screen containing the error message:

```
(v1.0) pkg> build HDF5
  Building LibCURL ─→ `C:\Users\adurmus\.julia\packages\LibCURL\OoXMv\deps\build.log`
  Building WinRPM ──→ `C:\Users\adurmus\.julia\packages\WinRPM\Y9QdZ\deps\build.log`
  Building Homebrew → `C:\Users\adurmus\.julia\packages\Homebrew\l8kUw\deps\build.log`
  Building CMake ───→ `C:\Users\adurmus\.julia\packages\CMake\ZGawK\deps\build.log`
  Building Blosc ───→ `C:\Users\adurmus\.julia\packages\Blosc\lzFr0\deps\build.log`
  Building HDF5 ────→ `C:\Users\adurmus\.julia\packages\HDF5\ZOkE7\deps\build.log`
┌ Error: Error building `HDF5`:
│ ERROR: Can not delete output file : Access is denied. : C:\Users\adurmus\.julia\packages\WinRPM\Y9QdZ\deps\usr\x86_64-w64-mingw32\sys-root\mingw\bin\zlib1.dll
│ [ Info: Updating WinRPM package list
│ [ Info: Downloading https://cache.julialang.org/http://download.opensuse.org/repositories/windows:/mingw:/win32/openSUSE_Leap_42.2/repodata/repomd.xml
│ [ Info: Downloading https://cache.julialang.org/http://download.opensuse.org/repositories/windows:/mingw:/win64/openSUSE_Leap_42.2/repodata/repomd.xml
│ ┌ Info: Packages to update:
│ └   join(names(toup), ", ") = "zlib1"
│ ┌ Info: Packages to install:
│ └   join(names(todo), ", ") = "hdf5"
│ ┌ Info: Downloading:
│ └   name = "zlib1"
│ ┌ Info: Extracting:
│ └   name = "zlib1"
│ ERROR: LoadError: MethodError: no method matching pipeline_error(::Int64)
│ Closest candidates are:
│   pipeline_error(!Matched::Base.ProcessChain) at process.jl:701
[...]

Issue #526, opened 2018-10-19
Can't build on Ubuntu


piyose:
I tried to build HDF5 on Ubuntu 16.04.3 LTS, and I got this error message.

```
julia> Pkg.build("HDF5")
INFO: Building CMakeWrapper
INFO: Building Blosc
INFO: Building HDF5
Installing dependency hdf5-tools via `sudo apt-get install hdf5-tools`:
[sudo] password for hiroki: 
Reading package lists ... Done
Building dependency tree
Reading state information ... Done
hdf5-tools is already the newest version (1.8.16+docs-4ubuntu1.1).
================================[ ERROR: HDF5 ]=================================

LoadError: Provider BinDeps.PackageManager failed to satisfy dependency libhdf5
while loading /home/hiroki/.julia/v0.6/HDF5/deps/build.jl, in expression starting on line 41

================================================================================

================================[ BUILD ERRORS ]================================

WARNING: HDF5 had build errors.

 - packages with build errors remain installed in /home/hiroki/.julia/v0.6
 - build the package(s) and all dependencies with `Pkg.build("HDF5")`
 - build a single package by running its `deps/build.jl` script

================================================================================

```

The version of Julia is 0.6.4. I would like you to help.

musm:
should be fixed by the latest version when it goes through #555

Issue #521, opened 2018-10-09
Build error due to Xcode version


Paalon:
The following error
```julia
WARNING: Couldn't find bottle stanza in autoconf
Error: Your Xcode (1) is too outdated.
Please update to Xcode 9.2 (or delete it).
Xcode can be updated from the App Store.

Error: Your Command Line Tools are too outdated.
Update them from Software Update in the App Store.

==================================================================================[ ERROR: HDF5 ]===================================================================================

LoadError: failed process: Process(`/Users/paalon/.julia/v0.6/Homebrew/deps/usr/bin/brew install --ignore-dependencies staticfloat/juliatranslated/szip`, ProcessExited(1)) [1]
while loading /Users/paalon/.julia/v0.6/HDF5/deps/build.jl, in expression starting on line 41

====================================================================================================================================================================================

==================================================================================[ BUILD ERRORS ]==================================================================================

WARNING: HDF5 had build errors.

 - packages with build errors remain installed in /Users/paalon/.julia/v0.6
 - build the package(s) and all dependencies with `Pkg.build("HDF5")`
 - build a single package by running its `deps/build.jl` script

====================================================================================================================================================================================
```
is shown but my Xcode version is 10.0 and its newer than Xcode 9.2.
Environment: macOS 10.13.6

musm:
should be fixed by the latest version when it goes through #555

Issue #519, opened 2018-09-20
Installing HDF5 asks for sudo password


dawbarton:
I installed HDF5.jl (rather I installed MAT.jl and got HDF5 as a dependency) and was rather surprised that I was suddenly asked for my sudo password with no apparent reason. After further inspection I figured out that it was trying to install the hdf5 package via my distro package manager. This sort of automatic install is great, but it's probably worth printing a message saying what it is trying to do with sudo as I don't just type my passwords into terminals when randomly prompted...

vvjn:
Agreed. Also, on Ubuntu at least, if you type in the sudo password it doesn't actually install hdf5-tools (I think that's the package name) anyway. It just fails and I end up installing hdf5-tools on the terminal and rebuilding HDF5.jl. So most likely the best thing to do is to fail the build with a message saying what to do instead of trying to install hdf5-tools automatically.

musm:
should be fixed by the latest version when it goes through #555

PR #555, opened 2019-07-02
Update to use BinaryProvider

staticfloat:
Let's see if this passes tests!


Manifest.toml
```diff
@@ -0,0 +1,142 @@
+# This file is machine-generated - editing it directly is not advised
```
musm: Accidentally checked in?

musm:
closes 
https://github.com/JuliaIO/HDF5.jl/issues/539
https://github.com/JuliaIO/HDF5.jl/issues/526
https://github.com/JuliaIO/HDF5.jl/issues/519
https://github.com/JuliaIO/HDF5.jl/issues/521
https://github.com/JuliaIO/HDF5.jl/issues/516
https://github.com/JuliaIO/HDF5.jl/issues/504
https://github.com/JuliaIO/HDF5.jl/issues/483
https://github.com/JuliaIO/HDF5.jl/issues/481
https://github.com/JuliaIO/HDF5.jl/issues/471
https://github.com/JuliaIO/HDF5.jl/issues/447
https://github.com/JuliaIO/HDF5.jl/issues/410
https://github.com/JuliaIO/HDF5.jl/issues/334
https://github.com/JuliaIO/HDF5.jl/issues/97

award for most issues closed by a single PR ? 😆

Manifest.toml
staticfloat: Nope; it's nice to have a `Manifest.toml` so that sometime off in the future, people can still `pkg> instantiate` to get the exact versions that at least worked back in July of 2019.  It's kind of like a "known good point" in the space of all possible package versions.

musm:
Can we also now remove https://github.com/JuliaIO/HDF5.jl/blob/master/.travis.yml the hdf5 package downloads there ?

staticfloat:
There's a problem with Windows; investigating.  It's more complex than just missing the MSVCR90 runtime, unfortunately.

musm:
Do these release notes at all help 

from https://portal.hdfgroup.org/display/support/HDF5%201.10.4#releasenotes
```
New Features
============
 
    Configuration:
    -------------
    - Add toolchain and cross-compile support
 
      Added info on using a toolchain file to INSTALL_CMAKE.txt. A
      toolchain file is also used in cross-compiling, which requires
      CMAKE_CROSSCOMPILING_EMULATOR to be set. To help with cross-compiling
      the fortran configure process, the HDF5UseFortran.cmake file macros
      were improved. Fixed a Fortran configure file issue that incorrectly
      used #cmakedefine instead of #define.
```

staticfloat:
Windows issues may be resolved through a new BP release, which is currently making its way into `General`.

musm:
HDF5 version is 1.8 on windows, is this expected?

staticfloat:
Yes.  We're getting it via msys2, which only builds version 1.8.  I think that's not a problem for this package; HDF5.jl can deal with both versions.

musm:
Ok I'm working on trying to build HDF5 using a Cygwin Host environment and mingw, the build is just taking forever. Would we be able to utilize those? I believe it should be compatible with Julia.

staticfloat:
Likely, yes.  When you finish building it, upload them somewhere and I'll shove them through a BB audit pass to double-check.

staticfloat:
The windows tests work locally for me, but they're failing on Travis 32-bit, so I'm trying to figure out why that is.

staticfloat:
Okay so these are working now across all platforms.  Yay!  Here are the details:

* We use PyPI to source binaries for Linux and MacOS.  These pretty much "just work"; we rewrite the dependency chain to link against our BB-provided libz, which is the only real change.

* We use MSYS2 to source binaries for Win32/Win64, because the PyPI binaries use Visual Studio to build theirs, and we can't mix-and-match C runtime libraries.  This means that Windows is using 1.8.X whereas Linux and Mac are using 1.10.X, but that's not the end of the world.  Also, the win32 binaries depend on a different version of `libgcc` (they use `libgcc_s_dw2-1.dll` instead of the `sjl` variant) which I have had to bundle along inside the tarball.  I don't know what the ramifications of having multiple `libgcc_s.dll` files around are, but it seems to work, so I'm rolling with it for now.

In the future, it may be worthwhile to just manually compile tarballs with the compilers we know and trust, and bundle those up.  Long-term, we should continue to improve BB to add in emulation capabilities such that the HDF5 cross-compilation process can actually be completed within the BB environment.

simonbyrne:
> In the future, it may be worthwhile to just manually compile tarballs with the compilers we know and trust, and bundle those up. Long-term, we should continue to improve BB to add in emulation capabilities such that the HDF5 cross-compilation process can actually be completed within the BB environment.

If there are only small changes required to the HDF5 build process, we could also try to upstream those changes as well.

staticfloat:
> If there are only small changes required to the HDF5 build process, we could also try to upstream those changes as well.

The problem is that the HDF5 build process requires a lot of bootstrapped execution; they use their (compiled for target) binaries to generate source code that must then be compiled.  The "manual compilation" here is not because the build system is broken, it's because we can't do it inside of BB, because we can't run things like windows executables inside of BB (yet).

simonbyrne:
> The problem is that the HDF5 build process requires a lot of bootstrapped execution; they use their (compiled for target) binaries to generate source code that must then be compiled.

😱

aviks:
steven johnson has been attempting to remove the compile-and-run checks to autoconf checks. He thinks it's feasible, but its a long slog. see https://github.com/stevengj/HDF5builder and its forks for progress.

ViralBShah:
Pinging @stevengj so he knows about this.

musm:
I've been trying to build  windows binaries using a cygwin host and  mingw-w64 compiler 
```
$ ./configure --host=x86_64-w64-mingw32 --build=x86_64-w64-mingw32 --prefix=/cygdrive/c/Users/Mus/work/hdf5/out
$ make 
```
Unfortunately the build stalls at 
```
  CC       H5HFtest.lo
  CC       H5HFtiny.lo
  CC       H5HG.lo
  CC       H5HGcache.lo
  CC       H5HGdbg.lo
  CC       H5HGquery.lo
  CC       H5HL.lo
  CC       H5HLcache.lo
  CC       H5HLdbg.lo
  CC       H5HLint.lo
  CC       H5HLprfx.lo
  CC       H5HLdblk.lo
  CC       H5HP.lo
  CC       H5I.lo
H5I.c: In function ‘H5I__id_dump_cb’:
H5I.c:2125:47: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
     HDfprintf(stderr, "   obj   = 0x%08lx\n", (unsigned long)(item->obj_ptr));
                                               ^
  CC       H5Itest.lo
  CC       H5L.lo
  CC       H5Lexternal.lo
  CC       H5make_libsettings.o
  CCLD     H5make_libsettings.exe```

simonbyrne:
https://github.com/live-clones/hdf5/blob/f97ea92fe5450cefd4c4d9fb266e7f42a952f2f9/src/H5I.c#L2341

wow, they seem to just assume that `unsigned long` is the same size as the pointer?

staticfloat:
@musm have you tried doing a verbose build?  Try setting `V=1` or `VERBOSE=1` or something like that.

simonbyrne:
LGTM!

musm:
Unfortunately the build keeps freezing. I will try through MSYS2 and see if  that works.

Thanks so much for this fantastic PR.

PR #558, opened 2019-07-18
complex number support

### LINKED ISSUE #180: Read complex numbers from HDF5
Although there is no standardized way of storing complex numbers in an HDF5 file, a few de facto standards have emerged using compound datatypes.  Octave, for instance, uses a compound datatype with fields "real" and "imag", while pytables and h5py both use fields "r" and "i".

Since [the goal is to get to the point where we can read any HDF5 file](https://github.com/timholy/HDF5.jl/issues/92#issuecomment-42146612), I think it would be wonderful if HDF5.jl could load arrays from these files into arrays of complex numbers.

The following Python program can be used to make an HDF5 file with two matrices: one real and one complex:

```
#!/usr/bin/env python

import h5py

f = h5py.File("complex.h5", "w")
f["m1"] = [2, 1]
f["m2"] = [2, 1j]
f.close()
```

The goal is to get Julia to load both matrices using the following code:

```
using HDF5

println(h5read("complex.h5", "m1"))
println(h5read("complex.h5", "m2"))
```

However, currently it fails on the last line while reading the array `m2` containing complex numbers.

kleinhenz:
This closes #180. The approach mirrors what is done by `h5py`. Complex types are stored as a compound datatype with field names "r" and "i". This compound type is recognized during reading and loaded as the correct complex julia type. Real and complex field names are configurable via `HDF5.set_complex_field_names(real::String, imag::String)`. Complex support can be enabled/disabled via `HDF5.enable/disable_complex_support()`. All previous tests pass and I have added new tests for complex support specifically.

Although this is not part of the HDF5 standard I think it makes sense to include this here since it seems to be fairly basic and universally useful functionality that people coming from python would expect to just work. Also h5py has had this functionality for a while and doesn’t seem to have run into any major problems with their approach as far as I can tell.

@tknopp @musm what do you think of this?

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -2019,9 +2019,9 @@ function hdf5_to_julia_eltype(objtype)
             h5t_get_member_name(objtype.id, i-1)
           end
           membertypes = ntuple(N) do i
-            h5t_get_member_type(objtype.id, i-1) |> HDF5Datatype |> hdf5_to_julia_eltype
+            hdf5_to_julia_eltype(HDF5Datatype(h5t_get_member_type(objtype.id, i-1)))
           end
-          iscomplex = membernames == COMPLEX_FIELD_NAMES[] && membertypes[1] == membertypes[2] && membertypes[1] <: HDF5.HDF5Scalar
+          iscomplex = (membernames == COMPLEX_FIELD_NAMES[]) && (membertypes[1] == membertypes[2]) && (membertypes[1] <: HDF5.HDF5Scalar)
           T = iscomplex ? Complex{membertypes[1]} : HDF5Compound{N}
         else
           T = HDF5Compound{N}
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -323,6 +323,13 @@ cset(::Type{ASCIIChar}) = H5T_CSET_ASCII
 
 hdf5_type_id(::Type{C}) where {C<:CharType} = H5T_C_S1
 
+# global configuration for complex support
+complex_support = true
+complex_field_names = ("r", "i")
+enable_complex_support() = global complex_support = true
+disable_complex_support() = global complex_support = false
+set_complex_field_names(real::String, imag::String) = global complex_field_names = (real, imag)
+
 ## HDF5 uses a plain integer to refer to each file, group, or
 ## dataset. These are wrapped into special types in order to allow
 ## method dispatch.
@@ -1175,6 +1182,16 @@ datatype(dset::HDF5Attribute) = HDF5Datatype(h5a_get_type(checkvalid(dset).id), 
 datatype(x::HDF5Scalar) = HDF5Datatype(hdf5_type_id(typeof(x)), false)
 datatype(::Type{T}) where {T<:HDF5Scalar} = HDF5Datatype(hdf5_type_id(T), false)
 datatype(A::AbstractArray{T}) where {T<:HDF5Scalar} = HDF5Datatype(hdf5_type_id(T), false)
+function datatype(::Type{Complex{T}}) where {T<:HDF5Scalar}
+  complex_support || error("complex support disabled. call HDF5.enable_complex_support() to enable")
+  dtype = h5t_create(H5T_COMPOUND, 2*sizeof(T))
+  h5t_insert(dtype, complex_field_names[1], 0, hdf5_type_id(T))
+  h5t_insert(dtype, complex_field_names[2], sizeof(T), hdf5_type_id(T))
+  return HDF5Datatype(dtype)
+end
+datatype(x::Complex{T}) where {T<:HDF5Scalar} = datatype(typeof(x))
+datatype(A::AbstractArray{Complex{T}}) where {T<:HDF5Scalar} = datatype(eltype(A))
+
 function datatype(str::String)
     type_id = h5t_copy(hdf5_type_id(typeof(str)))
     h5t_set_size(type_id, max(sizeof(str), 1))
@@ -1203,7 +1220,8 @@ dataspace(dset::HDF5Dataset) = HDF5Dataspace(h5d_get_space(checkvalid(dset).id))
 dataspace(attr::HDF5Attribute) = HDF5Dataspace(h5a_get_space(checkvalid(attr).id))
 
 # Create a dataspace from in-memory types
-dataspace(x::T) where {T<:HDF5Scalar} = HDF5Dataspace(h5s_create(H5S_SCALAR))
+dataspace(x::Union{T, Complex{T}}) where {T<:HDF5Scalar} = HDF5Dataspace(h5s_create(H5S_SCALAR))
+
 function _dataspace(sz::Tuple{Vararg{Int}}, max_dims::Union{Dims, Tuple{}}=())
     dims = Vector{Hsize}(undef,length(sz))
     any_zero = false
@@ -1301,18 +1319,20 @@ function read(obj::DatasetOrAttribute)
     read(obj, T)
 end
 # Read scalars
-function read(obj::DatasetOrAttribute, ::Type{T}) where {T<:HDF5Scalar}
+function read(obj::DatasetOrAttribute, ::Type{T}) where {T<:Union{HDF5Scalar, Complex{<:HDF5Scalar}}}
     x = read(obj, Array{T})
     x[1]
 end
 # Read array of scalars
-function read(obj::DatasetOrAttribute, ::Type{Array{T}}) where {T<:HDF5Scalar}
+function read(obj::DatasetOrAttribute, ::Type{Array{T}}) where {T<:Union{HDF5Scalar, Complex{<:HDF5Scalar}}}
     if isnull(obj)
         return T[]
     end
     dims = size(obj)
     data = Array{T}(undef,dims)
-    readarray(obj, hdf5_type_id(T), data)
+    dtype = datatype(data)
+    readarray(obj, dtype.id, data)
+    close(dtype)
     data
 end
 # Empty arrays
@@ -1700,7 +1720,7 @@ for (privatesym, fsym, ptype) in
             obj, dtype
         end
         # Scalar types
-        ($fsym)(parent::$ptype, name::String, data::Union{T, AbstractArray{T}}, plists...) where {T<:ScalarOrString} =
+        ($fsym)(parent::$ptype, name::String, data::Union{T, AbstractArray{T}}, plists...) where {T<:Union{ScalarOrString, Complex{<:HDF5Scalar}}} =
             ($privatesym)(parent, name, data, plists...)
         # VLEN types
         ($fsym)(parent::$ptype, name::String, data::HDF5Vlen{T}, plists...) where {T<:Union{HDF5Scalar,CharType}} =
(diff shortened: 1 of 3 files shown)


tknopp:
Looks good, thank you! @musm could you do the final decision / review?

kleinhenz:
One downside of this approach is that the hdf5 datatype for complex numbers is created/destroyed on every read/write. I think this is necessary in order to enable dynamically enabling/disabling complex support and updating the field names. Another possibility is to create the types once at library initialization like in #341. I have also implemented this approach [here](https://github.com/kleinhenz/HDF5.jl/tree/complex_static) although I am seeing some intermittent segfaults which I am not sure are related.

I'm not sure which approach is better. From a compatibility perspective it is nice to be able to dynamically update complex number handling but creating the type repeatedly will have at least a little performance impact. What do people think is better?

src/HDF5.jl
```diff
@@ -323,6 +323,13 @@ cset(::Type{ASCIIChar}) = H5T_CSET_ASCII
 
 hdf5_type_id(::Type{C}) where {C<:CharType} = H5T_C_S1
 
+# global configuration for complex support
+complex_support = true
+complex_field_names = ("r", "i")
```
benchislett: Would `("re", "im")` be better here, since that's what Julia uses?

tknopp: I would stick with the default convention h5py uses.

tknopp:
ping @musm: Can this be merged?

src/HDF5.jl
```diff
@@ -323,6 +323,13 @@ cset(::Type{ASCIIChar}) = H5T_CSET_ASCII
...
+enable_complex_support() = global complex_support = true
+disable_complex_support() = global complex_support = false
+set_complex_field_names(real::String, imag::String) = global complex_field_names = (real, imag)
+
```
musm: do we really need these global setter functions?

What's the rationale for being able to enable/disable complex support? Same with set_complex_field_names?

src/HDF5.jl
musm: @tknopp  why do you think it is  better to stick to `h5py`  convention?

tknopp: The goal of HDF5 is to be language agnostic. It should not matter what programming language you use to access a file. Unfortunately, HDF5 itself defines no standard for complex numbers. But there is a very popular package (h5py) that proposes or uses a convention. Why should this convention be different for Julia, when we can stick to the convention h5py uses?

musm: Sounds reasonable.  MATLAB does not have a convention. So we can use the `h5py` precedent.

We should document  this however.

src/HDF5.jl
musm: In any case a better way to deal with such global constants is to use `Ref`'s with appropriate setters.

kleinhenz: I implemented it this way following what is done by h5py and for compatibility with existing practice/workflows. Since there is no actual hdf5 standard for complex numbers allowing the fieldnames to be set dynamically lets you interface more easily with existing conventions. For example I think that octave uses real/imag and I have also seen re/im. Also it's conceivable that somebody is storing a compound type with these fieldnames that doesn't represent a complex number. I'm not sure if this flexibility is necessary/worth it but I decided to follow the h5py precedent in allowing it.

benchislett: I have trouble seeing why we would ever need to disable complex support?
Surely it's not just there for performance?

src/HDF5.jl
kleinhenz: Yes the idea was to seamlessly compatible with h5py by default. If people are happy with this approach and want to merge I can add some documentation explaining the choice.

src/HDF5.jl
benchislett: Not sure why they would want to do that, but even if they did couldn't they just change the field names?

musm: @kleinhenz that makes sense 

Can we please update all the global constants to use refs (it's  precompile friendly, as opposed to globals)

```julia
const COMPLEX_SUPPORT = Ref(true) 
const COMPLEX_FIELD_NAMES = Ref(("r", "i"))

enable_complex_support() = COMPLEX_SUPPORT[] = true
disable_complex_support() = COMPLEX_SUPPORT[] = false
set_complex_field_names(real::AbstractString, imag::AbstractString) =  COMPLEX_FIELD_NAMES[] = ((real, imag))
```


And update everywhere else to use `COMPLEX_SUPPORT[]` instead of  `complex_support`

src/HDF5.jl
musm: I'm in favor of  sticking to the `h5py` precedent

Manifest.toml
```diff
@@ -11,9 +11,9 @@ version = "0.8.10"
 
 [[BinaryProvider]]
 deps = ["Libdl", "Logging", "SHA"]
-git-tree-sha1 = "8153fd64131cd00a79544bb23788877741f627bb"
+git-tree-sha1 = "c7361ce8a2129f20b0e05a89f7070820cfed6648"
```
musm: FYI we have deleted the manifest.toml file , so no need to make any changes.

src/HDF5.jl
tknopp: I can see the rational of being able to disable this feature. It allows to to read the complex numbers in its "raw" representation, which otherwise would not be possible. It also does not really hurt making this optional.

kleinhenz:
@musm I've updated the global constants to use refs and fixed the merge conflicts. I will add some documentation later today.

src/HDF5.jl
```diff
@@ -1993,7 +2013,19 @@ function hdf5_to_julia_eltype(objtype)
...
-        T = HDF5Compound{N}
+        # check if should be interpreted as complex
+        if COMPLEX_SUPPORT[] && N == 2
+          membernames = ntuple(N) do i
+            h5t_get_member_name(objtype.id, i-1)
+          end
+          membertypes = ntuple(N) do i
+            h5t_get_member_type(objtype.id, i-1) |> HDF5Datatype |> hdf5_to_julia_eltype
+          end
+          iscomplex = membernames == COMPLEX_FIELD_NAMES[] && membertypes[1] == membertypes[2] && membertypes[1] <: HDF5.HDF5Scalar
```
```suggestion
          iscomplex = (membernames == COMPLEX_FIELD_NAMES[]) && (membertypes[1] == membertypes[2]) && (membertypes[1] <: HDF5.HDF5Scalar)
```

src/HDF5.jl
```diff
@@ -1993,7 +2013,19 @@ function hdf5_to_julia_eltype(objtype)
...
     elseif class_id == H5T_COMPOUND
         N = Int(h5t_get_nmembers(objtype.id))
-        T = HDF5Compound{N}
+        # check if should be interpreted as complex
+        if COMPLEX_SUPPORT[] && N == 2
+          membernames = ntuple(N) do i
+            h5t_get_member_name(objtype.id, i-1)
+          end
+          membertypes = ntuple(N) do i
+            h5t_get_member_type(objtype.id, i-1) |> HDF5Datatype |> hdf5_to_julia_eltype
```
```suggestion
            hdf5_to_julia_eltype(HDF5Datatype(h5t_get_member_type(objtype.id, i-1)))
```
The rest of the code base doesn't use `|>` syntax so let's just stick to the convention ?

src/HDF5.jl
musm: Looking at `h5py` they don't offer a seperate way to enable disable support.

I kinda agree with @benchislett that even if they would want to do that they could just change the field names. 

I'm not 100% convinced which approach is best. I'm leaning towards  no option to enable disable support since that seems a little overboard. OTOH we can always remove this non-user facing functions in the future if deemed overkill.

src/HDF5.jl
kleinhenz: done

kleinhenz:
is there anything more that I should do here?

musm:
I guess the only thing is 

https://github.com/JuliaIO/HDF5.jl/pull/558#discussion_r316301388  

Do you know if h5py has an option for this as well? I looked but it seems like they only allow you to change the field names? I could be wrong though.

src/HDF5.jl
kleinhenz: I think you are correct that h5py doesn't provide a mechanism for completely disabling complex support. I don't have a strong preference on this either. I lean towards including it just to be the least disruptive to existing workflows but I could see it the other way also. 

One point in favor is that changing the field names will prevent you from reading data as complex when you don't want to but it can't prevent you from accidentally writing complex data in an unwanted format where previously it would have thrown an exception so I think including the enable/disable thing is somewhat safer for people with a preexisting way of handling complex data.

musm: Ok that's convincing enough 👍

musm:
I'll go ahead and merge this since this has been open for a while without comment

@tknopp  @kleinhenz  do you all approve as well?

kleinhenz:
sounds good to me

tknopp:
looks good!

musm:
@kleinhenz  thanks for sticking to this PR and seeing it through. It's great to have others improve this package

kleinhenz:
@musm thanks! glad we were able to get it merged!

lazarusA:
Why this does not work? what is the appropriate syntax ? 

```
f = h5open("file.h5", "w")
dset_f = d_create(f, "psi", datatype(Complex{Float64}), dataspace(5, 5, 10))
for i in 1:10
    dset_f[:, :, i] = zeros(Complex{Float64}, 5,5)
end
flush(f)
close(f)
```

kleinhenz:
@lazarusA does #591 fix your issue?

lazarusA:
I did `add HDF5#master` and my code still doesn't work. 

` using HDF5
f = h5open("file.h5", "w")
dset_f = d_create(f, "psi", datatype(Complex{Float64}), dataspace(5, 5, 10))
for i in 1:10
    dset_f[:, :, i] = zeros(Complex{Float64}, 5,5)
end
flush(f)
close(f)`

kleinhenz:
It's not merged yet so you have to add it with `add https://github.com/kleinhenz/HDF5.jl#complex_hyperslab`

lazarusA:
Nice! now it has been merged. And it works just beautiful.

bhawkins:
Thanks for adding this feature! I just thought I'd mention that this supports MATLAB MAT files, too:

```MATLAB
>> x = 1:10;
>> y = x + 10;
>> z = x + 1j * y;
>> save('test.mat', 'z', '-v7.3')
```

```julia
julia> using HDF5

julia> HDF5.set_complex_field_names("real", "imag");

julia> z = h5read("/tmp/test.mat", "z")
1×10 Array{Complex{Float64},2}:
 1.0+11.0im  2.0+12.0im  3.0+13.0im  …  8.0+18.0im  9.0+19.0im  10.0+20.0im
```

musm:
@kleinhenz I just noticed this but is it really allowed for `Complex{<:HDF5Scalar}`, specifically the case with  `HDF5Reference`, or should it be restricted to `Complex{<:HDF5BitsKind}` ?

kleinhenz:
yeah this probably doesn't make much sense for `HDF5Reference`. OTOH I think in general we want to move more in the direction of removing dispatch restrictions rather than tightening them since this could conceivably make sense for types not in `HDF5Scalar`, e.g. `BigFloat`.

musm:
Totally agreed, and actually recent commits on master have started to remove some unnecessary type restrictions for the high level interfaces

Issue #408, opened 2017-05-19
Performance issue reading HDF5Compound data


tamasgal:
Dear all,

I noticed that the readout of HDF5Compound datasets is very slow compared to python (with pandas, which uses pytables as backend), namely a factor of 10.

The sample dataset is a 200k x 15 array and here are the readout times using pandas and Julia:

```
In [1]: import pandas as pd
In [2]: %time data = pd.read_hdf("run1.h5", "angles/99/hits")
CPU times: user 224 ms, sys: 17.8 ms, total: 242 ms
Wall time: 244 ms
```

```
v0.5.0> using HDF5
v0.5.0> using BenchmarkTools
v0.5.0> f = h5open("run1.h5", "r")
HDF5 data file: run1.h5
v0.5.0> @benchmark data = read(f["angles/99/hits"])
BenchmarkTools.Trial:
  memory estimate:  291.37 MiB
  allocs estimate:  8815153
  --------------
  minimum time:     2.897 s (5.34% GC)
  median time:      2.918 s (6.31% GC)
  mean time:        2.918 s (6.31% GC)
  maximum time:     2.939 s (7.27% GC)
  --------------
  samples:          2
  evals/sample:     1
```

I was a bit surprised since I switched to Julia to boost the analysis performance (which is totally faster than in python/numpy) but in the end it runs in the same time due to the big impact of the slow readout.

Am I doing something wrong? Is there a way to speed up the read-in?

Should I prepare a sample file?

timholy:
I seem to recall that @mbauman once suggested the interface `read(io, "var"=>T)` where `T` is a Julia immutable type with the same memory layout as the compound type on disk. I suspect that might solve all your performance problems.

We didn't implement it at the time because `Pair` (which is what `=>` gives you) was introduced in what was then "bleeding edge" julia-0.4pre and not available in stable julia versions; once 0.4 came out, this never got revisited. It would be a great contribution to implement this.

tamasgal:
Thanks for the information! What exactly should `"var"` represent? I am not sure how this `Pair` readout should work, but maybe you can point me to the discussion so I can try to implement it somehow ;)

timholy:
`"var"` was meant to be the name of the dataset, e.g., `"angles/99/hits"`. But of course there are other options, like `read(dset=>T)`. To be honest, it's been so long since I worked closely with this package that I don't remember many details about our previous thoughts. Nor was I successful in searching for the original discussion. Sorry.

[...]

PR #559, opened 2019-07-19
Add h5read_compound to read compound structures efficiently

tamasgal:
The `h5read_compound` function with its three methods is already in use in all of my projects where I have to read `HDF5Compound` and is roughly two orders of magnitudes faster than using `read` or `h5read` and then parsing the data manually.

I carry over these functions all over the place and I would really like them to be included in `HDF5.jl`.

Credits go to @damiendr who provided the base code snippet back in 2017: https://github.com/JuliaIO/HDF5.jl/issues/408

For demonstration purposes I created a file `compound.h5` using Python+PyTables with  a compound structure using this Python script:

```python
import tables as tb


class Foo(tb.IsDescription):
    a = tb.Int32Col()
    b = tb.Float64Col()


h5file = tb.open_file("compound.h5", mode="w")
table = h5file.create_table("/", 'foo', Foo)

foo = table.row

for i in range(100_000):
    foo['a'] = i
    foo['b'] = i * 1.1
    foo.append()
```

Here are the results of a simple benchmark:

```julia
using HDF5
using BenchmarkTools

struct Foo
    a::Int32
    b::Float64
end

h5f = h5open("compound.h5", "r")

@btime read(h5f, "/foo");
@btime h5read_compound(h5f["/foo"], Foo);
@btime h5read("compound.h5", "/foo");
@btime h5read_compound("compound.h5", "/foo", Foo);
```

```
 97.445 ms (799562 allocations: 31.37 MiB)
  1.052 ms (25 allocations: 1.53 MiB)
  99.236 ms (799568 allocations: 31.37 MiB)
  3.061 ms (31 allocations: 1.53 MiB)
```

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -19,7 +19,7 @@ export
     d_create, d_create_external, d_open, d_read, d_write,
     dataspace, datatype, exists, file, filename,
     g_create, g_open, get_chunk, get_create_properties, get_datasets,
-    h5open, h5read, h5rewrite, h5writeattr, h5readattr, h5write,
+    h5open, h5read, h5read_compound, h5rewrite, h5writeattr, h5readattr, h5write,
     has, iscontiguous, ishdf5, ismmappable, name,
     o_copy, o_delete, o_open, p_create,
     readmmap, @read, @write, root, set_dims!, t_create, t_commit
@@ -731,6 +731,33 @@ function h5read(filename, name::String, indices::Tuple{Vararg{Union{AbstractRang
         close(fid)
     end
     dat
+end
+
+"""
+    function h5read_compound(dset::HDF5.HDF5Dataset, T::DataType)
+
+Reads a dataset directly using the given datatype and returns a `Vector{T}`.
+The implementation is based on @damiendr's post in HDF5.jl Issue#408.
+"""
+function h5read_compound(dset::HDF5Dataset, T::DataType)
+    filetype = datatype(dset)  # packed layout on disk
+    memtype_id = h5t_get_native_type(filetype.id)  # padded layout in memory
+    @assert sizeof(T) == h5t_get_size(memtype_id) "Type sizes mismatch!"
+    out = Vector{T}(undef, length(dset))
+    h5d_read(dset.id, memtype_id, H5S_ALL, H5S_ALL, H5P_DEFAULT, out)
+    h5t_close(memtype_id)
+    out
+end
+
+function h5read_compound(h5f::HDF5File, h5loc::AbstractString, T::DataType)
+    h5read_compound(h5f[h5loc], T)
+end
+
+function h5read_compound(filename::AbstractString, h5loc::AbstractString, T::DataType)
+    h5f = h5open(filename, "r")
+    out = h5read_compound(h5f[h5loc], T)
+    close(h5f)
+    out
 end
 
 function h5writeattr(filename, name::String, data::Dict)
--- a/test/plain.jl
+++ b/test/plain.jl
@@ -390,6 +390,30 @@ if !isempty(HDF5.libhdf5_hl)
     rm(fn)
 end
 
+# Test h5read_compound
+struct Vec3
+    a::Float64
+    b::Float64
+    c::Float64
+end
+struct Data
+    wgt::Float64
+    xyz::Vec3
+    uvw::Vec3
+    E::Float64
+end
+function test_data(data)
+    @test 2 == length(data)
+    @test -2.45590412 ≈ data[1].xyz.a
+end
+fn = joinpath(test_files, "compound.h5")
+test_data(HDF5.h5read_compound(fn, "/data", Data))
+h5open(fn, "r") do f
+    test_data(HDF5.h5read_compound(f, "/data", Data))
+    test_data(HDF5.h5read_compound(f["/data"], Data))
+end
+
+
 # Test that switching time tracking off results in identical files
 h5open("tt1.h5", "w") do f
     f["x", "track_times", false] = [1, 2, 3]
--- a/test/plain.jl
+++ b/test/plain.jl
@@ -2,6 +2,20 @@ using HDF5
 using CRC32c
 using Test
 
+# Structures for the compound.h5 file
+struct Vec3
+    a::Float64
+    b::Float64
+    c::Float64
+end
+struct Data
+    wgt::Float64
+    xyz::Vec3
+    uvw::Vec3
+    E::Float64
+end
+
+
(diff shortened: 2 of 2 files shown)


tamasgal:
I'd like to add a test but I still did not manage to use HDF5.jl to write a compound structure.

Btw. I don't really understand the problem on appveyor, any ideas?

tamasgal:
I am a bit puzzled by this error on Travis:

```
ERROR: LoadError: LoadError: error compiling top-level scope: type definition not allowed inside a local scope
```

why is it running in a local scope? The test suite works fine on my desktop...

tamasgal:
OK got it, it has to be outside of the test suite block.

tknopp:
Can this parse arrays of compounds?

In particular, the challenge is parsing the following file format:
https://netix.dl.sourceforge.net/project/ismrmrd/data/simple_gre.h5

I made a custom parser but of course it would be great is HDF5.jl would support this directly.

tamasgal:
> Can this parse arrays of compounds?
> 
> In particular, the challenge is parsing the following file format:
> https://netix.dl.sourceforge.net/project/ismrmrd/data/simple_gre.h5
> 
> I made a custom parser but of course it would be great is HDF5.jl would support this directly.

Not sure if I understand what you mean. In the example above it does read an array of compounds. I was not able to read the HDF5 file you provided. Can you provide a `struct` with the matching compound structure? If so, you should be able to simply read it using 

```julia
h5read_compound("simple_gre.h5", "/dataset/data", TheStruct);
```

tamasgal:
Hmm, I think I got it. I see that each entry in your file is an array of compounds... Not sure how to deal with that though.

tknopp:
I don't want to hold up this PR but just mention that in the end it would also be cool to read such files. h5py is able to do that. It basically constructs the the structs in a dynamic fashion. Your approach has the advantage that all is type stable.

kleinhenz:
@tknopp with #592 it looks like I am able to automatically read your file

tknopp:
That sounds very interesting. Note that I wrote a custom parser for the data:

https://github.com/MagneticResonanceImaging/MRIReco.jl/blob/master/src/IO/ISMRMRD.jl#L14

Of course I am relying on the current implementation. 
One issue that I was facing is, that the struct offsets were in part pretty strange in files that I found in the wild. I therefore needed to do something like this:

https://github.com/MagneticResonanceImaging/MRIReco.jl/blob/master/src/IO/ISMRMRD.jl#L70

Is you implementation handling that well?

Other question: In my library I have already defined the struct the is comparable to the compound stored, see 

https://github.com/MagneticResonanceImaging/MRIReco.jl/blob/master/src/Datatypes/RawAcqData.jl#L6

Ideally I am able to convert the named tuple into that.

tknopp:
initial testing looks very good! Nice job

tamasgal:
I am not against deriving the types dynamically, it's actually a nice feature to have, but there are also other use cases where you would like to map an existing `struct` (like in my/our case). We have a defined structure which we also use in other languages.

I made some preliminary checks and it also seems that the static approach is much faster than the dynamic one.

On the other hand, both approaches are orthogonal to each other, so I think they could be both merged.

What's the status of the PR? Any chance of accepting it or should I close it?

musm:
Any chance we can reach some convergence on this PR and https://github.com/JuliaIO/HDF5.jl/pull/592 ? 

Some sort of unanimous decision would be great, so we can make progress.

kleinhenz:
I agree that the two approaches are mostly orthogonal so it wouldn't hurt to merge both. I think my approach is more in line with how things work in `HDF5.jl` currently where all the type translations are handled automatically but that doesn't cover the use case where you have a custom type that you want to read into like this does.

@tamasgal I'm curious where my approach was significantly slower? The actual hdf5 io routines called in both cases should be identical. On this [benchmark](https://gist.github.com/kleinhenz/ddb8fa7de366e917c2520b963e0bb1f4) I get basically identical performance. In the case where there are strings or other vlen members then I have some overhead from copying the data over into the native julia type but this is not really avoidable as far as I can tell.

tamasgal:
@kleinhenz I have to check carefully but I had no time yet. I'll try out your example and compare that!

src/HDF5.jl
```diff
@@ -733,6 +733,33 @@ function h5read(filename, name::String, indices::Tuple{Vararg{Union{AbstractRang
...
+    function h5read_compound(dset::HDF5.HDF5Dataset, T::DataType)
+
+Reads a dataset directly using the given datatype and returns a `Vector{T}`.
+The implementation is based on @damiendr's post in HDF5.jl Issue#408.
+"""
+function h5read_compound(dset::HDF5Dataset, T::DataType)
+    filetype = datatype(dset)  # packed layout on disk
+    memtype_id = h5t_get_native_type(filetype.id)  # padded layout in memory
+    @assert sizeof(T) == h5t_get_size(memtype_id) "Type sizes mismatch!"
+    out = Vector{T}(undef, length(dset))
```
kleinhenz: why not `out = Array{T}(undef, size(dset))` to preserve all the size information?

kleinhenz:
I think it would be nice to integrate this a little more with the existing api infrastructure before it is exported. For example currently this doesn't allow reading hyperslabs.

It's interesting to note that the current `read` function 
```
# Read array of scalars
function read(obj::DatasetOrAttribute, ::Type{Array{T}}) where {T<:Union{HDF5Scalar, Complex{<:HDF5Scalar}}}
    if isnull(obj)
        return T[]
    end
    dims = size(obj)
    data = Array{T}(undef,dims)
    dtype = datatype(data)
    readarray(obj, dtype.id, data)
    close(dtype)
    data
end
```

would already almost work for this if it weren't for the `HDF5Scalar` type restriction. I think it is probably possible to do something very similar to this that integrates with the rest of the existing api by replacing the places where `HDF5Scalar` is used for dispatch with some trait based system that the user can opt into. I'd be willing to spend some time trying to work this out if people are interested/think this is a good approach.

src/HDF5.jl
tamasgal: Sure!

tamasgal:
Yes good catch and idea! I’d appreciate if you could take over, I am quite overcommitted these days (well, who is not in this community 🙈 ).

tknopp:
@kleinhenz: I second my request that this API is able of handeling the ISMRMRD data format. #592 will break my code in MRIReco.jl and I hope that the present PR will rescue me.

musm:
What's the status here? Should we proceed or do something else?

kleinhenz:
#652 is my idea of how to do this.

PR #560, opened 2019-07-24
extend lastindex to use `end` at any index

scottstanie:
Without the extra method, `a[:, end]` has no method available.

The current method can't work at all with multi-dimensional arrays, it just throws `Wrong number of indices supplied, supplied length 1 but expected 2.`


musm:
Thanks!

Issue #453, opened 2017-11-22
Remove or update S<:String 


musm:
The library has a lot of places that  have `S<:String` which doesn't make sense. Either we should get rid of this or use `S<:AbstractString` if appropriate

musm:
help wanted here.   We probably need to add tests what verify usage with substrings

PR #563, opened 2019-08-02
Work around julia bug

Keno:
This works around julia bug #31696.

As noted in https://github.com/JuliaLang/julia/issues/32752. As an aside I suspect these methods were supposed to be on `AbstractString` rather than string and the two got confused somewhere along the line.


musm:
> As an aside I suspect these methods were supposed to be on AbstractString rather than string and the two got confused somewhere along the line.

agreed, this probably happened when  `AbstractString`  transition occurred

musm:
does this need a test  or at least a comment in the source code regarding the work around? Should I go ahead and merge

musm:
https://github.com/JuliaIO/HDF5.jl/issues/453

PR #565, opened 2019-08-14
Delete Manifest.toml

musm:


musm:
appveyor is picking this up leading to build failures so just get rid of it

Issue #562, opened 2019-08-02
v0.12 depends on binaries, breaking function on all but few platforms


hrobotron:
Hello,

Being a long time user of Julia on i686/amd64 PC, I am trying to start using it on (embedded) ARM. Unfortunately, the crucial HDF5 (required by JLD, so practically essential) failed to build. The same reason reported by other user here:
https://stackoverflow.com/questions/57184292/gcc-aarch64-linux-gnu-is-not-supported-by-julia-package-hdf5/57325113#57325113

It seems that in v0.12 the HDF5 introduces some new binary dependency. But the binaries exist only for a few platforms. My workaround was to pin to v0.11.1.

However, I am convinced the "new" way is very bad. There is plenty of platforms and Julia shall not fail to support them without a good reason.

I can agree the new approach with binaries could be a good _alternative_ to using system-wide libhdf5, but it shall not fail so horribly. As a hotfix, I would recommend at least to broaden the error msg of unsupported binaries to something like: pin older version, and install systemwide hdf5 lib(s)...

musm:
cc @staticfloat

hrobotron:
With versions up to 0.11.1, it was sufficient at least on my apt-based systems to just install hdf5 libraries (OS distro binaries, no custom changes), and HDF5 & JLD just worked. I can believe that having extra source for HDF5 binaries may have some sense for certain platforms. But please _do not break the basic function on the platforms, which are not explicitly supported by the custom binaries_.

The HDF5 is a necessary prerequisite for JLD, and without JLD the whole Julia is practically unusable. I am afraid a glitch like this, with not a very understandable error msg, may divert lot of possible Julia newcomers/adopters, who do not have one of the very few architectures provided by HDF5 0.12 bins.

I think that at least for the very beginning, the same 0.12 code shall _at least_ provide a more descriptive error msg, recommending to pin to 0.11.1 as a workaround.

musm:
It certainly wasn't the intention to break  support of other platforms. In contrast, we were trying to circumvent the numerous bugs people have had on the major operating systems with building HDF5.  We should definitely have a manual fallback if the platform isn't supported and this is something that can be implemented in the build script. Thanks for letting us know @hrobotron .

musm:
for reference we should implement a similar fallback as in
https://github.com/stevengj/Blosc.jl/blob/master/deps/build.jl

musm:
```
using BinaryProvider

platform_key_abi()
```

[...]

PR #568, opened 2019-08-21
Add manual fallback compilation mode

musm:
close https://github.com/JuliaIO/HDF5.jl/issues/562


musm:
woot seems to work on macos and travis!

musm:
Great  manual compilation  works on travis for linux and macos!

musm:
Tested this several times locally and on travis, which shows that manual compilation works on linux and macos. I'm going to merge and tag this so that we can quickly have a fix for those users experiencing problems with the latest HDF5 version

PR #572, opened 2019-08-22
Upgrade to latest hdf5 version Windows binaries

musm:


musm:
strange locally downloading the 32 bit julia binaries and testing this, tests pass, but for some reason on appveyor 32-bit builds fail?

PR #578, opened 2019-08-27
Remove conditionals related to old HDF5 versions

musm:
Are builds are now for HDF5 1.10 and greater  and manual compilation downloads this version as well, so these conditionals are not needed.


musm:
Will merge tomorrow sans objections

musm:
Any objections?

Issue #581, opened 2019-09-02
getindex doesn't handle negative step values


rafaqz:
Reverse ranges are useful and usually accepted by getindex. The error is a little obtuse:

```jujlia
h5read("smap/SMAP_L4_SM_gph_20160102T223000_Vv4011_001.h5", "Geophysical_Data/land_fr
action_wilting", (300:-1:1, 1:300))
ERROR: InexactError: check_top_bit(Int64, -1)
Stacktrace:
 [1] throw_inexacterror(::Symbol, ::Any, ::Int64) at ./boot.jl:583
 [2] check_top_bit at ./boot.jl:597 [inlined]
 [3] toUInt64 at ./boot.jl:708 [inlined]
 [4] Type at ./boot.jl:738 [inlined]
 [5] convert at ./number.jl:7 [inlined]
 [6] setindex!(::Array{UInt64,1}, ::Int64, ::Int64) at ./array.jl:767
 [7] hyperslab(::HDF5Dataset, ::StepRange{Int64,Int64}, ::Vararg{Union{Int64, AbstractRange{Int64}},N} where N) at /home/raf/.julia/packages/HDF5/Y9Znv/src/HDF5.jl:1859
 [8] _getindex(::HDF5Dataset, ::Type, ::StepRange{Int64,Int64}, ::Vararg{Union{Int64, Abstra
ctRange{Int64}},N} where N) at /home/raf/.julia/packages/HDF5/Y9Znv/src/HDF5.jl:1772
 [9] getindex(::HDF5Dataset, ::StepRange{Int64,Int64}, ::UnitRange{Int64}) at /home/raf/.jul
ia/packages/HDF5/Y9Znv/src/HDF5.jl:1766
 [10] h5read(::String, ::String, ::Tuple{StepRange{Int64,Int64},UnitRange{Int64}}) at /home/
raf/.julia/packages/HDF5/Y9Znv/src/HDF5.jl:745
 [11] top-level scope at none:0
```

Looks like there is a `UInt` conversion of `step` in hyperslab.

musm:
What's strange is that https://confluence.hdfgroup.org/display/HDF5/H5S_SELECT_HYPERSLAB 
says that step is a pointer to an array of hsize_t which is an unsigned int. BUT changing the code to allow negative ints works... 😕

simonbyrne:
the `stride` argument is `hsize_t` which is an unsigned integer, so I assume this means that negative step sizes are not supported (at least not officially)

PR #583, opened 2019-09-20
Allow negative step sizes in slab

musm:
fix https://github.com/JuliaIO/HDF5.jl/issues/581


musm:
cc @rafaqz

musm:
yeah I don't know why this fails on windows, but I don't think this PR should work?

rafaqz:
Weird that this just works! but also great, I thought it would need an intermediate op. Do you think the performance would be similar to regular indexing? I'm wondering wether to do this by default in GeoData.jl to standardise dimension order.

musm:
Yeah it's odd it just works because it does a convert of the Int to an UInt, which HDF5 somehow understands means negative indexing. Maybe this is because of the two's complements?

Performance is identical, so no hit. 

Note there are some Cmake build failure that prevent windows CI from running.

PR #584, opened 2019-09-20
Drop empty dims

musm:
Would love some feedback, since this seems a bit hacky but could also be perfectly acceptable?

### DIFF
--- a/test/extend_test.jl
+++ b/test/extend_test.jl
@@ -20,8 +20,8 @@ d[1, 1:5] = [1.1231, 1.313, 5.123, 2.231, 4.1231]
 set_dims!(d, (1, 5))
 @test size(d) == (1, 5)
 
-@test d[:, end] ≈ [4.1231]
-@test d[end, :] == [1.1231 1.313 5.123 2.231 4.1231]
+@test d[:, end] ≈ 4.1231
+@test d[end, :] == [1.1231, 1.313, 5.123, 2.231, 4.1231]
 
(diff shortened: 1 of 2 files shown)


musm:
this doesn't really resolve the setindex problem though..

test/extend_test.jl
```diff
@@ -20,8 +20,8 @@ d[1, 1:5] = [1.1231, 1.313, 5.123, 2.231, 4.1231]
 set_dims!(d, (1, 5))
 @test size(d) == (1, 5)
 
-@test d[:, end] ≈ [4.1231]
-@test d[end, :] == [1.1231 1.313 5.123 2.231 4.1231]
+@test d[:, end] ≈ 4.1231
```
rafaqz: I think this should still be an array... only indexing with an Integer drops dims:

```julia
julia> rand(1, 5)[:, end]
1-element Array{Float64,1}:
 0.4666714180536573

julia> rand(1, 5)[1, end]
0.013875531617609127
```

I can't find anything in the array interface docs that is really clear on this, just copying Base indexing is probably best. Maybe we need a "taking the array interface seriously" discussion on discourse

musm: Please feel free to open a PR if you have some ideas

rafaqz:
Yeah we should probably make in it comprehensive to be worth the API change. I've been a bit too busy with other packages to really look at it yet.

test/extend_test.jl
rafaqz: Ok I'll try, it might be a month or two with my current workload

Issue #485, opened 2018-06-15
Wrong ZLIB being linked, Affects importing other packages


v-i-s-h:
HDF5.jl (or the internal binary dependency) is loading wrong version of ZLIB. This is affecting usage of other packages:
Julia Info:
```
Julia Version 0.6.3
Commit d55cadc350 (2018-05-28 20:20 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, broadwell)
```
Examples (this works):
```
$ julia -e "using PyPlot; using HDF5"
```
Example 2 (Reverse the import order, this creates error):
```
$ julia -e "using HDF5; using PyPlot"
sys:1: UserWarning: 
This call to matplotlib.use() has no effect because the backend has already
been chosen; matplotlib.use() must be called *before* pylab, matplotlib.pyplot,
or matplotlib.backends is imported for the first time.

The backend was *originally* set to 'Qt5Agg' by the following code:
  File "/home/vish/.julia/v0.6/Conda/deps/usr/lib/python2.7/site-packages/matplotlib/backends/__init__.py", line 16, in <module>
    line for line in traceback.format_stack()


WARNING: No working GUI backend found for matplotlib
ERROR: InitError: PyError (PyImport_ImportModule

The Python package matplotlib.pyplot could not be found by pyimport. Usually this means
that you did not install matplotlib.pyplot in the Python version being used by PyCall.

PyCall is currently configured to use the Julia-specific Python distribution
installed by the Conda.jl package.  To install the matplotlib.pyplot module, you can
use `pyimport_conda("matplotlib.pyplot", PKG)`, where PKG is the Anaconda
package the contains the module matplotlib.pyplot, or alternatively you can use the
Conda package directly (via `using Conda` followed by `Conda.add` etcetera).

Alternatively, if you want to use a different Python distribution on your
system, such as a system-wide Python (as opposed to the Julia-specific Python),
you can re-configure PyCall with that Python.   As explained in the PyCall
documentation, set ENV["PYTHON"] to the path/name of the python executable
[...]

PR #585, opened 2019-10-02
Bundle zlib alongside HDF5

staticfloat:
This fixes problems in testing HDF5 on systems that have a system-wide `libz` that is too old for other packages that ship Julia's `libz` but HDF5 gets loaded first, and so loads the system `libz`.  So while HDF5 itself was compiled against a very old `libz` and is thus very compatible, it can cause problems for other packages.

This will be solved in Julia 1.3+ where the dependency of `HDF5` upon `Zlib` will be automatically encoded in the JLL package dependency graph, but this is an important fix for 1.2-.


musm:
fixes https://github.com/JuliaIO/HDF5.jl/issues/485

Issue #349, opened 2016-12-24
Memory leak when reading many different objects


LewisHein:
There seems to be a memory leak when reading large arrays from a file. For example:

```{.jl}
#causes memory leak
using HDF5

dim1, dim2 = (55, 30000)

function blah()
    array1= h5read("big.h5", "big1.h5")
    array2 = h5read("big.h5", "big2.h5")

    result1 = array1 .* array2
    return findmax(result1)[1]
end

while true
    try
        rm("big.h5")
    end
    h5write("big.h5", "big1.h5", rand(dim1, dim2))
    h5write("big.h5", "big2.h5", rand(dim1, dim2))

    x = blah()
end
```

The leak isn't in writing the file: otherwise, this would leak memory too

```{jl}
#Doesn't leak memory
using HDF5 

dim1, dim2 = (55, 30000)

function blah()
    array1= h5read("big.h5", "big1.h5")
    array2 = h5read("big.h5", "big2.h5") 

    result1 = array1 .* array2
    return findmax(result1)[1]
end

while true
    try
        rm("big.h5")
    end
    h5write("big.h5", "big1.h5", rand(dim1, dim2))
    h5write("big.h5", "big2.h5", rand(dim1, dim2))

    #x = blah() ##NOTE: blah() isn't called, but everything else is the same.
end
```

Oddly enough, this doesn't leak memory. I can't figure out why, that would be since it's doing the same number of reads.

```{jl}
# Doesn't leak memory
using HDF5

dim1, dim2 = (55, 30000)
try
    rm("big.h5")
end
h5write("big.h5", "big1.h5", rand(dim1, dim2))
h5write("big.h5", "big2.h5", rand(dim1, dim2))


function blah()
    array1= h5read("big.h5", "big1.h5")
    array2 = h5read("big.h5", "big2.h5")

    result1 = array1 .* array2
    return findmax(result1)[1]
end

while true
    x = blah()
end
```

ggggggggg:
How did you determine there was a memory leak? Need more info to try to reproduce.

I copied and pasted your first code block into a REPL (v0.5) and watched the memory usage in Activity Monitor (Mac OS) for about 1 minute. It varied from about 320 MB to 370 MB, going both up and down in the time I observed.

LewisHein:
[...]

PR #587, opened 2019-10-07
try suggested changes in #349

scottstanie:
@yuyichao suggested this two years ago in https://github.com/JuliaIO/HDF5.jl/issues/349#issuecomment-313563553

I can't tell if there are negative consequences to switching these to `true`/ why these are false in the first place... I just know that I have been using a local dev-copy of HDF5.jl with these changes for a couple months, and the only difference I see is that the memory usage is much much less.

For my use, if I run my program that uses multiprocessing and does a bunch of small writes to separate files in each process, in the current master branch, my final memory goes from ~7Gb at the start of the run to 17GB.
When I make these changes, it stays below 10Gb... this still makes me think the problem in #349 is not fully solved by it, but it seems much better to me.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -605,7 +605,7 @@ function h5open(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Bool
     end
     close_apl = false
     if apl.id == H5P_DEFAULT
-        apl = p_create(H5P_FILE_ACCESS, false)
+        apl = p_create(H5P_FILE_ACCESS, true)
         close_apl = true
         # With garbage collection, the other modes don't make sense
         apl["fclose_degree"] = H5F_CLOSE_STRONG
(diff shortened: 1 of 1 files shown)


musm:
I'm not in a position to carefully understand the implications of this change, but as it seems like it fixes a memory leak I'm inclined to merge this unless @yuyichao thinks to take a different approach. @yuyichao  What do you think about these changes?

musm:
Certainly allocation  7GB trying to read an HDF5 file is unacceptable. 

@yuyichao  You made the following comment
> (and in general the toclose seems fragile).``
Would you care elaborate how we could go about fixing this fragility? (Seperate PR)

scottstanie:
Sorry, to clarify about the 7GB of my run- this is just from includes/setup of the scripts on 20 processes, there is not 7GB of allocation from opening one HDF5 file- I included that to show the difference in final memory from my change (with the change, the run only ends up 2-3GB high RAM usage than the start, without the change, more like ~10GB somehow distributed across the processes)

musm:
Anyone else   have comments on this PR change? 

I'm merging with the following caveat:

I'm not fully understanding of the implications of this change. However (1) it fixes the issue and as pointed out by @yuyichao  it probably should be freed on Julia's side (2) several users have commented that this fixes the problem.

PR #589, opened 2019-10-15
README: Update link to HDF5 supporitng languages.

lassepe:
Link was broken. Now pointing to: https://en.wikipedia.org/wiki/Hierarchical_Data_Format#Interfaces


musm:
thanks!  Is there an official documentation like for this?

lassepe:
I think [this](https://support.hdfgroup.org/HDF5/tools5desc.html) is the content the link used to point to. However, this seems to be more outdated and less informative than the Wikipedia section about it. Maybe we should put both?

musm:
No I think this is fine,thanks!

PR #591, opened 2019-10-30
fix hyperslab support for complex datasets

kleinhenz:
Fix type check in `getindex/setindex!` to allow using hyperslabs with complex datasets. Should fix the issue mentioned [here](https://github.com/JuliaIO/HDF5.jl/pull/558#issuecomment-548123645).


musm:
Add some tests?

PR #592, opened 2019-11-01
improve performance/ergonomics of reading compound datatypes

### LINKED ISSUE #408: Performance issue reading HDF5Compound data

kleinhenz:
When reading compound datatypes this replaces `HDF5Compound` with a named tuple with the correct data layout. The data is directly read into the output buffer instead of going through an `IOBuffer`. This results in very large performance gains. 

I wrote a small [benchmark](https://gist.github.com/kleinhenz/ddb8fa7de366e917c2520b963e0bb1f4) to test the performance of reading a one million element dataset of a small compound datatype with three float fields.

Before
```
BenchmarkTools.Trial:
  memory estimate:  329.44 MiB
  allocs estimate:  9000084
  --------------
  minimum time:     1.689 s (19.72% GC)
  median time:      1.878 s (24.72% GC)
  mean time:        1.860 s (25.70% GC)
  maximum time:     2.013 s (31.63% GC)
  --------------
  samples:          3
  evals/sample:     1⏎
```

After:
```
BenchmarkTools.Trial:
  memory estimate:  22.89 MiB
  allocs estimate:  64
  --------------
  minimum time:     4.305 ms (20.43% GC)
  median time:      4.584 ms (20.69% GC)
  mean time:        4.717 ms (23.21% GC)
  maximum time:     53.451 ms (91.03% GC)
  --------------
  samples:          1059
  evals/sample:     1⏎
```

Additionally I fixed support for variable length strings as fields of compound datatypes which seemed to be broken before.

In the case where the compound type contains strings/arrays/vlen types a copy is made to convert from the hdf5 compatible type (e.g. `Cstring`, `FixedArray`, etc.) to native julia types. This has some performance cost but I think is almost always what you want to do. 

Besides being much much faster I think that getting an array of `NamedTuples` is much easier to work with than the `HDF5Compound` struct there was before.

This should close #408. I prefer this solution to #559 since it doesn't change the interface and doesn't require manual specification of the struct.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1578,7 +1578,6 @@ function read(obj::DatasetOrAttribute, ::Type{HDF5Vlen{T}}) where {T<:Union{HDF5
     for i = 1:len
         h = structbuf[i]
         data[i] = p2a(convert(Ptr{T}, h.p), Int(h.len))
-
     end
     data
 end
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -486,6 +486,12 @@ struct FixedString{N}
 end
 length(::Type{FixedString{N}}) where N = N
 
+struct VariableArray{T}
+  len::Csize_t
+  p::Ptr{Cvoid}
+end
+eltype(::Type{VariableArray{T}}) where T = T
+
 # VLEN objects
 struct HDF5Vlen{T}
     data
@@ -1459,7 +1465,6 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
     h5object(obj_id, parent)
 end
 
-
 # convert Cstring/FixedString to String
 function normalize_types(x::NamedTuple{T}) where T
 
@@ -1471,6 +1476,8 @@ function normalize_types(x::NamedTuple{T}) where T
       join(Char.(xi.data))
     elseif Ti <: FixedArray
       reshape(collect(xi.data), size(Ti)...)
+    elseif Ti <: VariableArray
+      copy(unsafe_wrap(Array, convert(Ptr{eltype(xi)}, xi.p), xi.len, own=false))
     elseif Ti <: NamedTuple
       normalize_types(xi)
     else
@@ -1504,10 +1511,10 @@ function read(dset::HDF5Dataset, T::Union{Type{Array{U}}, Type{U}}) where U <: N
   HDF5.h5d_read(dset.id, memtype_id, HDF5.H5S_ALL, HDF5.H5S_ALL, HDF5.H5P_DEFAULT, buf)
 
   types = get_all_types(U)
-  normalize = any(t -> t <: Union{Cstring, FixedString, FixedArray}, types)
+  normalize = any(t -> t <: Union{Cstring, FixedString, FixedArray, VariableArray}, types)
   out = normalize ? normalize_types.(buf) : buf
 
-  reclaim = any(t -> t <: Cstring, types)
+  reclaim = any(t -> t <: Union{Cstring, VariableArray}, types)
   if reclaim
     dspace = dataspace(dset)
     # NOTE I have seen this call fail but I cannot reproduce
@@ -1570,6 +1577,7 @@ function read(obj::DatasetOrAttribute, ::Type{HDF5Vlen{T}}) where {T<:Union{HDF5
     for i = 1:len
         h = structbuf[i]
         data[i] = p2a(convert(Ptr{T}, h.p), Int(h.len))
+
     end
     data
 end
@@ -2011,15 +2019,18 @@ function hdf5_to_julia_eltype(objtype)
           dtype = HDF5Datatype(h5t_get_member_type(objtype.id, i-1))
           ci = h5t_get_class(dtype.id)
 
-          if ci != H5T_STRING
-            return hdf5_to_julia_eltype(dtype)
-          else
+          if ci == H5T_STRING
             if h5t_is_variable_str(dtype.id)
               return Cstring
             else
               n = h5t_get_size(dtype.id)
               return FixedString{Int(n)}
             end
+          elseif ci == H5T_VLEN
+            superid = h5t_get_super(dtype.id)
+            T = VariableArray{hdf5_to_julia_eltype(HDF5Datatype(superid))}
+          else
+            return hdf5_to_julia_eltype(dtype)
           end
         end
 
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1510,6 +1510,7 @@ function read(dset::HDF5Dataset, T::Union{Type{Array{U}}, Type{U}}) where U <: N
   reclaim = any(t -> t <: Cstring, types)
   if reclaim
     dspace = dataspace(dset)
+    # NOTE I have seen this call fail but I cannot reproduce
     h5d_vlen_reclaim(memtype_id, dspace.id, H5P_DEFAULT, buf)
   end
 
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -481,6 +481,11 @@ end
 size(::Type{FixedArray{T,D,L}}) where {T,D,L} = D
 eltype(::Type{FixedArray{T,D,L}}) where {T,D,L} = T
 
+struct FixedString{N}
+  data::NTuple{N, Cchar}
+end
+length(::Type{FixedString{N}}) where N = N
+
 # VLEN objects
 struct HDF5Vlen{T}
     data
@@ -1454,14 +1459,60 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
     h5object(obj_id, parent)
 end
 
+
+# convert Cstring/FixedString to String
+function normalize_types(x::NamedTuple{T}) where T
+
+  vals = map(values(x)) do xi
+    Ti = typeof(xi)
+    if Ti == Cstring
+      unsafe_string(xi)
+    elseif Ti <: FixedString
+      join(Char.(xi.data))
+    elseif Ti <: FixedArray
+      reshape(collect(xi.data), size(Ti)...)
+    elseif Ti <: NamedTuple
+      normalize_types(xi)
+    else
+      xi
+    end
+  end
+
+  NamedTuple{T}(vals)
+end
+
+# get a vector of all the leaf types in a (possibly nested) named tuple
+function get_all_types(::Type{NamedTuple{T, U}}) where T where U
+  types = []
+  for Ui in fieldtypes(U)
+    if Ui <: NamedTuple
+      append!(types, get_all_types(Ui))
+    else
+      push!(types, Ui)
+    end
+  end
+  types
+end
+
 function read(dset::HDF5Dataset, T::Union{Type{Array{U}}, Type{U}}) where U <: NamedTuple
   filetype = HDF5.datatype(dset)
   memtype_id = HDF5.h5t_get_native_type(filetype.id)  # padded layout in memory
   @assert sizeof(U) == HDF5.h5t_get_size(memtype_id) "Type sizes mismatch!"
 
-  out = Array{U}(undef, size(dset))
+  buf = Array{U}(undef, size(dset))
 
-  HDF5.h5d_read(dset.id, memtype_id, HDF5.H5S_ALL, HDF5.H5S_ALL, HDF5.H5P_DEFAULT, out)
+  HDF5.h5d_read(dset.id, memtype_id, HDF5.H5S_ALL, HDF5.H5S_ALL, HDF5.H5P_DEFAULT, buf)
+
+  types = get_all_types(U)
+  normalize = any(t -> t <: Union{Cstring, FixedString, FixedArray}, types)
+  out = normalize ? normalize_types.(buf) : buf
+
+  reclaim = any(t -> t <: Cstring, types)
+  if reclaim
+    dspace = dataspace(dset)
+    h5d_vlen_reclaim(memtype_id, dspace.id, H5P_DEFAULT, buf)
+  end
+
   HDF5.h5t_close(memtype_id)
 
   if T <: NamedTuple
@@ -1956,7 +2007,19 @@ function hdf5_to_julia_eltype(objtype)
         end
 
         membertypes = ntuple(N) do i
-          hdf5_to_julia_eltype(HDF5Datatype(h5t_get_member_type(objtype.id, i-1)))
+          dtype = HDF5Datatype(h5t_get_member_type(objtype.id, i-1))
+          ci = h5t_get_class(dtype.id)
+
+          if ci != H5T_STRING
+            return hdf5_to_julia_eltype(dtype)
+          else
+            if h5t_is_variable_str(dtype.id)
+              return Cstring
+            else
+              n = h5t_get_size(dtype.id)
+              return FixedString{Int(n)}
+            end
+          end
         end
 
         # check if should be interpreted as complex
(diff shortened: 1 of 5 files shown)


tknopp:
So, from my perspective this looks very good. What I request is:
- a major version bumb, when this lands so that I can pin my package to the previous version.

@kleinhenz: While I think you approach is better than #559, it would help a lot when we there would be possibility to convert the named tuple into the equivalent struct representation. One issue I am currently seeing is that you represent FixedSize Vectors as Vectors, while the "in-memory" representation is a `NTuple`.

musm:
We can certainly make a major version bump. 

> One issue I am currently seeing is that you represent FixedSize Vectors as Vectors, while the "in-memory" representation is a NTuple.

Can you double check here.  The code here is using a a fixed size NTuple not Vectors for FixedArray?

kleinhenz:
In the `normalize_types` function I translate from `HDF5.FixedArray` which is just a wrapper around an `NTuple` with the shape information saved in the type into an actual `Array`. This is definitely a design decision that I could see going either way. My thought was that in most cases you will want to work with the data as an array so it makes sense to just translate. If we depended on `StaticArrays` we could just use that, but that is probably overkill for this package. 

The reason I decided to do the type translation here is because it is basically necessary for string types. You don't want to return named tuples with `Cstring` members which point to memory allocated by hdf5. So if in order to get sensible julia types we have to do some type translations for strings I thought we might as well also do it for arrays. This could easily be changed if having compatibility with existing structs is more important.

In the case where there are no string or array members then I think using `reinterpret` in order to get to an equivalent struct representation should just work.

tknopp:
Can we make the type normalization optional? Maybe with the current behavior being the default.

kleinhenz:
So thinking about this a little more I think that in the case where you have a memory compatible struct already defined it makes a lot more sense to use it directly like in #559 than to use this and then reinterpret. The nice thing about this approach is that it gives you an ergonomic representation without you having to specify a struct manually. But if the struct is already defined then I don't think there's really any advantage to reading as named tuples first. The main function in #559 is <10 lines of code so I don't think using an approach like that is too onerous. I think that makes more sense than complicating the default interface by making the type normalization configurable.

tknopp:
Is it really complicating the interface? I mainly asking for a backdoor, where the raw things are returned. I am actually wondering how you plan to realize the write support. In that case you will need to denormalize the types.

kleinhenz:
In order to make it optional it seems like you need to either add a new global flag which is a bit unappealing or else add a new mechanism for passing keyword arguments through all of the machinery which seems like sort of a big change just for this. I can't just add it to the relevant method because usually this will be invoked through `h5read` or `read(dset)`.

What would be the advantage of using this and then reinterpreting the data into your custom struct over just reading directly into the custom struct?

IMHO write support is separate and maybe out of scope. I guess `JLD` already solves the problem of how to map julia types to compound hdf5 types. It's probably best to leave general serialization logic there.

tknopp:
> IMHO write support is separate and maybe out of scope. I guess JLD already solves the problem of how to map julia types to compound hdf5 types. It's probably best to leave general serialization logic there.

I disagree with that. You seem to think about compounds as some serialization format where the concrete way things are stored don't matter too much. But there are file formats (e.g. https://ismrmrd.github.io) that use a fixed compound type definition. I therefore consider this to be a contract where the user needs full control to fulfill the contract.

But anyway, I can live with the normalization and will work around it in my code.

kleinhenz:
That's a good point. Still I think that is an issue that should to be addressed separately. As noted in #341 making the set of natively supported types dynamically modifiable is currently pretty non-trivial because `HDF5Scalar` is used in dispatch everywhere. It would be nice to be able to define `datatype(::Type{mytype})` and have all of the high level machinery just work. It would also be nice to provide an extension point to allow user specification of the type to be read into to override the default namedtuple scheme proposed here. However I think both of these things would require fairly big changes to the internal library code.

The change proposed here is much more minimal since it doesn't do anything about user defined types but I think it's worth making since it greatly improves performance and named tuples are much more convenient than the `HDF5Compound` type.

src/HDF5.jl
```diff
@@ -1990,18 +1994,41 @@ function hdf5_to_julia_eltype(objtype)
...
+              n = h5t_get_size(dtype.id)
+              return FixedString{Int(n)}
+            end
+          elseif ci == H5T_VLEN
+            superid = h5t_get_super(dtype.id)
+            T = VariableArray{hdf5_to_julia_eltype(HDF5Datatype(superid))}
+          else
+            return hdf5_to_julia_eltype(dtype)
           end
-          iscomplex = (membernames == COMPLEX_FIELD_NAMES[]) && (membertypes[1] == membertypes[2]) && (membertypes[1] <: HDF5.HDF5Scalar)
```
musm: unrelated changes.

kleinhenz: this part was refactored because now `membernames`/`membertypes` are used outside of of the `COMPLEX_SUPPORT` path.

kleinhenz:
@musm what do you think about this approach?

tknopp:
@musm, @kleinhenz so I support merging this although it will break my package hardly. Maybe #559 can be then worked on in a separate step. As @kleinhenz outlined they are orthogonal. In this PR one has no knowledge about the compound. In #559 one has the compound at hand. So it may even dispatch nicely.

musm:
Yeah I think the approach here is a good one. I need to play around with it a bit more. 

What about wrapping the `NamedTuple` in an `HDF5Compound` container? This might help users more easily upgrade or transition.

kleinhenz:
How would you want to wrap it? To maintain the old interface or to just provide an outer type? 

My instinct would be to just return the named tuple directly since

1) the old interface was inconvenient enough to justify a breaking change
2) people already know how named tuples work so adapting code to use the named tuples will have less friction than trying to figure out the semantics of a new `HDF5Compound` container.

musm:
Fair enough. I think we will need an upcoming breaking release containing this PR, changes to the array interface to align them with Base julia,  and the update to the JLL artifacts.

kleinhenz:
@musm does anything else have to happen with this before it can be merged?

test/compound.jl
```diff
@@ -0,0 +1,79 @@
...
+  d::NTuple{9, ComplexF64}
+  e::HDF5.Hvl_t
+end
+
+function unsafe_convert(::Type{foo_hdf5}, x::foo)
+  foo_hdf5(x.a,
+           Base.unsafe_convert(Cstring, x.b),
+           ntuple(i -> x.c[i], length(x.c)),
+           ntuple(i -> x.d[i], length(x.d)),
+           HDF5.Hvl_t(convert(Csize_t, length(x.e)), convert(Ptr{Cvoid}, pointer(x.e)))
```
musm:
```suggestion
           HDF5.Hvl_t(length(x.e), pointer(x.e))
```

test/compound.jl
```diff
@@ -0,0 +1,79 @@
...
+  HDF5.h5t_set_size(vlenstr_dtype, HDF5.H5T_VARIABLE)
+  HDF5.h5t_set_cset(vlenstr_dtype, HDF5.H5T_CSET_UTF8)
+  HDF5.h5t_insert(dtype, "b", fieldoffset(foo_hdf5, 2), vlenstr_dtype)
+
+  fixedstr_dtype = HDF5.h5t_copy(HDF5.H5T_C_S1)
+  HDF5.h5t_set_size(fixedstr_dtype, 10 * sizeof(Cchar))
+  HDF5.h5t_set_cset(fixedstr_dtype, HDF5.H5T_CSET_UTF8)
+  HDF5.h5t_insert(dtype, "c", fieldoffset(foo_hdf5, 3), fixedstr_dtype)
+
+  hsz = HDF5.Hsize[3,3]
```
musm: Why [3,3] ?

kleinhenz: just to test loading from an `H5T_ARRAY` with rank > 1.

src/HDF5.jl
```diff
@@ -1575,6 +1578,7 @@ function read(obj::DatasetOrAttribute, ::Type{HDF5Vlen{T}}) where {T<:Union{HDF5
     for i = 1:len
         h = structbuf[i]
         data[i] = p2a(convert(Ptr{T}, h.p), Int(h.len))
+
```
```suggestion
```

test/compound.jl
musm: I see this number is linked to L60

kleinhenz: yeah that's right

src/HDF5.jl
```diff
@@ -1459,73 +1465,70 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
...
-        push!(data, read_row(iobuff, membertype, membersize))
-    end
-    # convert HDF5Compound type parameters to tuples
-    membername = (membername...,)
-    membertype = (membertype...,)
-    if T === HDF5Compound{N}
-        return HDF5Compound(data[1], membername, membertype)
+# get a vector of all the leaf types in a (possibly nested) named tuple
+function get_all_types(::Type{NamedTuple{T, U}}) where T where U
+  types = []
```
musm: should probably be 
```suggestion
  types = DataType[]
```

src/HDF5.jl
```diff
@@ -1459,73 +1465,70 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
...
-    while !eof(iobuff)
-        push!(data, read_row(iobuff, membertype, membersize))
-    end
-    # convert HDF5Compound type parameters to tuples
-    membername = (membername...,)
-    membertype = (membertype...,)
-    if T === HDF5Compound{N}
-        return HDF5Compound(data[1], membername, membertype)
+# get a vector of all the leaf types in a (possibly nested) named tuple
+function get_all_types(::Type{NamedTuple{T, U}}) where T where U
```
musm: I'm guessing there's a better way to do this that I can't currently think of, but seems fine for now.

musm: This seems faster / better

```julia
function get_all_types1(::Type{NamedTuple})
	types = DataType[]
	for Ui in U.types
	    if Ui <: NamedTuple
	      append!(types, get_all_types(Ui))
	    else
	      push!(types, Ui)
	    end
	  end
	  return types
end
```

src/HDF5.jl
```diff
@@ -482,9 +475,22 @@ end
 struct EmptyArray{T} end
 
 # Stub types to encode fixed-size arrays for H5T_ARRAY
-struct FixedArray{T,D} end
-size(::Type{FixedArray{T,D}}) where {T,D} = D
-eltype(::Type{FixedArray{T,D}}) where {T,D} = T
+struct FixedArray{T,D,L}
+  data::NTuple{L, T}
```
musm: how / where's this field being used ?

src/HDF5.jl
kleinhenz: maybe this would be better/more julian?

```
do_normalize(::Type{T}) where T = false
do_normalize(::Type{NamedTuple{T, U}}) where T where U = any(i -> do_normalize(fieldtype(U,i)), 1:fieldcount(U))
do_normalize(::Type{T}) where T <: Union{Cstring, FixedString, FixedArray, VariableArray} = true

do_reclaim(::Type{T}) where T = false
do_reclaim(::Type{NamedTuple{T, U}}) where T where U =  any(i -> do_reclaim(fieldtype(U,i)), 1:fieldcount(U))
do_reclaim(::Type{T}) where T <: Union{Cstring, VariableArray} = true
```

the only reason I collect all the types is to see if I need to call reclaim or do an extra normalization step to turn the data into a native julia type.

musm: yeah that might be simpler. It's certainly in our best favor to make the code run as fast as possible and be as simple as possible to read for maintenance.  You can also benchmark performance.

src/HDF5.jl
kleinhenz: This is used so that the julia type you construct for an `H5T_ARRAY` is memory compatible and can be directly written into.

src/HDF5.jl
```diff
@@ -1459,73 +1465,70 @@ function getindex(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, r::HDF5Refere
...
-            push!(row, reshape(val, size(dtype)))
-        elseif dtype<:HDF5BitsKind
-            push!(row, read(io, dtype))
-        else
-            # for other types, just store the raw bytes and let the user
-            # decide what to do
-            push!(row, read!(io, Vector{UInt8}(undef,dsize)))
-        end
+# convert Cstring/FixedString to String
+function normalize_types(x::NamedTuple{T}) where T
```
musm: where was this normalization obtained from?

src/HDF5.jl
```diff
@@ -2007,18 +2010,41 @@ function hdf5_to_julia_eltype(objtype)
         T = HDF5Vlen{hdf5_to_julia_eltype(HDF5Datatype(super_id))}
     elseif class_id == H5T_COMPOUND
         N = Int(h5t_get_nmembers(objtype.id))
```
```suggestion
        N = h5t_get_nmembers(objtype.id)
```
probably can do without the `Int` conversion just fine

src/HDF5.jl
```diff
@@ -2007,18 +2010,41 @@ function hdf5_to_julia_eltype(objtype)
...
+        membertypes = ntuple(N) do i
+          dtype = HDF5Datatype(h5t_get_member_type(objtype.id, i-1))
+          ci = h5t_get_class(dtype.id)
+
+          if ci == H5T_STRING
+            if h5t_is_variable_str(dtype.id)
+              return Cstring
+            else
+              n = h5t_get_size(dtype.id)
+              return FixedString{Int(n)}
```
musm: Again I'm guessing it's fine not do `Int` here as well. Unless you foresee negative consequences.

musm:
Thanks, I think we are almost there, just a few more tweaks and we can merge and tag a new major.

musm:
Any chance we could also get this to work with a variant of https://github.com/JuliaIO/HDF5.jl/pull/341 in the future?

src/HDF5.jl
kleinhenz: This one is actually necessary since NTuple requires the first type parameter to be `Int64`.

src/HDF5.jl
kleinhenz: this was taken from the current handling of strings/vlen types.

kleinhenz: except that for vlen arrays I use a non-owning unsafe wrap and then make a copy. I think this is the right thing to do since hdf5 wants to manage its own memory.

kleinhenz:
yeah I think that should work fine. I already checked that this integrates with the complex support so float16 should work the same way. One thing that doesn't work currently is hyperslab support since we have an annoying check for `T <: HDF5Scalar` in the `getindex` code. I think a lot of that code should probably be reworked to be more generic.

Issue #594, opened 2019-11-26
Use HDF5_jll.jl


crstnbr:
Are there yet any concrete plans to switch to using [HDF5_jll.jl](https://github.com/JuliaBinaryWrappers/HDF5_jll.jl) given that 1.3 will be released soon?

musm:
Did you create the repo? I'm  not sure who made that, but hopefully we can.

crstnbr:
> Did you create the repo? I'm not sure who made that, but hopefully we can.

No, it wasn't me. I assume it has been created by an Azure pipline based on [this `build_tarballs.jl`](https://github.com/JuliaPackaging/Yggdrasil/tree/master/H/HDF5) in Yggdrasil.

musm:
Thanks. Let me try to see if I can wire it up here.

PR #602, opened 2019-12-05
Use HDF5_jll

musm:
close https://github.com/JuliaIO/HDF5.jl/issues/594


musm:
Are we okay with not providing a manual compilation fallback? I'm not sure on the best way to integrate that option.

staticfloat:
It's fine with me, but someone who is a maintainer of this package should probably speak up about how important that is.  @timholy perhaps?

musm:
@timholy hasn't maintained this package for over several years. I'm basically acting like the janitor on this package trying to do my best with keeping it workable...

staticfloat:
Ah, my bad.  Then I'd say just do whatever makes it easier for you to maintain.  :)

musm:
We will need to figure out how to add a fallback manual compilation mode, but perhaps this is OK for now.

PR #603, opened 2019-12-06
getindex returns arrays with correct dims, or scalars

rafaqz:
This standardises the return value of `getindex` to match base julia `Array` behaviour. 

Dimensions indexed with Int are dropped using `reshape`. If all are dropped, a scalar is returned.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -5,7 +5,7 @@ using Base: unsafe_convert, StringVector
 import Base:
     close, convert, eltype, lastindex, flush, getindex, ==,
     isempty, isvalid, length, names, ndims, parent, read,
-    setindex!, show, size, sizeof, write, isopen, iterate
+    setindex!, show, size, sizeof, write, isopen, iterate, eachindex, axes
 
 import Libdl
 import Mmap
@@ -1752,8 +1752,15 @@ write(parent::Union{HDF5File, HDF5Group}, name::String, data::Union{T, AbstractA
 # For datasets, "write(dset, name, val)" means "a_write"
 write(parent::HDF5Dataset, name::String, data::Union{T, AbstractArray{T}}, plists...) where {T<:ScalarOrString} = a_write(parent, name, data, plists...)
 
-# Reading arrays using getindex: data = dset[:,:,10]
-function getindex(dset::HDF5Dataset, indices::Union{AbstractRange{Int},Int}...)
+
+# Indexing
+
+Base.eachindex(::IndexLinear, A::HDF5Dataset) = Base.OneTo(length(A))
+Base.axes(dset::HDF5Dataset) = map(Base.OneTo, size(dset))
+
+getindex(dset::HDF5Dataset, I::Union{AbstractRange,Integer,Colon}...) =
+    _getindex(dset, Base.to_indices(dset, I)...)
+function _getindex(dset::HDF5Dataset, I::Union{AbstractRange{Int},Int}...)
     local T
     dtype = datatype(dset)
     try
@@ -1761,14 +1768,14 @@ function getindex(dset::HDF5Dataset, indices::Union{AbstractRange{Int},Int}...)
     finally
         close(dtype)
     end
-    _getindex(dset,T, indices...)
+    _getindex(dset,T, I...)
 end
-function _getindex(dset::HDF5Dataset, T::Type, indices::Union{AbstractRange{Int},Int}...)
+function _getindex(dset::HDF5Dataset, T::Type, I::Union{AbstractRange{Int},Int}...)
     if !(T <: Union{HDF5Scalar, Complex{<:HDF5Scalar}})
         error("Dataset indexing (hyperslab) is available only for bits types")
     end
-    dsel_id = hyperslab(dset, indices...)
-    ret = Array{T}(undef,map(length, indices))
+    dsel_id = hyperslab(dset, I...)
+    ret = Array{T}(undef,map(length, I))
     memtype = datatype(ret)
     memspace = dataspace(ret)
     try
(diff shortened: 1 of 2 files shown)


rafaqz:
The coverage decrease is just inlining error

rafaqz:
I've updated this to also use `Base.to_indices` to align indexing with `AbstractArray` behaviour. `eachindex` is required for this to work on vectors.

musm:
I guess we need a breaking change with this PR and several others

musm:
Should we try to also fix missing trailing  indices?

musm:
I guess that is a bit orthogonal however

I think this is good to go

src/HDF5.jl
```diff
@@ -1830,8 +1835,12 @@ function setindex!(dset::HDF5Dataset, x::Number, indices::Union{AbstractRange{In
     setindex!(dset, X, indices...)
 end
 
-getindex(dset::HDF5Dataset, I::Union{AbstractRange{Int},Int,Colon}...) = getindex(dset, ntuple(i-> isa(I[i], Colon) ? (1:size(dset,i)) : I[i], length(I))...)
-setindex!(dset::HDF5Dataset, x, I::Union{AbstractRange{Int},Int,Colon}...) = setindex!(dset, x, ntuple(i-> isa(I[i], Colon) ? (1:size(dset,i)) : I[i], length(I))...)
+getindex(dset::HDF5Dataset, I::Union{AbstractRange{Int},Int,Colon}...) =
```
```suggestion
function getindex(dset::HDF5Dataset, inds...)
   _getindex(dset, Base.to_indices(dset, inds)...)
end
```

musm: And rename `getindex` to `_getindex` on L1751

src/HDF5.jl
```diff
@@ -1830,8 +1835,12 @@ function setindex!(dset::HDF5Dataset, x::Number, indices::Union{AbstractRange{In
     setindex!(dset, X, indices...)
 end
 
-getindex(dset::HDF5Dataset, I::Union{AbstractRange{Int},Int,Colon}...) = getindex(dset, ntuple(i-> isa(I[i], Colon) ? (1:size(dset,i)) : I[i], length(I))...)
-setindex!(dset::HDF5Dataset, x, I::Union{AbstractRange{Int},Int,Colon}...) = setindex!(dset, x, ntuple(i-> isa(I[i], Colon) ? (1:size(dset,i)) : I[i], length(I))...)
+getindex(dset::HDF5Dataset, I::Union{AbstractRange{Int},Int,Colon}...) = 
+    getindex(dset, Base.to_indices(dset, I)...)
+setindex!(dset::HDF5Dataset, x, I::Union{AbstractRange{Int},Int,Colon}...) =
```
```suggestion
function setindex!(dset::HDF5Dataset, x, inds...)
    setindex!(dset, x, Base.to_indices(dset, inds)...)
end```

Comment (musm) [reply src/HDF5.jl]: This way indexing should work with, e.g. `Int32` 

`d[Int32(1),Int32(1)]`

**Comment by musm** [comment]
This is a net improvement, made a couple of comments. 

We should also try to fix the trailing indices. 

I think after this PR and tackling the dropping trailining indices. The array interface should match Base correct?

**Comment by rafaqz** [comment]
Trailing indices should already be omitted with `Base.to_indices`? We should probably test it tho. I'll add a few more tests.

**Comment by rafaqz** [comment]
So you are right trailing indices are still broken. Unfortunately the base methods that deal with it are very `AbstractArray` specific. So it would involve writing a custom method to drop trailing 1s. 

Some changes made are I cleaned up the indexing code and added `Array()` constructor so you don't have to use `HDF5.read` to get the whole array.

File: src/HDF5.jl [hunk 3]
```diff
@@ -1752,23 +1752,35 @@ write(parent::Union{HDF5File, HDF5Group}, name::String, data::Union{T, AbstractA
...
-function getindex(dset::HDF5Dataset, indices::Union{AbstractRange{Int},Int}...)
+
+# Indexing
+
+Array(dset::HDF5Dataset) = read(dset)
+
+Base.eachindex(::IndexLinear, A::HDF5Dataset) = Base.OneTo(length(A))
+Base.axes(dset::HDF5Dataset) = map(Base.OneTo, size(dset))
+
+getindex(dset::HDF5Dataset, I::Union{AbstractRange{Int},Int,Colon}...) = begin
```
```
function getindex(dset::HDF5Dataset, I...)
    _getindex(dset, Base.to_indices(dset, I)...)
end
```

we want to be able to index with UInt or Int32 , the method restriction is necessary.

src/HDF5.jl
```diff
@@ -1752,23 +1752,35 @@ write(parent::Union{HDF5File, HDF5Group}, name::String, data::Union{T, AbstractA
 # For datasets, "write(dset, name, val)" means "a_write"
 write(parent::HDF5Dataset, name::String, data::Union{T, AbstractArray{T}}, plists...) where {T<:ScalarOrString} = a_write(parent, name, data, plists...)
 
-# Reading arrays using getindex: data = dset[:,:,10]
-function getindex(dset::HDF5Dataset, indices::Union{AbstractRange{Int},Int}...)
+
+# Indexing
+
+Array(dset::HDF5Dataset) = read(dset)
```
musm: Unrelated? Why add this?

src/HDF5.jl
musm: I'm also not confident about the bounds checking approach you took, in fact it doesn't actually do any bounds checking. 
We'd need something like
```
@boundscheck checkbounds(A, I...)
```
if you look `hyperslab` is doing to bounds checking right now.  Maybe just leave this to another PR?

src/HDF5.jl
```diff
@@ -1778,29 +1790,38 @@ function _getindex(dset::HDF5Dataset, T::Type, indices::Union{AbstractRange{Int}
...
 
+_dropint(x::Int, args...) = _dropint(args...)
+_dropint(x::AbstractRange, args...) = (x, _dropint(args...)...)
+_dropint() = ()
+
+
 # Write to a subset of a dataset using array slices: dataset[:,:,10] = array
-function setindex!(dset::HDF5Dataset, X::Array, indices::Union{AbstractRange{Int},Int}...)
+
+setindex!(dset::HDF5Dataset, x, I::Union{AbstractRange{Int},Int,Colon}...) =
```
musm: Same thing here, no need to restrict the `I` type
```
function setindex!(dset::HDF5Dataset, x, I...)
    _setindex!(dset, x, Base.to_indices(dset, I)...)
end
```

src/HDF5.jl
rafaqz: I'll put it it another PR for clarity.

rafaqz:
I simplified the PR removing the bounds check and `Array()`. 

Dispatch is on `{AbstractRange,Integer,Colon}` so the correct `MethodError` is given on `getindex` instead of showing all the `_getindex` options. 

Also added a few more tests for completeness.

musm:
great thanks

src/HDF5.jl
musm: did you forget to remove this for this PR?

src/HDF5.jl
```diff
@@ -1752,23 +1752,32 @@ write(parent::Union{HDF5File, HDF5Group}, name::String, data::Union{T, AbstractA
...
 
-# Reading arrays using getindex: data = dset[:,:,10]
-function getindex(dset::HDF5Dataset, indices::Union{AbstractRange{Int},Int}...)
+
+# Indexing
+
+Array(dset::HDF5Dataset) = read(dset)
+
+Base.eachindex(::IndexLinear, A::HDF5Dataset) = Base.OneTo(length(A))
+Base.axes(dset::HDF5Dataset) = map(Base.OneTo, size(dset))
```
musm: I don't this this method is needed for this PR

musm:
a few minor comments

src/HDF5.jl
rafaqz: must have undone it and rebased it back in.

src/HDF5.jl
rafaqz: Its used in `to_indices` to convert `Colon` to range:

https://github.com/JuliaLang/julia/blob/master/base/indices.jl#L321

musm: strange mine was working without this definition

musm:
great thanks for this! Major improvement

PR #605, opened 2019-12-14
add Array constructor for HDF5Dataset

rafaqz:
This adds a constructor for `Array`, which allows loading the whole dataset  (ie without specifying indices) as you can do with any real `AbstractArray`. This removes the need for boilerplate special casing of HDF5 with `HDF5.read` when working with multiple data sources.


musm:
thank you

Issue #606, opened 2019-12-17
MPI support and HDF5_jll


jipolanco:
Hi,

I have noticed that the recent switch to using `HDF5_jll` (c26820bfdc) to provide the HDF5 binaries has removed the possibility to use MPI-enabled HDF5 libraries. Previously, this was possible via the `FORCE_COMPILE_HDF5` environment variable, which has no effect in the current master.

Are there any plans to bring back an approach to use parallel HDF5 libraries?

I feel like, in this case, it would be acceptable to use system provided HDF5 libraries, since often one wants to run on HPC clusters that already have some version of HDF5 compiled with parallel support. Couldn't this be done via an optional environment variable, similarly to [what MPI.jl does](https://juliaparallel.github.io/MPI.jl/latest/installation/#Building-1) to detect the system MPI implementation to use?

If this is an acceptable strategy, I could make an attempt at supporting user-provided HDF5 libraries.

musm:
@jipolanco  we definitely want to provide a "fallback" mechanism which would allow users to compile HDF5 themselves and bypass `HDF5_jll`.  Feel free to open a PR as we would take it very seriously and is something I want to have in before we make a new release.  Take a look at how FFTW.jl handles this (albeit a different mechanism)

jipolanco:
Ok, thanks for your answer. I'll start working on it.

jipolanco:
Closed by e8035363e1848d1d91fcdea192c1a0c36b1238ed

PR #607, opened 2019-12-19
Add option to bypass HDF5_jll

jipolanco:
As discussed in #606, this is an initial proposal to provide a fallback mechanism for using a custom HDF5 build, instead of `HDF5_jll`.

This mechanism is enabled by setting the `JULIA_HDF5_LIBRARY_PATH` environment variable to the path of the HDF5 libraries at build time. By default, if this variable is not set, we use `HDF5_jll`.

The implementation is modelled after similar mechanisms used by FFTW.jl and MPI.jl.


musm:
LGTM. Presumably this worked locally?

jipolanco:
Yes, forgot to mention that. I tested it with HDF5 linked to MPICH, installed from the Ubuntu repos. MPI functionality works as well.

src/HDF5.jl
```diff
@@ -2556,6 +2563,8 @@ function __init__()
 
     register_blosc()
 
+    check_deps()
```
musm: Maybe move this up to the start of the function? L2559

deps/build.jl
```diff
@@ -0,0 +1,45 @@
...
+                # Do not edit
+
+                function check_deps()
+                    if libhdf5_size != filesize(Libdl.dlpath(libhdf5))
+                        error("HDF5 library has changed, re-run Pkg.build(\\\"HDF5\\\")")
+                    end
+                end
+                """
+               )
+        println(io, :( const libhdf5 = $libhdf5 ))
```
musm:
```suggestion
        println(io, :(const libhdf5 = $libhdf5))
```

musm:
Do you mind adding a very short statement about the env variable in the README as well?

deps/build.jl
jipolanco: Updated this and the position of the `check_deps` call in 69e6c9d.

src/HDF5.jl
```diff
@@ -26,6 +24,15 @@ export
     o_copy, o_delete, o_open, p_create,
     readmmap, @read, @write, root, set_dims!, t_create, t_commit
 
+# This is taken from FFTW.jl
```
musm: unecessary comment

```suggestion
```

musm:
We'll need to update 
https://github.com/JuliaIO/HDF5.jl/blob/master/README.md#installation

with the new instructions here

jipolanco:
Yes, I'm working on the instructions

jipolanco:
I just pushed a proposal for the instructions, but feel free to make any changes

musm:
Thanks a ton!

PR #614, opened 2020-04-01
unpad fixed length strings in compound datatypes

### LINKED ISSUE #379: Not yet supported

kleinhenz:
This is a minor followup to #592. Now unpadding is applied to fixed length strings in compound datatypes to match the behavior of plain fixed length strings.  Should close #379. I found this oversight while reading the file linked in that issue.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1780,7 +1780,7 @@ _dropint() = ()
 
 # Write to a subset of a dataset using array slices: dataset[:,:,10] = array
 
-setindex!(dset::HDF5Dataset, x, I::Union{AbstractRange,Integer,Colon}...) = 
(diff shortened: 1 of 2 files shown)


kleinhenz:
Just to be clear, compound datatypes with fixed length strings already work, this just applies the unpadding procedure in order to be consistent with the non compound case.

musm:
lgtm

musm:
This looks fairly uncontroversial, thank you!

musm:
Should be published soon, thanks again!

PR #617, opened 2020-04-21
Virtual Dataset

AStupidBear:


src/HDF5.jl
```diff
@@ -778,6 +791,7 @@ checkvalid(obj) = isvalid(obj) ? obj : error("File or object has been closed")
 # Close functions that should try calling close regardless
 function close(obj::HDF5File)
     if obj.id != -1
+        flush(obj)
```
musm, line 794: is this something we forgot ?

src/HDF5.jl
```diff
@@ -656,15 +657,16 @@ Open or create an HDF5 file where `mode` is one of:
...
+function h5open(filename::AbstractString, mode::AbstractString="r", pv...; swmr=false, fclose_degree = H5F_CLOSE_STRONG, auto_fclose = true)
     checkprops(pv...)
     # pv is interpreted as pairs of arguments
     # the first of a pair is a key of hdf5_prop_get_set
     # the second of a pair is a property value
     fapl = p_create(H5P_FILE_ACCESS, true, pv...) # file access property list
     # With garbage collection, the other modes don't make sense
     # (Set this first, so that the user-passed properties can overwrite this.)
-    fapl["fclose_degree"] = H5F_CLOSE_STRONG
+    fclose_degree != H5F_CLOSE_STRONG && @eval GC.gc(true)
```
musm, line 668: why this?

musm:
There's also the option to enable:
https://github.com/JuliaIO/HDF5.jl/blob/master/src/HDF5.jl#L2586
but we still want to capture that error so it's available to the user somehow

src/HDF5.jl
musm: Sorry, I meant why the GC.gc on this line

src/HDF5.jl
AStupidBear: This is only necessary when `fclose_degree` is `H5F_CLOSE_WEAK`. I think libhdf5 will handle this internally when `fclose_degree` is `H5F_CLOSE_STRONG`.

AStupidBear:
@musm `fclose_degree` is only a workaround for now. `h5py` also use `H5F_CLOSE_STRONG` as default but dosn't have issues like `HDF5.jl`. I don't know why.

AStupidBear:
@musm Any idea?

src/HDF5.jl
musm: https://github.com/JuliaIO/HDF5.jl/issues/620#issuecomment-619417143 
So maybe we are missing something similar here causing the prevention of reading the virutal dataset?

I just don't understand the necessity of this `GC.gc(true)` which we should try to avoid.

test/runtests.jl
```diff
@@ -12,6 +12,9 @@ include("gc.jl")
 include("external.jl")
 include("swmr.jl")
 include("mmap.jl")
+if Sys.islinux()
```
musm, line 15: linux only?

AStupidBear: It failed on windows, so I temporaly disabled that in test.

AStupidBear:
#621 didn't solve this problem

src/HDF5.jl
AStupidBear: #621 didn't solve this problem

musm:
Should we rebase and try to get this in again?

mkitti:
Thank you for this implementation and providing it in https://github.com/AStupidBear/HDF5Utils.jl/blob/master/src/virtual.jl . I think we will need to redesign this to make it fit into the current HDF5.jl. However, the HDF5Utils.jl design is a good starting point to consider.

PR #621, opened 2020-04-25
Check that HDF5Properties is valid before closing

### LINKED ISSUE #620: h5_close and property lists in h5open
The following example causes an "error closing property list":

```julia
using HDF5

h5open("proplist.h5", "w", "alignment", (0, 64)) do ff
end

HDF5.h5_close()
```

In this example, a file access property list is created inside `h5open` (in this case, to set [`alignment properties`](https://portal.hdfgroup.org/display/HDF5/H5P_SET_ALIGNMENT)). The property list is created with a finaliser, which will close the HDF5 object when the GC destroys its Julia wrapper. Later, when `h5_close` is called, the HDF5 object is closed a second time (or a first time, if the GC has not done its job yet), which results in an error.

This causes problems when using MPI (see [related issue](https://github.com/JuliaIO/HDF5.jl/issues/619#issuecomment-619406315)). In that case, one wants to call `h5_close` before `MPI.Finalize` to exit the process cleanly.

jipolanco:
This fixes issue when a property list is closed by a call to `h5_close` before its `HDF5Properties` wrapper being destroyed by the GC.

Fixes #620

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -832,7 +832,9 @@ end
 
 function close(obj::HDF5Properties)
     if obj.toclose && obj.id != -1
-        h5p_close(obj.id)
+        if isvalid(obj)
+            h5p_close(obj.id)
+        end
         obj.id = -1
     end
     nothing
--- a/test/gc.jl
+++ b/test/gc.jl
@@ -63,6 +63,11 @@ for i = 1:10
     close(file)
 end
 GC.enable(true)
+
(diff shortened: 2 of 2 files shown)


musm:
Thank you! can you add the test in the issue to prevent regressions?

jipolanco:
Done!

musm:
Thanks again. Will merge when CI is green.

jipolanco:
Curiously the test after gc.jl fails... I'll try to fix this.

Issue #619, opened 2020-04-25
Update MPI I/O support + Requires


jipolanco:
The `test/mpio.jl` script fails with recent versions of MPI.jl.

While I started working on a fix, I was thinking that it would be nice to have a friendlier way of calling MPI-IO functions such as `h5p_set_fapl_mpio`. Right now, one has to pass low-level `MPI_Comm` and `MPI_Info` objects that must be reinterpreted as `Hmpih32` or `Hmpih64`, depending on their size, which makes things a bit complicated for the user. In my opinion, it would be great to be able to directly pass a `MPI.Comm` or a `MPI.Info` object, which are wrapper types defined by MPI.jl.

I started working on a solution that uses [Requires](https://github.com/JuliaPackaging/Requires.jl) to define functions such as `h5p_set_fapl_mpio` only if MPI was loaded before HDF5. This allows to have access to types such as `MPI.Comm`, and define specific variants of MPI-IO functions that accept these types. And thanks to Requires, MPI.jl is not added to the dependency list of HDF5.jl.

Before I submit a PR, I wanted to know if there's interest in this kind of solution.

To be a bit more explicit, this is what part of the PR would look like:
```julia
function h5p_set_fapl_mpio(fapl_id, comm::MPI.Comm, info::MPI.Info)
    h5comm = h5_mpihandle(comm.val)
    h5info = h5_mpihandle(info.val)
    h5p_set_fapl_mpio(fapl_id, h5comm, h5info)
end

h5p_set_fapl_mpio(fapl_id, comm::Hmpih32, info::Hmpih32) =
    h5p_set_fapl_mpio32(fapl_id, comm, info)
h5p_set_fapl_mpio(fapl_id, comm::Hmpih64, info::Hmpih64) =
    h5p_set_fapl_mpio64(fapl_id, comm, info)
```

ufdup:
Yes, this looks nice but maybe we should get the test/mpio.jl running first. As of now it is broken due to 2 things: The first is related to how MPI.jl handles communicators nowadays (C-based), and it's an easy fix. The second (I believe) has to do with the library not being able to release objects properly (check the section on finalizers in https://juliaparallel.github.io/MPI.jl/stable/usage/).

After updating the communicator and the info variables to be in-line with MPI.jl, I still get errors like
~~~
#001: H5I.c line 1300 in H5I_dec_app_ref(): can't decrement ID ref count
major: Object atom`
minor: Unable to decrement reference count
~~~

[...]

PR #622, opened 2020-04-26
Lazy-load MPI using Requires

jipolanco:
As discussed in #619, this proposal modifies parallel HDF5 support, defining certain MPI I/O specific functions only if MPI.jl was previously loaded. This is done using [Requires](https://github.com/JuliaPackaging/Requires.jl), so that MPI.jl is not added to the dependency list of HDF5.jl.

The main advantage of this approach is that we now have access to types such as `MPI.MPI_Comm` (a C object) and `MPI.Comm` (its Julia wrapper). In particular, we know their sizes, which simplifies the previous implementation, and solves some (minor) type inference issues.

MPI I/O functions such as `h5p_set_fapl_mpio` and `h5p_get_fapl_mpio` now have definitions that accept (or return) MPI wrappers, which are properly managed by MPI.jl. This greatly simplifies opening parallel HDF5 files, as `MPI.Comm` and `MPI.Info` objects no longer need to be converted to their HDF5 counterparts by the user.

In particular, the following is now possible with the proposed changes:

```julia
comm = MPI.COMM_WORLD
info = MPI.Info()

h5open("filename.h5", "w", "fapl_mpio", (comm, info)) do ff
    # do stuff...
end
```

In its current form, this proposal breaks compatibility with previous versions with regard to parallel HDF5 support, since the `mpihandles` dict and `h5_mpihandle` function were removed. If this is not acceptable, I guess these could be kept for backwards compatibility. It also breaks compatibility in a second way (very minor, in my opinion): due to Requires, MPI necessarily needs to be imported before HDF5 for parallel I/O to work.

Finally, this updates the `test/mpio.jl` script (which was broken under recent MPI.jl versions) to use the new functionality.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -362,6 +362,116 @@ HDF5 is a large library, and the low-level wrap is not complete. However, many o
 Note that Julia's HDF5 directly uses the "2" interfaces, e.g., `H5Dcreate2`, so you need to have version 1.8 of the HDF5 library or later.
 
 
+## System-provided HDF5 libraries
+
+Starting from Julia 1.3, the HDF5 binaries are by default downloaded using the
+[HDF5_jll](https://github.com/JuliaBinaryWrappers/HDF5_jll.jl) package.
+To use system-provided HDF5 binaries instead, set the environment variable
+`JULIA_HDF5_LIBRARY_PATH` to the HDF5 library path and then run
+`Pkg.build("HDF5")`.
+This is in particular needed for [parallel HDF5 support](@ref Parallel-HDF5),
+which is not provided by the `HDF5_jll` binaries.
+
+For example, you can set `JULIA_HDF5_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/hdf5/mpich/`
+if you're using the system package [`libhdf5-mpich-dev`](https://packages.ubuntu.com/focal/libhdf5-mpich-dev)
+on Ubuntu 20.04.
+
+
+## Parallel HDF5
+
+It is possible to read and write [parallel
+HDF5](https://portal.hdfgroup.org/display/HDF5/Parallel+HDF5) files using MPI.
+For this, the HDF5 binaries loaded by HDF5.jl must have been compiled with
+parallel support, and linked to the specific MPI implementation that will be used for parallel I/O.
+
+Parallel-enabled HDF5 libraries are usually included in computing clusters and
+linked to the available MPI implementations.
+They are also available via the package manager of a number of Linux
+distributions.
+
+Finally, note that the MPI.jl package is lazy-loaded by HDF5.jl
+using [Requires](https://github.com/JuliaPackaging/Requires.jl).
+In practice, this means that in Julia code, `MPI` must be imported *before*
+`HDF5` for parallel functionality to be available.
+
+### Setting-up Parallel HDF5
+
+The following step-by-step guide assumes one already has access to
+parallel-enabled HDF5 libraries linked to an existent MPI installation.
+
+#### 1. Using system-provided MPI libraries
+
+Set the environment variable `JULIA_MPI_BINARY=system` and then run
+`]build MPI` from Julia.
+For more control, one can also set the `JULIA_MPI_PATH` environment variable
+to the top-level installation directory of the MPI library.
+See the [MPI.jl
+docs](https://juliaparallel.github.io/MPI.jl/stable/configuration/#Using-a-system-provided-MPI-1)
+for details.
+
+#### 2. Using parallel HDF5 libraries
+
+As detailed in [System-provided HDF5 libraries](@ref), set the
+`JULIA_HDF5_LIBRARY_PATH` environment variable to the directory where
+the HDF5 libraries compiled with parallel support are found.
+
+#### 3. Loading MPI-enabled HDF5
+
+In Julia code, MPI.jl must be loaded *before* HDF5.jl for MPI functionality to
+be available:
+
+```julia
+using MPI
+using HDF5
+```
+
+### Reading and writing data in parallel
+
+A parallel HDF5 file may be opened by passing a `MPI.Comm` (and optionally a
+`MPI.Info`) object to [`h5open`](@ref).
+For instance:
+
+```julia
+comm = MPI.COMM_WORLD
+info = MPI.Info()
+ff = h5open(filename, "w", comm, info)
+```
+
+MPI-distributed data is typically written by first creating a dataset
+describing the global dimensions of the data.
+The following example writes a `10 × Nproc` array distributed over `Nproc` MPI
+processes.
+
+```julia
+Nproc = MPI.Comm_size(comm)
+myrank = MPI.Comm_rank(comm)
+M = 10
+A = fill(myrank, M)  # local data
+dims = (M, Nproc)    # dimensions of global data
+
+# Create dataset
+dset = d_create(ff, "/data", datatype(eltype(A)), dataspace(dims))
+
+# Write local data
+dset[:, myrank + 1] = A
+```
+
+Note that all MPI processes must call `d_create` with the same arguments.
+
+Sometimes, it may be more efficient to write data in chunks, so that each
+process writes to a separate chunk of the file.
+This is especially the case when data is uniformly distributed among MPI
+processes.
+In this example, this can be achieved by passing `chunk=(M, 1)` to `d_create`.
+
+For better performance, it is sometimes preferable to perform [collective
+I/O](https://portal.hdfgroup.org/display/HDF5/Introduction+to+Parallel+HDF5)
+when reading and writing datasets in parallel.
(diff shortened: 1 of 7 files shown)


musm:
We plan to make a breaking release soon.  Perhaps this should go on?

src/HDF5.jl
```diff
@@ -2572,6 +2555,18 @@ const ASCII_ATTRIBUTE_PROPERTIES = Ref{HDF5Properties}()
 
 const DEFAULT_PROPERTIES = HDF5Properties(H5P_DEFAULT, false, H5P_DEFAULT)
 
+const HAS_PARALLEL = Ref(false)
```
musm, line 2558: Does this need to be a  constant ref?

src/mpio.jl
```diff
@@ -0,0 +1,61 @@
+using .MPI
```
musm, line 1:
```suggestion
using MPI
```
?

musm:
I'm not too familiar with the MPI stuff, but seems reasonable. Perhaps some documentation would be in order as well.

jipolanco:
Thanks for looking at this. I just rebased to the current master.

Tests don't pass yet, as I'm still working on adapting things to the changes in #632 (which are a great improvement!).

I'll also work on the docs.

src/mpio.jl
jipolanco: Should be `using .MPI`. That's just how Requires works.

From their [README](https://github.com/JuliaPackaging/Requires.jl/blob/master/README.md):

> In the @require block, or any included files, you can use or import the package, but note that you must use the syntax `using .Gadfly` or `import .Gadfly`, rather than the usual syntax. Otherwise you will get a warning about Gadfly not being in dependencies.

src/HDF5.jl
jipolanco: That value gets modified to `Ref(true)` if the `mpio.jl` file gets loaded via Requires, and if the HDF5 libraries include parallel support.

https://github.com/JuliaIO/HDF5.jl/blob/a12b66f9af28a76cbf7b099ed81709187744767f/src/mpio.jl#L4-L7

I thought it was a simple way to allow for this kind of dual behaviour (depending on whether code is loaded or not), but there may be a better way to do this.

src/HDF5.jl
```diff
@@ -709,7 +709,7 @@ function h5read(filename, name::String; pv...)
     local dat
     fid = h5open(filename, "r"; pv...)
     try
-        obj = fid[name, pv...]
```
musm, line 713: Could we split this out to a separate PR  ?

musm: cc @kleinhenz  I think we missed this in the big update

jipolanco: I thought about that... I agree.

jipolanco:
Well, the parallel HDF5 tests (`test/mpio.jl`) are not run in CI, because they require MPI-enabled HDF5 libraries. This is not the case for the default HDF5 libraries provided by BinaryBuilder. It is however possible to test parallel HDF5 on Travis (I have a package where [this is done](https://github.com/jipolanco/PencilFFTs.jl/blob/master/.travis.yml)).

The proposed interface is already easier to use, but I think we can do better. For the moment I tried to follow the HDF5 C API (e.g. by explicitly passing `*_mpio` property lists), but it's true that we can take inspiration from h5py. For instance, it would make sense to have a `h5open(filename, [mode], comm::MPI.Comm, [info::MPI.Info])` method that would create a parallel HDF5 file. We could ask @simonbyrne to review this, if he's interested.

As for the documentation, I already wrote a bit on parallel HDF5 in the [docs for my PencilFFTs.jl package](https://jipolanco.github.io/PencilFFTs.jl/latest/PencilIO/), which wraps the existent HDF5.jl interface. I think it would make sense to transfer part of those docs to HDF5.jl (and I'd have no problem with doing that).

src/HDF5.jl
jipolanco: Done, see #651

musm:
Looking at this again, it is certainly an improvement even if I'm not familiar with the MPI code, I can tell that. 

How close do you think this PR is? Are we just waiting on documentation?

jipolanco:
Sorry for the delay, I was a bit busy with other things.

I just added a `h5open` variant for opening files in parallel more conveniently, as I proposed in my previous post.

I also added a section on parallel HDF5 to the docs. It includes a guide on how to set-up parallel HDF5, as well as a small example. To make things more consistent, I also copied a section of the README on how to use custom HDF5 libraries (which is needed for parallel HDF5) into the docs.

Personally now I'm happy with the changes, and I'd say this is ready to be merged.

src/mpio.jl
```diff
@@ -0,0 +1,94 @@
...
+Equivalent to `h5open(filename, mode; fapl_mpio=(comm, info), pv...)`.
+Throws an informative error if the loaded HDF5 libraries do not include parallel
+support.
+
+See the [HDF5 docs](https://portal.hdfgroup.org/display/HDF5/H5P_SET_FAPL_MPIO)
+for details on the `comm` and `info` arguments.
+"""
+function h5open(
+        filename::AbstractString, mode::AbstractString,
+        comm::MPI.Comm, info::MPI.Info = MPI.Info(); pv...
```
musm, line 87: I'm thinking these should actually be keyword arguments.  Would `driver` be a more appropriate name for `info` ?

docs/src/index.md
```diff
@@ -362,6 +362,116 @@ HDF5 is a large library, and the low-level wrap is not complete. However, many o
...
+myrank = MPI.Comm_rank(comm)
+M = 10
+A = fill(myrank, M)  # local data
+dims = (M, Nproc)    # dimensions of global data
+
+# Create dataset
+dset = d_create(ff, "/data", datatype(eltype(A)), dataspace(dims))
+
+# Write local data
+dset[:, myrank + 1] = A
```
musm, line 456: Why is ` +1 ` always passed when dealing with rank? Should the library automatically add think?

musm:
Great thanks, docs look great!

src/mpio.jl
jipolanco: I guess you're thinking about the h5py interface. Note that the `driver` argument in h5py is different from the `info` argument here, which [allows to pass "hints"](https://portal.hdfgroup.org/display/knowledge/How+to+pass+hints+to+MPI+from+HDF5) to the MPI implementation. This can be very important for I/O performance in large computing clusters. I'm a bit surprised that h5py doesn't support this (or maybe it does, but it's not mentioned in the docs).

In my opinion, having a `driver` argument is redundant. If the user passes MPI objects to `h5open`, then it should be clear that they want to perform parallel I/O. I also like the fact that we can take advantage of multiple dispatch to provide a `h5open` variant taking MPI objects, without the need for keyword arguments.

docs/src/index.md
jipolanco: That's an issue coming from the MPI side, and I think there's not much we can do about it.

In MPI.jl, the process ranks returned by `MPI.Comm_rank` are in `0:Nproc-1`. Although this is not the most natural choice in Julia, it is consistent with the MPI C and Fortran APIs, and MPI.jl users are usually aware of this.

From the HDF5 side, we can't assume that the index passed to `setindex!` is an MPI rank, as it could be anything else.

src/mpio.jl
musm: Yeah I was. 

Regarding the arguments, should these be keyword arguments instead of positional. I feel like they should be.

jipolanco: In my opinion they should be positional arguments, but I guess that's just personal preference. I'd say it's more natural in Julia, as it takes advantage of multiple dispatch to enable different functionalities.

More practically, an advantage of using positional arguments is that Julia doesn't perform dispatch based on keyword arguments. Therefore, if we were to provide a `h5open(args...; comm, info, etc...)` method, then the new keyword arguments `comm` and `info` would need to be added to the main `h5open` variants in `src/HDF5.jl`. These variants are not aware of MPI, so we cannot set e.g. `info = MPI.Info()` as default value, and we would need to set default values such as `comm = nothing, info = nothing`. In the end, I think this would just increase complexity and visual noise in the main HDF5.jl file, and would go against the idea of putting all MPI-related code in a separate file.

musm: Thanks for the clarification. So the main argument for keeping them positional,  is that default values can be passed with values that depend on `MPI`  as their defaults, and we obviously can't have these as default values if we made them keyword arguments.

musm: I guess keyword based dispatch (a la https://github.com/simonbyrne/KeywordDispatch.jl )  would help since that would free the keywords `comm, info` since they would dispatch on `MPI` types, but it also seems perhaps out of scope.

musm:
Thanks again, and sorry this took so long to review and merge. Keep bumping until you force me to do it next time 😄

jipolanco:
Great, thanks for merging!

Issue #537, opened 2018-11-19
Memory leak with arrays in conjuciton with HDF5


fremling:
I'm quite new to julia, but i've noticed that there seems to be a memory leak in the HDF5 package in julia version 1.0.2 with the "HDF5" package is 0.10.2.

If i run the following code (Ubuntu 18.04)

    using HDF5
    using LinearAlgebra
    function HDF5MemoryLeak1(Num)
        MyArray=randn(1000,100)
        fid=h5open("testfile.hdf5","w")
        close(fid)
    end
    function HDF5MemoryLeak2(Num)
        #MyArray=randn(1000,100)
        fid=h5open("testfile.hdf5","w")
        close(fid)
    end
    function HDF5MemoryLeak3(Num)
        MyArray=randn(1000,100)
        #fid=h5open("testfile.hdf5","w")
        #close(fid)
    end
    function HDF5MemoryLeak4(Num)
        NyMarray=randn(1000,100)
        NyMarray=nothing
        fid=h5open("testfile.hdf5","w")
        close(fid)
    end

    N=2*10^3
    for j in 1:N
        println("$j of $N")
        HDF5MemoryLeak(j)
    end

If i run this file and replace in the loops HDF5MemoryLeak by HDF5MemoryLeak1, HDF5MemoryLeak2 or HDF5MemoryLeak3 i get the following reads on top

    <Call_Name>         VIRT      RES    SHR 
    HDF5MemoryLeak1 12332192   1.683g  80252
    HDF5MemoryLeak2   802256   235908  79020
    HDF5MemoryLeak3   810720   242440  79500
    HDF5MemoryLeak4    2532724 1.720g 106868

So it looks like one their own, the hdf5 module and the linear algebra package to not leak memory, but together they do?

Ive seen this discussion https://github.com/JuliaIO/HDF5.jl/issues/349 on memory leaks, but i'm not sure this is the same issue.

ludvigak:
I believe I'm seeing the same thing (just posted in #349).

musm:
Leaks 1 to 3 are fixed by PR https://github.com/JuliaIO/HDF5.jl/pull/629

musm:
closed by https://github.com/JuliaIO/HDF5.jl/pull/629 thanks to @MarkusSchildhauer  !

PR #629, opened 2020-07-29
Fix property memory leak

MarkusSchildhauer:
The change solves issue #630 in an orderly manner. The public API is not altered. All tests are running through. The snippet in #630 produces no memory leaks anymore.

The change ensures, that Julia `HDF5Properties` are **always** instantiated **with a finalizer** that runs `h5p_close(id)` on the corresponding C HDF5 property list. In other words it's not allowed that valid property list IDs outlive their corresponding Julia objects. This is no problem since the property list objects passed to the constructor functions of the HDF5 C objects (like `H5Dcreate2`) are not used after their creation (HDF5 makes private copies of the options, see e.g. line 129 https://support.hdfgroup.org/ftp//HDF5/prev-releases/hdf5-1.8/hdf5-1.8.21/src/unpacked/src/H5D.c). 
We just have to ensure that the corresponding Julia `HDF5Properties` objects are still referenced when the constructor call happens. Immediately after that the wrapped property lists may be closed along with the Julia garbage collection .
All corresponding function calls have been adapted to the internal interface change (eliminating the `toclose` argument). 
This change especially affects the `HDF5Dataset` which now consequently provides the `HDF5Properties` object instead of a pure ID for the `xfer` property.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -367,7 +367,7 @@ mutable struct HDF5Properties
     class::Hid
     function HDF5Properties(id, class::Hid = H5P_DEFAULT)
         p = new(id, class)
-        finalizer(close, p)
+        finalizer(close, p) #Essential, otherwise we get a memory leak, since closing file with CLOSE_STRONG is not doing it for us
         p
     end
 end
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -362,12 +362,24 @@ function show(io::IO, g::HDF5Group)
     end
 end
 
+mutable struct HDF5Properties
+    id::Hid
+    class::Hid
+    function HDF5Properties(id, class::Hid = H5P_DEFAULT)
+        p = new(id, class)
+        finalizer(close, p)
+        p
+    end
+end
+HDF5Properties() = HDF5Properties(H5P_DEFAULT)
+convert(::Type{Hid}, p::HDF5Properties) = p.id
+
 mutable struct HDF5Dataset
     id::Hid
     file::HDF5File
-    xfer::Hid
+    xfer::HDF5Properties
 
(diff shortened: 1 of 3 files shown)


musm:
Fantastic! Great to see a solid test suite in here. Will merge in a day or to. Please ping if I haven't gotten around to it.

MarkusSchildhauer:
That was fast! Many Thanks!

musm:
NP. I was hoping this would also help fix some other memory leaks we have been seeing aka https://github.com/JuliaIO/HDF5.jl/issues/537 , but those persist and are unrelated.

MarkusSchildhauer:
Why do you think, they still persist? https://github.com/JuliaIO/HDF5.jl/issues/349#issuecomment-313563553 is definitely gone now! I couldn't find a snippet in #349 or #537 that could reproduce a memory leak on macOS or ubuntu with Julia 1.4.2. Before checking the memory one should manually call GC.gc() anyway. Otherwise the results are random. So far as I can see no one did that really. 
For instance #537 has constant memory for me:
~~~Julia
using HDF5, LinearAlgebra
const PRINT_MEMORY = `ps -p $(getpid()) -o rss=,vsz=`
macro memtest(ex)
    println("Memory: RSS     VIRT")
    quote
        for i in 1:100
            for _ in 1:100
                $ex
            end
            # HDF5.h5_garbage_collect()
            GC.gc(); 
            print(rpad(i, 8)); run(PRINT_MEMORY)
        end
    end
end
@memtest h5open("testfile.hdf5","w") do fid
    MyArray=randn(1000,100)
    close(fid)
end
~~~

musm:
You are right! I was missing a GC when testing the memory 😄

PR #632, opened 2020-08-15
RFC: use keyword args for setting properties

kleinhenz:
I decided to try this while looking at #631. This replaces the current method of setting properties using varargs with keyword arguments. This is breaking so we should discuss if it is worth it. In #462 where the current behavior was introduced it looks like this wasn't done because the behavior of keyword arguments was still stabilizing in julia. Now that julia has stabilized I think this is a better interface.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,0 +1,7 @@
+## v0.14
+
+* Properties are now set using keyword arguments instead of by pairs of string and value positional arguments.
+  For example `dset = d_create(h5f, "A", datatype(Int64), dataspace(10,10), "chunk", (3,3))` is now written as
+  `dset = d_create(h5f, "A", datatype(Int64), dataspace(10,10), chunk=(3,3))`. Additionally the key type used for
+  directly setting `HDF5Properties` objects has changed from a `String` to a `Symbol`, e.g.
+  `apl["fclose_degree"] = H5F_CLOSE_STRONG` is now written as `apl[:fclose_degree] = H5F_CLOSE_STRONG` ([#632](https://github.com/JuliaIO/HDF5.jl/pull/632)).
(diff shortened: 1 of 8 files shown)


musm:
This is definitely the correct way to go. I wonder about switching from strings to symbols. Any particular reason to change that? 

I'm not sure on how best to handle the breaking change, but I definitely want to see this in.

kleinhenz:
The key type of the slurped kwargs is a symbol so I just wanted to avoid an unnecessary conversion to a string. This changes the case where you set a property explicitly e.g. `apl[:fclose_degree] = H5F_CLOSE_STRONG` but in the common case where the properties are set by using the kwargs this should be invisible.

musm:
Makes sense. Would it be ok to merge this and make a breaking release? I don't see an obvious way of making this non-breaking or adding a deprecation message given how the previous structure was coded.

musm:
Perhaps we should add a major note in the README

kleinhenz:
Yeah a breaking release is probably the way to go. I can't think of a great way to deprecate this either. I added a `HISTORY.md` file to keep track of changes between versions.

musm:
We'll also need to update the syntax in the `docs`

kleinhenz:
Updated the documentation for this. The new docs look great BTW!

musm:
I think I will first merge  https://github.com/JuliaIO/HDF5.jl/pull/642/commits/978050c8c39f9f6f59eb8cc46a22ba444190293a after reviewing it. Make a minor release. Then merge this and make a breaking release. Any objections?

musm:
Also just want to say this is a massive improvement over the very old style of 'keyword' handling that HDF5 is using (very legacy)

musm:
On second thought I think we should just merge this first and  the following PRs  and include them in the major release. That way if code magically breaks due to someone relying on internals  it won't be an issue.

PR #637, opened 2020-08-17
Update actions and CI

musm:
- Removes coveralls, and uses codecov only.
- Adds docs
- Cleans up CI and adds github actions


musm:
Looks like we have some pre-existing errors on x86 on linux. Going to merge and try to address those in a separate PR.

PR #642, opened 2020-08-20
Fix support for reading more deeply nested types

### LINKED ISSUE #634: Type size mismatch when reading compound datatype
Dear developers,

I have experienced an error "_LoadError: AssertionError: Type sizes mismatch!_" when trying to read a complicated compound datatype containing a integers, reals, and string arrays produced by our simulation code. From what I could find out, this issue seems to be caused by the size of the HDF5.FixedArray{String,(1,),1} datatype not being consistent with the size of the string array in the HDF5 file.
In the attachment, I have provided a small example (a Python script to create the HDF5 file and a Julia script to read it) where the compound datatype just contains a string array. With Julia 1.4.2 and a freshly installed HDF5.jl, it produces the following output:

`ERROR: LoadError: AssertionError: Type sizes mismatch!
Stacktrace:
 [1] read(::HDF5Dataset, ::Type{Array{NamedTuple{(:name,),Tuple{HDF5.FixedArray{String,(1,),1}}},N} where N}) at /home/<username>/.julia/packages/HDF5/cvKpY/src/HDF5.jl:1490
 [2] read(::HDF5Dataset) at /home/<username>/.julia/packages/HDF5/cvKpY/src/HDF5.jl:1306
 [3] read(::HDF5File, ::String, ::Bool) at /home/<username>/.julia/packages/HDF5/cvKpY/src/HDF5.jl:1294
 [4] read(::HDF5File, ::String) at /home/<username>/.julia/packages/HDF5/cvKpY/src/HDF5.jl:1293
 [5] top-level scope at /home/<username>/tests/test_hdf5_jl_bug/read_file.jl:5
 [6] include(::Module, ::String) at ./Base.jl:377
 [7] exec_options(::Base.JLOptions) at ./client.jl:288
 [8] _start() at ./client.jl:484
in expression starting at /home/<username>/tests/test_hdf5_jl_bug/read_file.jl:5
`

I hope that this example helps to figure out the issue. Another observation I made is that with Julia 1.2 and an older version of HDF5.jl, the String arrays are recognized as UInt8 and the compound datatype can be read successfully (however, one needs to decode the String manually).

Thank you for this very useful library.

[example.zip](https://github.com/JuliaIO/HDF5.jl/files/5083647/example.zip)

kleinhenz:
This should close #634. I also fixed unicode support in fixed sized strings which had previously been broken and maybe fixed the ci failure from the `h5d_vlen_reclaim` call although it is still mysterious to me why that failure only appears on some builds.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -2460,16 +2460,6 @@ function vlen_get_buf_size(dset::HDF5Dataset, dtype::HDF5Datatype, dspace::HDF5D
     sz[]
 end
 
-function hdf5array(objtype)
-    nd = h5t_get_array_ndims(objtype.id)
-    dims = Vector{Hsize}(undef,nd)
-    h5t_get_array_dims(objtype.id, dims)
-    eltyp = HDF5Datatype(h5t_get_super(objtype.id))
-    T = get_mem_compatible_jl_type(objtype)
-    dimsizes = ntuple(i -> Int(dims[nd-i+1]), nd)  # reverse order
-    FixedArray{T, dimsizes, prod(dimsizes)}
-end
-
 ### Property manipulation ###
 get_create_properties(d::HDF5Dataset)   = HDF5Properties(h5d_get_create_plist(d.id), H5P_DATASET_CREATE)
 get_create_properties(g::HDF5Group)     = HDF5Properties(h5g_get_create_plist(g.id), H5P_GROUP_CREATE)
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -2376,6 +2376,18 @@ function h5l_get_info(link_loc_id::Hid, link_name::String, lapl_id::Hid)
     info[]
 end
 
+function h5lt_dtype_to_text(dtype_id; max_len=512)
+  buf = Vector{UInt8}(undef, max_len)
+  sz = Csize_t[length(buf)]
+  status = ccall((:H5LTdtype_to_text, libhdf5_hl), Int, (HDF5.Hid, Ptr{UInt8}, Int, Ptr{Csize_t}), dtype_id, buf, 0, sz)
+  if status < 0
+    error("Error getting dtype text representation")
+  end
+  buf[end] = 0 # nullterm
+  dtype_text = rstrip(unsafe_string(pointer(buf)))
+  return dtype_text
+end
+
 # Set MPIO properties in HDF5.
 # Note: HDF5 creates a COPY of the comm and info objects.
 function h5p_set_fapl_mpio(fapl_id, comm, info)
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1491,7 +1491,7 @@ function read(dset::HDF5Dataset, T::Union{Type{Array{U}}, Type{U}}) where U <: N
     h5type_str = h5lt_dtype_to_text(memtype.id)
     error("""
           Type size mismatch
-          sizeof($T) = $(sizeof(T))
+          sizeof($U) = $(sizeof(U))
           sizeof($h5type_str) = $(h5t_get_size(memtype.id))
           """)
   end
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -479,7 +479,7 @@ eltype(::Type{FixedArray{T,D,L}}) where {T,D,L} = T
 eltype(x::T) where T <: FixedArray = eltype(T)
 
 struct FixedString{N, PAD}
-  data::NTuple{N, Cchar}
+  data::NTuple{N, UInt8}
 end
 length(::Type{FixedString{N, PAD}}) where {N, PAD} = N
 length(x::T) where T <: FixedString = length(T)
@@ -1470,7 +1470,7 @@ end
 normalize_types(x) = x
 normalize_types(x::NamedTuple{T}) where T = NamedTuple{T}(map(normalize_types, values(x)))
 normalize_types(x::Cstring) = unsafe_string(x)
-normalize_types(x::FixedString) = unpad(join(Char.(x.data)), pad(x))
+normalize_types(x::FixedString) = unpad(String(collect(x.data)), pad(x))
 normalize_types(x::FixedArray) = reshape(collect(x.data), size(x)...)
 normalize_types(x::VariableArray) = copy(unsafe_wrap(Array, convert(Ptr{eltype(x)}, x.p), x.len, own=false))
 
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -2386,15 +2386,16 @@ function h5l_get_info(link_loc_id::Hid, link_name::String, lapl_id::Hid)
     info[]
 end
 
-function h5lt_dtype_to_text(dtype_id; max_len=512)
-  buf = Vector{UInt8}(undef, max_len)
-  sz = Csize_t[length(buf)]
-  status = ccall((:H5LTdtype_to_text, libhdf5_hl), Int, (HDF5.Hid, Ptr{UInt8}, Int, Ptr{Csize_t}), dtype_id, buf, 0, sz)
-  if status < 0
-    error("Error getting dtype text representation")
-  end
-  buf[end] = 0 # nullterm
-  dtype_text = rstrip(unsafe_string(pointer(buf)))
+function h5lt_dtype_to_text(dtype_id)
+  len = Csize_t[0]
+  status = ccall((:H5LTdtype_to_text, libhdf5_hl), Int, (HDF5.Hid, Ptr{UInt8}, Int, Ptr{Csize_t}), dtype_id, C_NULL, 0, len)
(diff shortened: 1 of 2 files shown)


src/HDF5.jl
```diff
@@ -2239,7 +2263,7 @@ for (jlname, h5name, outtype, argtypes, argsyms, ex_error) in
      (:h5f_open, :H5Fopen, Hid, (Cstring, Cuint, Hid), (:pathname, :flags, :fapl_id), :(error("Error opening file ", pathname))),
      (:h5g_create, :H5Gcreate2, Hid, (Hid, Ptr{UInt8}, Hid, Hid, Hid), (:loc_id, :pathname, :lcpl_id, :gcpl_id, :gapl_id), :(error("Error creating group ", h5i_get_name(loc_id), "/", pathname))),
      (:h5g_get_create_plist, :H5Gget_create_plist, Hid, (Hid,), (:group_id,), :(error("Error getting group create property list"))),
-     (:h5g_get_objname_by_idx, :H5Gget_objname_by_idx, Hid, (Hid, Hsize, Ptr{UInt8}, Csize_t), (:loc_id, :idx, :pathname, :size), :(error("Error getting group object name ", h5i_get_name(loc_id), "/", pathname))),
+     (:h5g_get_objname_by_idx, :H5Gget_objname_by_idx, Cssize_t, (Hid, Hsize, Ptr{UInt8}, Csize_t), (:loc_id, :idx, :pathname, :size), :(error("Error getting group object name ", h5i_get_name(loc_id), "/", pathname))),
```
musm, line 2266: Best if this was a separate PR

src/HDF5.jl
```diff
@@ -2386,15 +2386,16 @@ function h5l_get_info(link_loc_id::Hid, link_name::String, lapl_id::Hid)
...
-  sz = Csize_t[length(buf)]
-  status = ccall((:H5LTdtype_to_text, libhdf5_hl), Int, (HDF5.Hid, Ptr{UInt8}, Int, Ptr{Csize_t}), dtype_id, buf, 0, sz)
-  if status < 0
-    error("Error getting dtype text representation")
-  end
-  buf[end] = 0 # nullterm
-  dtype_text = rstrip(unsafe_string(pointer(buf)))
+function h5lt_dtype_to_text(dtype_id)
+  len = Csize_t[0]
+  status = ccall((:H5LTdtype_to_text, libhdf5_hl), Int, (HDF5.Hid, Ptr{UInt8}, Int, Ptr{Csize_t}), dtype_id, C_NULL, 0, len)
```
musm, line 2391: This ensures the buffer is always large enough, I'm assuming?

src/HDF5.jl
```diff
@@ -1484,21 +1484,31 @@ do_reclaim(::Type{T}) where T <: Union{Cstring, VariableArray} = true
...
+          sizeof($T) = $(sizeof(T))
+          sizeof($h5type_str) = $(h5t_get_size(memtype.id))
+          """)
+  end
 
   buf = Array{U}(undef, size(dset))
+  memspace = dataspace(buf)
 
-  h5d_read(dset.id, memtype_id, H5S_ALL, H5S_ALL, H5P_DEFAULT, buf)
+  h5d_read(dset.id, memtype.id, memspace.id, HDF5.H5S_ALL, dset.xfer.id, buf)
```
musm, line 1502:
```suggestion
  h5d_read(dset.id, memtype.id, memspace.id, H5S_ALL, dset.xfer.id, buf)
```

src/HDF5.jl
```diff
@@ -2362,6 +2386,19 @@ function h5l_get_info(link_loc_id::Hid, link_name::String, lapl_id::Hid)
     info[]
 end
 
+function h5lt_dtype_to_text(dtype_id)
+  len = Csize_t[0]
+  status = ccall((:H5LTdtype_to_text, libhdf5_hl), Int, (HDF5.Hid, Ptr{UInt8}, Int, Ptr{Csize_t}), dtype_id, C_NULL, 0, len)
```
musm, line 2391:
```suggestion
  status = ccall((:H5LTdtype_to_text, libhdf5_hl), Int, (Hid, Ptr{UInt8}, Int, Ptr{Csize_t}), dtype_id, C_NULL, 0, len)
```

src/HDF5.jl
```diff
@@ -2362,6 +2386,19 @@ function h5l_get_info(link_loc_id::Hid, link_name::String, lapl_id::Hid)
...
+  status < 0 && error("Error getting dtype text representation")
+
+  buf = Vector{UInt8}(undef, len[1])
+  status = ccall((:H5LTdtype_to_text, libhdf5_hl), Int, (HDF5.Hid, Ptr{UInt8}, Int, Ptr{Csize_t}), dtype_id, buf, 0, len)
```
musm, line 2395:
```suggestion
  status = ccall((:H5LTdtype_to_text, libhdf5_hl), Herr, (Hid, Ptr{UInt8}, Int, Ptr{Csize_t}), dtype_id, buf, 0, len)
```

src/HDF5.jl
```diff
@@ -2362,6 +2386,19 @@ function h5l_get_info(link_loc_id::Hid, link_name::String, lapl_id::Hid)
...
+  status < 0 && error("Error getting dtype text representation")
+
+  dtype_text = unsafe_string(pointer(buf))
+  return dtype_text
+end
```
musm, line 2400:
```suggestion
function h5lt_dtype_to_text(dtype_id)
    len = Ref{Csize_t}()
    status = ccall((:H5LTdtype_to_text, libhdf5_hl), Herr, (Hid, Ptr{UInt8}, Int, Ref{Csize_t}), dtype_id, C_NULL, 0, len)
    status < 0 && error("Error getting dtype text representation")

    buf = Vector{UInt8}(undef, len[])
    status = ccall((:H5LTdtype_to_text, libhdf5_hl), Herr, (Hid, Ptr{UInt8}, Int, Ref{Csize_t}), dtype_id, buf, 0, len)
    status < 0 && error("Error getting dtype text representation")

    dtype_text = unsafe_string(pointer(buf))
    return dtype_text
end
```

src/HDF5.jl
```diff
@@ -1484,21 +1484,31 @@ do_reclaim(::Type{T}) where T <: Union{Cstring, VariableArray} = true
...
 function read(dset::HDF5Dataset, T::Union{Type{Array{U}}, Type{U}}) where U <: NamedTuple
   filetype = datatype(dset)
-  memtype_id = h5t_get_native_type(filetype.id)  # padded layout in memory
-  @assert sizeof(U) == h5t_get_size(memtype_id) "Type sizes mismatch!"
+  memtype = HDF5Datatype(h5t_get_native_type(filetype.id))  # padded layout in memory
+  close(filetype)
+
+  if sizeof(U) != h5t_get_size(memtype.id)
+    h5type_str = h5lt_dtype_to_text(memtype.id)
+    error("""
```
musm, line 1492: When I try to trigger this error (without the corresponding fix to reading nested types)

I get an error on the julia size from `sizeof(T)`: 
`Argument is an incomplete Array type and does not have a definite size.`

src/HDF5.jl
```diff
@@ -1945,6 +1945,40 @@ function hdf5_to_julia_eltype(objtype)
...
+        super_id = h5t_get_super(objtype.id)
+        T = HDF5Vlen{hdf5_to_julia_eltype(HDF5Datatype(super_id))}
+    elseif class_id == H5T_COMPOUND
+        T = get_mem_compatible_jl_type(objtype)
+    elseif class_id == H5T_ARRAY
+        T = get_mem_compatible_jl_type(objtype)
+    else
+        error("Class id ", class_id, " is not yet supported")
+    end
+    T
```
musm, line 1979:
```suggestion
    return T
```

src/HDF5.jl
```diff
@@ -2019,20 +2031,22 @@ function hdf5_to_julia_eltype(objtype)
...
+          return Complex{membertypes[1]}
         else
-            T = NamedTuple{Symbol.(membernames), Tuple{membertypes...}}
+          return NamedTuple{Symbol.(membernames), Tuple{membertypes...}}
         end
     elseif class_id == H5T_ARRAY
-        T = hdf5array(objtype)
-    else
-        error("Class id ", class_id, " is not yet supported")
+        nd = h5t_get_array_ndims(objtype.id)
```
musm, line 2049: Delete `hdf5array` ?

src/HDF5.jl
kleinhenz: Good catch. Should have been `U` instead of `T`. Should be fixed now.

src/HDF5.jl
kleinhenz: yeah deleting it probably makes sense as long as we don't think many people are using it. I don't really like the name since it seems pretty unclear exactly what to expect `hdf5array` to do.

musm:
Great, thanks!

PR #645, opened 2020-08-22
Add ability to set dataset create-time allocation policy

jmert:
This PR adds the `H5D_alloc_time_t` constants and getter/setter functions to control the dataset allocation policy, as described by https://portal.hdfgroup.org/display/HDF5/H5P_SET_ALLOC_TIME.

This is built upon #644 since I found that bug while learning about early vs late allocation, and I've got a use case where I'd like to try using early allocation with mmaped arrays. I also see #632 is proposed to overhaul the keyword handling a bunch, so I can rebase this change on that if it lands first.

~~I haven't added any tests that the `get_alloc_time()` function works since I didn't see any obvious place them, but I could create a new file/testset just for setting and getting property values if desired (which will probably be needed if more of the properties such as fill value and fill time are hooked up later).~~ Added in a new commit.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -184,6 +184,31 @@ Note that `readmmap` returns an `Array` rather than an HDF5 object.
 **Note**: if you use `readmmap` on a dataset and subsequently close the file, the array data are still available---and file continues to be in use---until all of the arrays are garbage-collected.
 This is in contrast to standard HDF5 datasets, where closing the file prevents further access to any of the datasets, but the file is also detached and can safely be rewritten immediately.
 
+Under the default
+[allocation-time policy](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetAllocTime),
+a newly added `ismmappable` dataset can only be memory mapped after it has been written
+to.
+The following fails:
+```julia
+vec_dset = d_create(g, "v", datatype(Float64), dataspace(10_000,1))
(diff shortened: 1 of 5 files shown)


musm:
Will need to be rebased otherwise LGTM

musm:
Should we add some simple docs  on this at https://github.com/JuliaIO/HDF5.jl/blob/master/docs/src/index.md

jmert:
Rebased to use properties, and I've added a short description to the Memory Mapping section on why you might want to change the allocation-time policy.

One thing I've left out is any mention of the fact that early allocation in some cases is still very cheap due to OS/filesystem support for [sparse files](https://en.wikipedia.org/wiki/Sparse_file) because (a) that seems kind of out of the scope of this package's documentation (even thought it's quite useful to know), and (b) I have no clue (since I only use Linux) how sparse files are handled across Mac/Windows filesystems in concert with whatever HDF5 library is used there.

musm:
Looks great, 


> One thing I've left out is any mention of the fact that early allocation in some cases is still very cheap due to OS/filesystem support for sparse files because (a) that seems kind of out of the scope of this package's documentation (even though it's quite useful to know), and (b) I have no clue (since I only use Linux) how sparse files are handled across Mac/Windows filesystems in concert with whatever HDF5 library is used there.

Right, it does seem a little out of scope. It's always possible to add a footnote in that section regarding this important point.

PR #646, opened 2020-08-23
Explicitly generated ccall wrappers

musm:
Currently a macro generates all the ccall wrappers.

In this PR, we explicitly save the generated ccall wrappers to a file called `api.jl` instead. 

IMO this makes debugging way easier, along with code introspection and tooling.


musm:
cc @timholy 

(sorry for pinging you on this repo which you no longer maintain, but I'm hoping to get at least some feedback before making such a change)

musm:
cc @kleinhenz @jmert  on thoughts on such a change

jmert:
I think the change is a good way forward. In the future, maybe an auto-wrapping like CUDA does (https://github.com/JuliaGPU/CUDA.jl/tree/master/res/wrap) based on the `*.h` headers themselves could replace the manual lists, but for now I think just separating the raw ccall wrappers from anything else that has some custom logic is a good step.

musm:
Oh nice, yes it seems that they use `Clang.jl` which we should probably also look into using for this purpose as well. I agree for the future we should look into this upgrade.

kleinhenz:
Yeah this seems like a good idea. In general I like splitting out the wrapper parts of the library from the julia interface parts. Anything to make `HDF5.jl` slightly less huge is nice.

jmert:
I felt like playing around with macro tools today, so I took a stab at replacing the big `for` loops with a single implementation which is capable of making wrapper functions for the three independent cases (no return value, boolean return value, and returning the C function's output). My branch is at https://github.com/JuliaIO/HDF5.jl/compare/master...jmert:wrapper.

I haven't actually filled out all the documentation comments/docstrings that it will ultimately need, but it was an interesting exercise alone in revealing some odd naming of functions. From the `bind_generator.jl` script where some explicit naming exceptions are defined, the following 5 are oddities:
```julia
push!(bind_exceptions, :h5a_delete_by_idx => :H5delete_by_idx)
push!(bind_exceptions, :h5a_delete_by_name => :H5delete_by_name)
push!(bind_exceptions, :h5d_oappend => :H5DOappend)
push!(bind_exceptions, :h5f_get_intend => :H5Fget_intent)
push!(bind_exceptions, :hf5start_swmr_write => :H5Fstart_swmr_write)
```

The first two should be `H5A*` functions according to the documentation, and the symbols don't appear to exist either, so obviously these aren't being tested:
```julia
julia> using Libdl, HDF5

julia> Libdl.dlsym(HDF5.HDF5_jll.libhdf5_handle, :H5delete_by_name)
ERROR: could not load symbol "H5delete_by_name":
...
```

The final three just appear to be weird misspellings of the intended names. They could easily be renamed with a deprecated binding to aid any external users.

musm:
That's pretty nice actually!.  If we are going this route, should we just use Clang.jl? That might be more effort than it's worth since we would still probably require to customize it to spit the error messages we want etc.  Haven't fully reviewed it, but It's a positive improvement, and better than this PR.

jmert:
> That's pretty nice actually!. If we are going this route, should we just use Clang.jl? That might be more effort than it's worth since we would still probably require to customize it to spit the error messages we want etc. Haven't fully reviewed it, but It's a positive improvement, and better than this PR.

As you say, the disadvantage of using a Clang.jl-based auto-generated list of bindings is that you lose the custom error messages. I threw that out there as an idea just because I knew there was precedent for it, but I'm not sure there's much to gain from it either (and it certainly adds an extra complication w.r.t. generating against the correct set of headers — for instance, my machine has the HDF5 v1.12 headers in `/usr/include` whereas HDF5_jll is still using v1.10, so that'd add a barrier wherein running the generator would require getting a copy of the correct library version headers).

I can make a new PR with my branch if you'd like (otherwise you could also just pull my branch onto yours and keep using this PR — whatever's easiest).

musm:
Either works for me, whatever is easier for you.

PR #649, opened 2020-08-25
Misc minor improvements

musm:
- Use StringVector for buffer

- h5g_get_objname_by_idx is deprecated in favor of h5l_get_name_by_idx

- Simplify h5a_get_name_by_idx logic

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -76,6 +76,7 @@ const H5_ITER_N       = 3
 const H5_INDEX_UNKNOWN   = -1
 const H5_INDEX_NAME      = 0
 const H5_INDEX_CRT_ORDER = 1
+const H5_INDEX_N = 2
 # dataset constants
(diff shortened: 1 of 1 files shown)


src/HDF5.jl
```diff
@@ -2369,11 +2368,11 @@ function h5lt_dtype_to_text(dtype_id)
     status = ccall((:H5LTdtype_to_text, libhdf5_hl), Herr, (Hid, Ptr{UInt8}, Int, Ref{Csize_t}), dtype_id, C_NULL, 0, len)
     status < 0 && error("Error getting dtype text representation")
 
-    buf = Vector{UInt8}(undef, len[])
+    buf = Base.StringVector(len[]-1)
```
musm, line 2371: cc @kleinhenz  
I took another look at this and realized it's much better to allocate string buffers as follows. Reduces an allocation and this is more idiomatic Julia

PR #651, opened 2020-08-25
Fix passing properties in read and h5read

jipolanco:
I'm not sure if it's possible to test this. I'm not aware of any HDF5 property that is relevant for reading files and datasets, and that is currently supported by HDF5.jl, other than MPI-IO related properties.


musm:
> I'm not sure if it's possible to test this. I'm not aware of any HDF5 property that is relevant for reading files and datasets, and that is currently supported by HDF5.jl, other than MPI-IO related properties.

Fair enough, good that you caught this regardless!

PR #652, opened 2020-08-26
Re-haul and implement a more generic read

### LINKED ISSUE #169: Improve reading of plain composite types
#168 reminded me that reading of composite types could probably be improved for "plain" HDF5 files. Rather than returning an opaque object, it would seem desirable to let the user specify the object type. One problem I see is that it breaks the API that's been established for all other operations.

I wonder whether `read(file, ("varname", T))`, i.e., grouping the name and type together in a tuple, would be a reasonable way to minimize trouble. Thoughts? I don't have a lot of time right now, this is just planning for the future.

### LINKED ISSUE #625: Colon indexing should work for non bits types
I have a 1D HDF5 string dataset `"values"` in my `fid::HDF5File`

If I do `read(fid, "values")`, it gives them back OK. If I do `fid["values"][:]`, it tells me that `ERROR: Dataset indexing (hyperslab) is available only for bits types`.

Would it be possible to special-case `getindex(::HDF5Dataset, ::Colon)` to ignore the bitstypes restrictions, since it is equivalent to `read` as there isn't actually any slicing going on?

### LINKED ISSUE #627: EXCEPTION_ACCESS_VIOLATION on Windows
I've run into an `EXCEPTION_ACCESS_VIOLATION` that I can reproduce on Windows 10, but not my Mac.  I'm unfortunately unable to share the file/code to reproduce it.

The problem is connected to `unsafe_wrap` (discovered through the assistance of @JeffBezanson) and I was able to get past the error by changing the keyword arg to `own=false` in the two functions below:

https://github.com/JuliaIO/HDF5.jl/blob/1d740136bf7225bd05e7a561fd8721c6ffedeeab/src/HDF5.jl#L1541-L1546

Does `own=false` likely cause a memory leak here?

kleinhenz:
This adds a new generic (not type constrained) read method

```
read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
```

which reads a hyperslab of a dataset or an attribute into an array of type T.

It also adds convenience methods 

```
read(parent::Union{HDF5File, HDF5Group}, name_type_pair::Pair{String, DataType}; pv...)
h5read(filename, name_type_pair::Pair{String, DataType}; pv...)
```

which can be used like `v_read = read(h5f, "data"=>Simple)` or `v_read = h5read(fn, "data"=>Simple)` to read
`data` into an array with element type `Simple` (see `test/custom.jl`). This provides the functionality of #559 in a way that is more integrated with the rest of the library than adding a new separate `h5read_compound` method and closes #169.

Although the function does not have a type constraint I do check that the type is the correct size so this should still prevent you from shooting yourself in the foot too easily. The behavior of the generic read can be customized by defining `do_reclaim`, `do_normalize` and `normalize_types` on your custom type. If we decide to do this, I will write up some documentation for customization.

Internally, I was able to replace a bunch of separate read methods for different types with this one generic method. This works by separating reading from hdf5 into a memory compatible datatype from mapping that data into a native julia type. The first part can be handled generically (`HDF5Scalar` types are not special) and the second part can be handled by the `normalize_types` methods which we have been using with compound types and can be extended. I think this is a win in terms of simplification and reducing the number of lines of code. The only type that requires special handling is now `HDF5Opaque`.

As a side-effect of this hyperslab indexing works generically for everything besides `HDF5Opaque`. Closes #625.
As another side-effect, vlen strings are now handled uniformly via `unsafe_string` so this should hopefully close #627.

The rules for mapping from hdf5 to julia types now uniformly follow the conventions we have been using for reading compound types which shouldn't be too breaking but does cause a few changes in behavior. In particular the data

```
HDF5 "output.h5" {
DATASET "/salut_vlen" {
   DATATYPE  H5T_VLEN { H5T_STRING {
      STRSIZE 1;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_ASCII;
      CTYPE H5T_C_S1;
   }}
   DATA {
   (0): ("H", "i"), ("t", "h", "e", "r", "e")
   }
}
}
```

is now read as `salut_vlenr = [["H", "i"], ["t", "h", "e", "r", "e"]]` instead of `["Hi", "There"]`. However, I think the new behavior is actually a more correct interpretation of the datatype since `STRSIZE=1`. Also now hdf5 strings are read into julia strings even if they are length 1 whereas previously these would have been read in character types. This could be changed if people think that it is important. Other than the `salut_vlenr` example all other tests pass without modification.


kleinhenz:
@tknopp, @tamasgal would this work for your usecases?

tamasgal:
Looks good to me!

musm:
> As a side-effect of this hyperslab indexing works generically for everything besides HDF5Opaque. Closes #625.
As another side-effect, vlen strings are now handled uniformly via unsafe_string so this should hopefully close #627.

👍 

> is now read as salut_vlenr = [["H", "i"], ["t", "h", "e", "r", "e"]] instead of ["Hi", "There"]. However, I think the new behavior is actually a more correct interpretation of the datatype since STRSIZE=1. Also now hdf5 strings are read into julia strings even if they are length 1 whereas previously these would have been read in character types. This could be changed if people think that it is important. 

I don't have a strong opinion whether it should a string or character. Perhaps there is some efficiency lost by it not being a character, when `STRSIZE=1`. But wouldn't that mean we would need to special case `STRSIZE=1`, which seems cumbersome/inelegant.

src/HDF5.jl
```diff
@@ -1342,7 +1347,7 @@ function read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
   elseif isempty(I)
     sz, _ = get_dims(dspace)
   else
-    sz = map(length, filter(i -> !isa(i, Int), indices))
```
musm: So this doesn't work on 1.3?

src/HDF5.jl
```diff
@@ -1952,6 +1820,15 @@ function hdf5_to_julia_eltype(objtype)
     return T
 end
 
+function get_jl_type(objtype)
```
musm, line 1823: Not sure I understand the point of this method.

The `hdf5_to_julia_eltype` method 
already has defined `class_id == H5T_OPAQUE` 
https://github.com/JuliaIO/HDF5.jl/blob/ece0dcd35e8578231a80e8315d76754fce206945/src/HDF5.jl#L1808

So it seems that without this method things should just work already.

musm, line 1823: maybe add a comment that this is temporarily special cased for `H5T_OPAQUE`

src/HDF5.jl
```diff
@@ -1269,66 +1284,121 @@ function read(parent::Union{HDF5File, HDF5Group}, name::String; pv...)
...
+  dtype = datatype(dset)
+  T = get_jl_type(dtype)
+  read(dset, T, I...)
 end
-# Read array of scalars
-function read(obj::DatasetOrAttribute, ::Type{Array{T}}) where {T<:Union{HDF5Scalar, Complex{<:HDF5Scalar}}}
-    if isnull(obj)
-        return T[]
+
+# could be replaced by filter(i -> !isa(i, Int), args) in julia 1.4
```
musm, line 1311: `filter(i -> !isa(i, Int), args)` works  for me on Julia 1.3 which is the minimum version stated for HDF5.jl

musm:
massive improvement in simplicity and readability.

musm:
feel free to merge whenever

src/HDF5.jl
```diff
@@ -1269,66 +1284,121 @@ function read(parent::Union{HDF5File, HDF5Group}, name::String; pv...)
...
+    close(obj)
+    val
+end
+
 # "Plain" (unformatted) reads. These work only for simple types: scalars, arrays, and strings
 # See also "Reading arrays using getindex" below
 # This infers the Julia type from the HDF5Datatype. Specific file formats should provide their own read(dset).
 const DatasetOrAttribute = Union{HDF5Dataset, HDF5Attribute}
+
 function read(obj::DatasetOrAttribute)
```
musm, line 1299: Can we fix the indentation here also? Should be 4 space indented.

src/HDF5.jl
```diff
@@ -1269,66 +1284,121 @@ function read(parent::Union{HDF5File, HDF5Group}, name::String; pv...)
...
-    if isnull(obj)
-        return T[]
+
+# could be replaced by filter(i -> !isa(i, Int), args) in julia 1.4
+_dropint(x::Int, args...) = _dropint(args...)
+_dropint(x::AbstractRange, args...) = (x, _dropint(args...)...)
+_dropint() = ()
+
+# generic read function
+function read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
```
musm, line 1317: 4-space indent here as well.

musm:
What should we do with the `UTF8Char` and `ASCIIChar` types

src/HDF5.jl
kleinhenz: Yeah this mostly duplicates `hdf5_to_julia_eltype` but changes the string handling part so there are no special cases for the length=1 case. I wrote a new function instead of changing the original because I wanted to only touch the read part of the library. There is probably some consolidation/cleanup possible.

src/HDF5.jl
kleinhenz: I think the filter method for tuples was only added in 1.4. This is what I get in 1.3.

```
filter(i -> !isa(i, Int), (1, 1:10, 1))
ERROR: MethodError: no method matching filter(::var"#3#4", ::Tuple{Int64,UnitRange{Int64},Int64})
Closest candidates are:
  filter(::Any, ::Array{T,N}) where {T, N} at array.jl:2342
  filter(::Any, ::BitArray) at bitarray.jl:1710
  filter(::Any, ::AbstractArray) at array.jl:2354
  ...
```

kleinhenz:
Cool, if you are overall happy with this approach I can clean things up a bit and write some documentation. I left some stuff that can probably be removed like the `hdf5_to_julia_eltype` function which mostly but not quite duplicates `get_jl_type` because I was trying to only touch the read part of the library and some of that is used in the write part. However there is probably at least some cleanup/consolidation possible.

I was actually wrong about the string reading behavior. Fixed length strings with length 1 are already read as strings not characters. The only place where the character types are actually used to change behavior is for reading `H5T_VLEN` datatypes with fixed string element types. In that case, if `STRSIZE=1` then the data is read as an array of strings instead of an array of arrays of characters. This is what is going on in the `salut_vlenr` example. We could replicate the current behavior by adding a method for `normalize_types(x::VariableArray{FixedString})`. This is a fairly weird example though since the natural thing to do would be to just use a variable length string type. I need to check if the character types are important for any of the write methods but otherwise I think they can be removed.

src/HDF5.jl
musm: Ah sorry for some reason I tested it on Array types, which is why I was surprised.

musm:
> Cool, if you are overall happy with this approach I can clean things up a bit and write some documentation. I left some stuff that can probably be removed like the hdf5_to_julia_eltype function which mostly but not quite duplicates get_jl_type because I was trying to only touch the read part of the library and some of that is used in the write part. However there is probably at least some cleanup/consolidation possible.

Right now I think is the appropriate time to make such changes in the upcoming breaking release. It would certainly be good to clean up as much of the old cruft and any ugly parts of the code that we can, so I'm in support.

src/HDF5.jl
musm: Let's  just use https://github.com/JuliaLang/Compat.jl/releases/tag/v3.1.0 instead of this hack?

tknopp:
@kleinhenz: I have switched to the regular compound reading with named tuples: https://github.com/MagneticResonanceImaging/MRIReco.jl/commit/9b1a219c52684e68a47ee7c75057d533575cfab8

works fine although there was some drop in performance.

musm:
@kleinhenz LMK if this is g2g, I'm happy with the changes here.

kleinhenz:
yeah I think this is good to go. There is some more cleanup of internals possible but it is only sort of related to the changes here and I don't want to bloat this pull request too much.

src/HDF5.jl
```diff
@@ -1991,19 +1874,22 @@ function get_mem_compatible_jl_type(objtype)
         finally
             h5t_close(super_type)
         end
+    elseif class_id == H5T_REFERENCE
+        # TODO update to use version 1.12 reference functions/types
```
musm, line 1878: https://github.com/JuliaIO/HDF5.jl/pull/739

PR #655, opened 2020-09-01
Add deprecated bindings for backwards compatibility with ecosystem packages

jmert:
Namely, get the tests of JLD.jl and MAT.jl to pass again.

As noted in the new `deprecated.jl` file, PRs #629 and #632 have broken JLD.jl and MAT.jl for current HDF5.jl master. This PR restores them to working order by adding a bunch of deprecation methods.

Of particular note is re-adding default property values to the `d_create` method — JLD.jl was already using the method and relying on the fact that it only had to provide 2 out of the 4 property lists. I'm also preparing a PR to MAT.jl which ends up using this method as well to fix an error it has after #631.

PRs #629 and #631 have already been tagged in minor releases v0.13.3 and v0.13.4, respectively, so it'd be nice to get some form of deprecated backwards-compatibility layer in place so that until a major version bump is made, external packages continue to work.

Issue #653, opened 2020-08-28
File organization and renaming?


musm:
Changes I'd like to see, for example reorg into submodules the following:

- Library (H5)
- Attribute (H5A)
- Dataset (H5D)
- Dimension Scale (H5DS)
- Error (H5E)
- File (H5F)
- Group (H5G)
- Identifier (H5I)
- Link (H5L)
- MATLAB (H5ML)
- Object (H5O)
- Property (H5P)
- Reference (H5R)
- HDF5 references
- Dataspace (H5S)
- Datatype (H5T)
- Filters and Compression (H5Z)

So that `H5Fget_access_plist` becomes `H5F.get_name` instead of `h5f_get_name`. I guess this could be done without a breaking change because these are internal implementation details in the package. 

I also think the exported names from this package are not very readable/julian:

- `g_create`  should just be `create_group` this would also help with things like tab-completion etc.  Or perhaps even better
`H5G.create` so that  `a_create`  would be   `H5A.create` 

These could be a deprecated in a  breaking release.

- Also more fully supporting the AbstractDict interface and replacing `DataFile` supertype with the `AbstractDict` one makes a lot of sense to me.

I'm not sure if such changes would warrant a new package, e.g.,  H5JL where we can implement such breaking changes without having to worry about deprecation policies.  I'm just throwing this out there as a wild idea.  
Perhaps this is not worth it and it would be better to incrementally improve this package as we are doing (I'm also not against this) 

Curious about what others think about this. 

Furthermore:
```
    HDF5Attribute, HDF5File, HDF5Group, HDF5Dataset, HDF5Datatype,
    HDF5Dataspace, HDF5Object, HDF5Properties, HDF5Vlen, HDF5ChunkStorage,
```
should not have the HDF5 prefix.

cc @kleinhenz @jmert

kleinhenz:
I think the first change is probably actually quite breaking since although these low level functions are technically not exported one of the main uses of the library is low-level access to hdf5 library routines so I suspect the bare wrapper functions are used quite a bit in user code. I think if we did this we would need to have a good deprecation strategy.

I definitely agree that the higher level routines could be reworked to be more readable, and moving to an AbstractDict type of interface seems like a good idea. It seems like some renaming of functions could be done relatively easily with deprecations.

[...]

PR #657, opened 2020-09-02
Add deprecations to aid in next major revision transition

jmert:
This is a superset of #655 which fixed compatibility on the release-0.13 branch (cherry-picked that commit) which adds additional deprecations for the property keyword change of #632.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -932,13 +932,9 @@ function g_create(f::Function, parent::Union{HDF5File, HDF5Group}, args...)
     end
 end
 
-# Note: lcpl does not have a default to distinguish it from the no-keyword method of the
-#      following method.
 function d_create(parent::Union{HDF5File, HDF5Group}, path::String, dtype::HDF5Datatype,
-                  dspace::HDF5Dataspace, lcpl::HDF5Properties,
-                  dcpl::HDF5Properties = DEFAULT_PROPERTIES,
-                  dapl::HDF5Properties = DEFAULT_PROPERTIES,
-                  dxpl::HDF5Properties = DEFAULT_PROPERTIES)
+                  dspace::HDF5Dataspace, lcpl::HDF5Properties, dcpl::HDF5Properties,
+                  dapl::HDF5Properties, dxpl::HDF5Properties)
     HDF5Dataset(h5d_create(checkvalid(parent).id, path, dtype.id, dspace.id, lcpl.id,
                 dcpl.id, dapl.id), file(parent), dxpl)
 end
@@ -1007,28 +1003,23 @@ function setindex!(p::HDF5Properties, val, name::Symbol)
     funcset(p, val...)
     return p
 end
-# Create a dataset with properties: obj[path, prop1, set1, ...] = val
+# Create a dataset with properties: obj[path, prop = val, ...] = val
 function setindex!(parent::Union{HDF5File, HDF5Group}, val, path::String; pv...)
-    dcpl = p_create(H5P_DATASET_CREATE)
-    dxpl = p_create(H5P_DATASET_XFER; pv...)
-    dapl = p_create(H5P_DATASET_ACCESS; pv...)
-
-    need_chunks = any(k in chunked_props for k in keys(pv))
+    need_chunks = any(k in keys(chunked_props) for k in keys(pv))
     have_chunks = any(k == :chunk for k in keys(pv))
 
     chunk = need_chunks ? heuristic_chunk(val) : Int[]
-    discard_chunks = need_chunks && isempty(chunk)
-    if need_chunks && !have_chunks && !discard_chunks
-        dcpl[:chunk] = chunk
-    end
 
     # ignore chunked_props (== compression) for empty datasets (issue #246):
-    for (k,v) in pairs(pv)
-      if !(k in chunked_props && discard_chunks)
-        dcpl[k] = v
-      end
+    discard_chunks = need_chunks && isempty(chunk)
+    if discard_chunks
+        pv = pairs(Base.structdiff((; pv...), chunked_props))
+    else
+        if need_chunks && !have_chunks
+            pv = pairs((; chunk = chunk, pv...))
+        end
     end
-    write(parent, path, val, p_create(H5P_LINK_CREATE), dcpl, dapl, dxpl)
+    write(parent, path, val; pv...)
 end
 
 # Check existence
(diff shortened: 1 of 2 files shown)


jmert:
The second commit "Restore defaults in d_create method to support partial specification" is necessary for current MAT.jl and JLD.jl to pass their tests. I've re-added default arguments, but should this be redone as an explicitly deprecated method that requires all arguments be given explicitly in the future?

musm:
> I've re-added default arguments, but should this be redone as an explicitly deprecated method that requires all arguments be given explicitly in the future?

I don't think in practice it will matter, since the 'parent' https://github.com/JuliaIO/HDF5.jl/pull/632/files#diff-02840edcdbd42c5b41e89b3f700b9a28R916 sets the default values.

But I believe it would still be best to just add these to `deprecate`.

YeI believe that would be better, otherwise it won't actually show up as a deprecated method.

jmert:
The reason it's necessary in the short-term is due to MAT and JLD using the `d_create(parent, name, data, plist...)` methods defined [here](https://github.com/JuliaIO/HDF5.jl/blob/a629a3e1e8fbfbb5350259779c59636a9632ca7a/src/HDF5.jl#L1706-L1727), so the parent method you linked isn't what gets invoked by them.

I guess if you think there's a good chance something like a major reorganization happens (e.g. #653), I'd say it's fine to leave as-is to keep the deprecation path simpler (waiting for the more significant migration to start). If instead the current interface is going to be kept for a while, it's probably better to deprecate the `d_create(parent, name, data, plist...)` methods completely and transition users to `d_create(parent, name, data; kws...)`.

musm:
> If instead the current interface is going to be kept for a while, it's probably better to deprecate the d_create(parent, name, data, plist...) methods completely and transition users to d_create(parent, name, data; kws...).

I do think the current interface will be kept for sometime. 

I havent' received much feedback on https://github.com/JuliaIO/HDF5.jl/issues/653 but it might warrant a new package to keep things simpler

src/HDF5.jl
```diff
@@ -913,7 +913,13 @@ function g_create(f::Function, parent::Union{HDF5File, HDF5Group}, args...)
     end
 end
 
-function d_create(parent::Union{HDF5File, HDF5Group}, path::String, dtype::HDF5Datatype, dspace::HDF5Dataspace, lcpl::HDF5Properties, dcpl::HDF5Properties, dapl::HDF5Properties, dxpl::HDF5Properties)
+# N.B. lcpl does not have a default to distinguish it from the no-keyword method of the
```
musm, line 916: What does N.B. stand for?

musm:
Thoughts on https://github.com/JuliaIO/HDF5.jl/pull/657#issuecomment-685894587  ?
Ideally I think we should just get most people to use the new kwargs methods

jmert:
I generally agree — might as well "rip off the bandaid" leading into the next LTS (Julia v1.6) and just get the interface cleaned up. And since I'm only now developing a use-case, I'm more than happy to just develop for the new syntax without any worries of backward compatibility.

(I intend to make some of those preliminary changes here, but it might be a week or so before I get around to it — I need to resist my fun distractions and get back to some other pressing obligations :-P .)

musm:
Sounds good, I'll wait till then 😄

jmert:
@musm I've made an attempt at deprecating the property-list versions of `{a,d}_{create_write}`. (See the last commit.) It was a bit messy — I've essentially duplicated the `@eval` loops from the main file since there's no easy way to turn `HDF5Properties` objects back into key-value pairs. It also required fixing up a `setindex!` implementation to use keywords as well instead of the property-list syntax. This notably requires keeping the default-values version of the low-level `d_create` as part of the implementation, so I've just added a note that those default values can/should be removed when the deprecations are.

I also added another deprecation for a `read` function which was changed in #652 that I discovered by trying to run the MAT.jl tests. Unfortunately, this PR still isn't enough to get MAT.jl working again on master HDF5.jl — there's an ambiguity which I couldn't trivially solve (and was stalling me on just getting at least this much work shared).

src/HDF5.jl
```diff
@@ -931,7 +931,13 @@ function g_create(f::Function, parent::Union{HDF5File, HDF5Group}, args...)
...
```
musm, line 934:
```suggestion
# NOTE: lcpl does not have a default to distinguish it from the no-keyword method of the following method.
```
Maybe we should also not that this is added only for deprecations purposes and can be removed in the next minor version bump?

src/HDF5.jl
```diff
@@ -931,7 +931,13 @@ function g_create(f::Function, parent::Union{HDF5File, HDF5Group}, args...)
...
+#      following method.
```
musm, line 935:
```suggestion
# NOTE: lcpl does not have a default to distinguish it from the no-keyword method of the following method.
```

jmert: Ah, stepping away made me realize there's another way around this — just dispatch through another private `_d_create` that sets the default properties, and that hands off to the true low-level `d_create`.

In looking at this, though, I've noticed `a_create` never actually accepts keywords/properties, even though these higher-level wrapper suggest one could use them. Should that lie be maintained?

jmert:
@kleinhenz The change related to `STRLEN=1` reads in #652 broke a use in MAT.jl. (The first test failure occurs when trying to read the `"MATLAB_fields"` attribute shown below.)

I don't really understand yet what's going on / what it means for those property settings, so could you help figure out if there's a way to make a deprecation path?

```julia
julia> using MAT

julia> cd(joinpath(dirname(pathof(MAT)), "..", "test"))

julia> matread("empty_struct.mat")
ERROR: MethodError: Cannot `convert` an object of type Array{String,1} to an object of type String
Closest candidates are:
  convert(::Type{T}, ::T) where T<:AbstractString at strings/basic.jl:229
  convert(::Type{T}, ::AbstractString) where T<:AbstractString at strings/basic.jl:230
  convert(::Type{T}, ::T) where T at essentials.jl:171
Stacktrace:
 [1] setindex!(::Dict{String,Any}, ::Array{Any,1}, ::Array{String,1}) at ./dict.jl:372
 [2] Dict{String,Any}(::Base.Generator{Array{Array{String,1},1},MAT.MAT_HDF5.var"#1#2"}) at ./dict.jl:103
 [3] m_read(::HDF5Dataset) at /home/justin/.julia/dev/MAT/src/MAT_HDF5.jl:146
 [4] read(::MAT.MAT_HDF5.MatlabHDF5File, ::String) at /home/justin/.julia/dev/MAT/src/MAT_HDF5.jl:264
 [5] read(::MAT.MAT_HDF5.MatlabHDF5File) at /home/justin/.julia/dev/HDF5/src/datafile.jl:45
 [6] matread(::String) at /home/justin/.julia/dev/MAT/src/MAT.jl:133
 [7] top-level scope at REPL[54]:1

shell> h5dump -Hp empty_struct.mat
HDF5 "empty_struct.mat" {
GROUP "/" {
   DATASET "a" {
      DATATYPE  H5T_STD_U64LE
      STORAGE_LAYOUT {
         COMPACT
         SIZE 16
      }
      FILTERS {
         NONE
      }
      FILLVALUE {
         FILL_TIME H5D_FILL_TIME_IFSET
         VALUE  H5D_FILL_VALUE_DEFAULT
      }
      ALLOCATION_TIME {
         H5D_ALLOC_TIME_EARLY
      }
      ATTRIBUTE "MATLAB_class" {
         DATATYPE  H5T_STRING {
            STRSIZE 6;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
      }
      ATTRIBUTE "MATLAB_empty" {
         DATATYPE  H5T_STD_U8LE
         DATASPACE  SCALAR
      }
      ATTRIBUTE "MATLAB_fields" {
         DATATYPE  H5T_VLEN { H5T_STRING {
            STRSIZE 1;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }}
      }
   }
}
}
```

kleinhenz:
hmm I guess this is why that behavior was there. It still seems quite weird to me to use 
```
DATATYPE  H5T_VLEN { H5T_STRING {
            STRSIZE 1;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }}
```

as a string type rather than just a variable length string. Does anyone know if this is actually part of the spec? I can't seem to find a reference online and unfortunately I don't have matlab installed on my machine anymore. I could change the normalize_types method to handle this case but I don't really love the idea of special casing this IMHO weird usage in this general package. 

I don't really see a good deprecation path for this since the relevant methods are mostly called internally. My instinct would be to fix this on the MAT side and make a simultaneous release. I can make a pull request for MAT if people want.

musm:
I tend to agree, ideally it would be best if we could fix this on the MAT.jl side. 

I have matlab installed, let me know if you want me to check the output of something specific.

jmert:
> I don't really see a good deprecation path for this since the relevant methods are mostly called internally. My instinct would be to fix this on the MAT side and make a simultaneous release. I can make a pull request for MAT if people want.

A fix in MAT is fine — it's clearly a quirk in how Matlab chooses to store that particular attribute (the "MATLAB_class" attribute is being saved as a string...). Your help is appreciated with this since I'm mostly teaching myself HDF5 data semantics via trying to load some specific data into a project while digging through this package, and I haven't fully internalized what the different representations really mean.

musm:
Are the integrations supposed to fail? My assumption was that the CI is pulling in this PR to do the testing, but it looks like perhaps that's not the case.

musm:
I suppose this PR should now be merged? It looks like we can't do anything about the MAT.jl failure other than fix it in that library directly

jmert:
> Are the integrations supposed to fail? My assumption was that the CI is pulling in this PR to do the testing, but it looks like perhaps that's not the case.

It is pulling in this PR — the MAT.jl integration gets further on this branch than master does, but Julia nightly identifies a method ambiguity that v1.5 doesn't see, and even on v1.5 it'll fail when it gets to reading of `"MATLAB_fields"` as mentioned earlier.

I haven't looked into the JLD.jl failure.

> I suppose this PR should now be merged? It looks like we can't do anything about the MAT.jl failure other than fix it in that library directly.

Yes, MAT.jl is going to need at least one fix itself, so it alone shouldn't hold up merging at least these deprecations. There's maybe something to do on this end about JLD's failure and/or the method ambiguity problem, though, and that could happen either here or another PR.

musm:
Yeah, let's go ahead and merge this and we can work towards fixing the rest of the deprecations instead of holding this up?

musm (approved):

kleinhenz:
yeah it looks like I broke a few things with #652. I agree that we should merge this and then work on fixing the errors.

kleinhenz (approved):

musm:
Many thanks @jmert

PR #660, opened 2020-09-08
Add convenient get_access_properties() functions

jmert:
This adds a method similar to `get_create_properties` which gets the access property list for HDF5 datasets and files — as far as I've found trying to look through the online documentation, those are the only two for which `h5?_get_access_plist` exists, but I might have missed one.

Note that this does add a new export for `get_access_properties`, which can be undone if desired.

This was motivated by trying to figure out how to query the alignment access property which also led to #659 — I just broke up the changes into two smaller PRs.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -22,7 +22,8 @@ export
     a_create, a_delete, a_open, a_read, a_write, attrs,
     d_create, d_create_external, d_open, d_read, d_write,
     dataspace, datatype, exists, file, filename,
-    g_create, g_open, get_chunk, get_create_properties, get_datasets,
+    g_create, g_open, get_access_properties, get_create_properties,
+    get_chunk, get_datasets,
     h5open, h5read, h5rewrite, h5writeattr, h5readattr, h5write,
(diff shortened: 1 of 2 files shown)


musm:
lgtm, I do think it makes sense to export  this if `get_create_properties` is also exported.

jmert:
What do you think of backporting both this and #659 changes to the 0.13 branch? (I guess the question is, should bug fixes and minor features still be added to the 0.13 series while master is working towards a bigger breaking change to a new 0.14 breaking release?)

musm:
We can do that, #659 is more a bug fix. I'm not sure this one should be backported since it exports, but then again, they go hand in hand, so I don't have a problem with it. 


I do think we are close to tagging 0.14 in any case, we are probably only waiting on the compatibility PR to smooth the transition. But likely we will tag a release in a week or so.

jmert:
OK, that's a pretty short timescale, I'd say it can be ignored unless/until someone requests it on 0.13 specifically. #659 doesn't actually interact with any exported code paths, so while it is a bug fix, you do have to dig through internals to try to make use of it.

Issue #661, opened 2020-09-10
Testing impact to downstream packages


jmert:
In recent PRs (#655 and #657), I have tried to smooth over changes made by other PRs that impacted external ecosystem packages, namely by running MAT.jl and JLD.jl's tests and investigating failures. Some of them have been intended to be breaking (such as #632) and will clearly require accompanying changes in downstream packages, but others (like #631) are well-motivated fixes that nonetheless require follow-ups when used "in the wild" (e.g. JuliaIO/MAT.jl#142).

It might be useful to add a testing phase — chain-invoking MAT and JLD's own tests as a testset? a separate CI job? other? — which can protect against unintended breakages and/or clarify what intended changes will require deprecation methods or corresponding external work.

musm:
That's a good idea. I wonder if anyone else has done anything similar?

musm:
test-set dependencies and call `Pkg.test` on them?

PR #662, opened 2020-09-10
MAT.jl and JLD.jl downstream integration testing (#662)

musm:
close https://github.com/JuliaIO/HDF5.jl/issues/661


musm:
Which platforms should we run integration tests and julia version?
Currently I assume
- Julia 1.5 and nightly
- Ubunutu and x64 only

musm:
Do we need this for JLD also? I believe that package is in maintenance mode

jmert:
Yeah, this is the kind of thing I had in mind.

> Which platforms should we run integration tests and julia version?
I like to test against earliest supported [*], latest stable, and nightly Julia — so that'd be 1.3, 1.5, and nightly, as I understand it — but I'm willing to hear arguments for/against that. For this type of test, I also assume Linux-only will be good enough since this should be just testing the API-level integration between the packages. Any OS-specific behavior tests should happen within the regular test sets of the responsible package.

> Do we need this for JLD also? I believe that package is in maintenance mode
Not sure, I guess. I've gotten involved here mainly because I want to use HDF5 for a specific use case, and then I want to allow (through `@requires` hooks) importing of a sparse matrix by name with any of HDF5, MAT, JLD, and JLD2 — i.e. try to be as convenient as possible for the user. If JLD's not going to get upgraded for the v0.14 world, I guess I just drop support for that option since I probably am going to require and make use of the HDFv0.14 in my project (and JLD2 is still a very easy option for Julia users).

musm:
> I like to test against earliest supported [*], latest stable, and nightly Julia — so that'd be 1.3, 1.5, and nightly, as I understand it — but I'm willing to hear arguments for/against that. For this type of test, I also assume Linux-only will be good enough since this should be just testing the API-level integration between the packages. Any OS-specific behavior tests should happen within the regular test sets of the responsible package.

Me too, but I think for integration testing it might be overkill on CI to test on all three versions? I only opted for 1.5 and nightly, but I don't mind either going for 1.3 (earliest) and nightly (latest) or all three. I don't have a strong preference TBH. 


> Not sure, I guess. I've gotten involved here mainly because I want to use HDF5 for a specific use case, and then I want to allow (through @requires hooks) importing of a sparse matrix by name with any of HDF5, MAT, JLD, and JLD2 — i.e. try to be as convenient as possible for the user. If JLD's not going to get upgraded for the v0.14 world, I guess I just drop support for that option since I probably am going to require and make use of the HDFv0.14 in my project (and JLD2 is still a very easy option for Julia users).

I'm happy to add JLD, but personally I don't use the package. Do tests pass with JLD on master HDF5? I guess I could test it out.

musm:
I ended up trimming it just to 1.5, it's a lot of visual noise in the "checks" box to keep track of.

musm:
Feel free to open further PR's on this if you think we should change it. For now it's certainly better than nothing. I think for now testing on one Julia version is ok, it will catch any glaring integration failures, which is what I think we want this for anyways.

PR #663, opened 2020-09-11
Do not leak a dlopen library pointer into the precompiled image

jmert:
The library handle is a pointer, so it should not be cached during precompilation.

With some larger refactoring, the function `read_const` could be hidden completely, but at least now trying to use the function won't segfault Julia, and instead you just get an error about a NULL library handle.

---

This led down a rabbit hole, and as far as I can tell, all of the uses of `read_const` are technically not precompilation-safe either — in principle, the constants being retrieved are actually runtime-generated answers:
```julia
julia> using Libdl

julia> libhdf5 = dlopen("libhdf5");

julia> unsafe_load(convert(Ptr{Int64}, dlsym(libhdf5, :H5P_CLS_DATASET_CREATE_ID_g)))
-1

julia> # Initialize the HDF5 library
       ccall(dlsym(libhdf5, :H5open), Int64, ())
0

julia> unsafe_load(convert(Ptr{Int64}, dlsym(libhdf5, :H5P_CLS_DATASET_CREATE_ID_g)))
720575940379279369
```
Emprically, though, they must be generated in some mostly-stable fashion for this to have gone unnoticed (or maybe just unmentioned in the code, but already known/discussed in an old issue?).

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -66,8 +66,10 @@ primitive type Hmpih64 <: Hmpih 64 end # OpenMPI C: pointers (mostly 64 bit)
 
 # Function to extract exported library constants
 # Kudos to the library developers for making these available this way!
-let libhdf5handle = Libdl.dlopen(libhdf5)
-    global read_const(sym::Symbol) = unsafe_load(convert(Ptr{Hid}, Libdl.dlsym(libhdf5handle, sym)))
+function read_const(sym::Symbol)
+    handle = Libdl.dlopen(libhdf5)
+    ccall((:H5open, libhdf5), Herr, ()) < 0 && error("error initializing the HDF5 library")
+    return unsafe_load(convert(Ptr{Hid}, Libdl.dlsym(handle, sym)))
 end
 
 # iteration order constants
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -66,8 +66,9 @@ primitive type Hmpih64 <: Hmpih 64 end # OpenMPI C: pointers (mostly 64 bit)
 
 # Function to extract exported library constants
 # Kudos to the library developers for making these available this way!
-const libhdf5handle = Libdl.dlopen(libhdf5)
-read_const(sym::Symbol) = unsafe_load(convert(Ptr{Hid}, Libdl.dlsym(libhdf5handle, sym)))
+let libhdf5handle = Libdl.dlopen(libhdf5)
+    global read_const(sym::Symbol) = unsafe_load(convert(Ptr{Hid}, Libdl.dlsym(libhdf5handle, sym)))
+end
 
 # iteration order constants
 const H5_ITER_UNKNOWN = -1


musm:
Makes sense, how'd you manage to segfault Julia? Curious what the MWE was that triggered that.

yuyichao (changes requested):
The correct fix is to initialize this in `__init__`

musm:
Do you mean all the constants generated by `read_const` need to be initialized in `__init__`?

yuyichao:
I mean if this was meant to be callable after package loading then it should be fixed properly by initializing the handle in `__init__`. If it was never meant to be called outside then this change isn't needed.

musm:
If I understand correctly, all these constants, e.g.
```
const H5P_OBJECT_CREATE    = read_const(:H5P_CLS_OBJECT_CREATE_ID_g)
```
`H5P_OBJECT_CREATE ` should instead be a constant ref,  that is initialized in `__init__` right?

yuyichao:
No

yuyichao:
By this I mean this function, i.e. `read_const`

yuyichao:
and here I'm assuming these return values are actually constant and doesn't change from run to run. If that's not the case then a much more involved fix would be necessary and this PR is still wrong. From a quick look I don't think any change to the use of `read_const` is needed.

The only question is: is `read_const` supposed to be used from outside the package? If not, it should just be made clear and no code change is necessary. If yes, then it should be fixed properly by making it works. It is not a problem to have function defined in a package that cannot be called externally.

------

On a related note, `cglobal` should probably work.

musm:
> The only question is: is read_const supposed to be used from outside the package? If not, it should just be made clear and no code change is necessary. If yes, then it should be fixed properly by making it works. It is not a problem to have function defined in a package that cannot be called externally.

Yeah, read_const is not meant to be used outside the package. It's only used as a helper to pull out constants from the HDF5 public api.

jmert:
> Makes sense, how'd you manage to segfault Julia? Curious what the MWE was that triggered that.

Trying to `Libdl.dlsym` with the stale `libhdf5handle`.

> and here I'm assuming these return values are actually constant and doesn't change from run to run. If that's not the case then a much more involved fix would be necessary and this PR is still wrong. From a quick look I don't think any change to the use of read_const is needed.
>
> > The only question is: is read_const supposed to be used from outside the package? If not, it should just be made clear and no code change is necessary. If yes, then it should be fixed properly by making it works. It is not a problem to have function defined in a package that cannot be called externally.
> 
> Yeah, read_const is not meant to be used outside the package. It's only used as a helper to pull out constants from the HDF5 public api.

I dug through the HDF5 source a little bit, and in principle the "constants" being loaded can change since they're initialized by [implicit calls to] `H5open()` on first use of the library, but I didn't dig far enough to see if those variables are in practice fixed (e.g. for a given library version). (A small amount of searching docs didn't reveal any concrete guarantees either, but I could have easily just missed such statements.) Empirically they must be effectively fixed since this scheme works in the real world.

Despite `read_const()` and `libhdf5handle` being purely internal, I wanted to just clean this up since segfaulting is never a nice experience. I was attempting to `dlsym` something from the library directly as part of debugging other issues, and I made the mistake of trying to use this stale `libhdf5handle`, when what I wanted to use was `HDF5.HDF5_jll.libhdf5_handle`. (I needed to be sure I was accessing the jll-provided library [at v1.10] and not my OS's installed copy [at v1.12], so I was aiming to reuse the existing handle and just skipped over the minor spelling difference.)

This change is enough to avoid the segfault, and the larger issue of whether the "constants" should all be fixed up to be `__init__`-time `Ref`s can be taken up in a separate PR.

yuyichao:
> Despite read_const() and libhdf5handle being purely internal, I wanted to just clean this up since segfaulting is never a nice experience. I was attempting to dlsym something from the library directly as part of debugging other issues, and I made the mistake of trying to use this stale libhdf5handle, when what I wanted to use was HDF5.HDF5_jll.libhdf5_handle. (I needed to be sure I was accessing the jll-provided library [at v1.10] and not my OS's installed copy [at v1.12], so I was aiming to reuse the existing handle and just skipped over the minor spelling difference.)

No there's nothing wrong with the crash you see here. You can't just expect to call random functions in a package and expect nothing wrong to happen. If you think the names are too discoverable, renaming and document as internal is certainly fine. The issue is that this is not a clean up and this doesn't do what the title says it does. The dlopen library pointer is still in the precompiled image and the new code is more complicated.

> the larger issue of whether the "constants" should all be fixed up to be __init__-time Refs can be taken up in a separate PR.

And that's exactly why I said this PR doesn't fix any real issue. The potentially real issues are the incorrect value at runtime and internal functions looks too much like external ones.

jmert:
> No there's nothing wrong with the crash you see here. You can't just expect to call random functions in a package and expect nothing wrong to happen. If you think the names are too discoverable, renaming and document as internal is certainly fine. The issue is that this is not a clean up and this doesn't do what the title says it does. The dlopen library pointer is still in the precompiled image and the new code is more complicated.

OK, I've updated the function to always load and initialize[*] the library before reading the constants. That gets rid of any stored pointer _in this function_. (I acknowledge that there's still stored runtime values in all the constants being loaded.)

[*] You can actually break the existing package by just commenting out [this line](https://github.com/JuliaIO/HDF5.jl/blob/9640a4198414c85f9ee4f3cb5008cb8bdeb4b0ed/src/HDF5.jl#L49) which initializes the library early; without the initialization, the reads give invalid property handles.

> > the larger issue of whether the "constants" should all be fixed up to be `__init__`-time `Refs` can be taken up in a separate PR.
> 
> And that's exactly why I said this PR doesn't fix any real issue. The potentially real issues are the incorrect value at runtime and internal functions looks too much like external ones.

I still think that mitigating one way someone can break something is an improvement, but I won't argue that this is a band-aid on a much more serious problem — it's just one I don't want to spend the time to work on at the moment.

yuyichao:
No, this is still just making things slower without any benefit.

> [*] You can actually break the existing package by just commenting out this line which initializes the library early; without the initialization, the reads give invalid property handles.

It does seems that the initialization is stable so that's not really a problem. It needs to be documented though so that no one breaks it or if anyone does it can easily be tracked down.

> I still think that mitigating one way someone can break something is an improvement, but I won't argue that this is a band-aid on a much more serious problem — it's just one I don't want to spend the time to work on at the moment.

No, not really,

1. It does not prevent anyone from breaking anything. The initialization is still not guaranteed (and doesn't have to be) which is the only thing that matters for this package and is also an internal matter that should be fixed by documentation.

2. What you are doing is to allow ppl to call an internal function without satisfying the assumption made. Removing such assumptions on functions no one should use by adding overhead to the intended user is not an improvement.

-----

And again, I'm not saying nothing should change. But you must either fix a real problem or make a real cleanup. The options for real problems are:

1. decide if HDF5.jl should rely on the ABI of the version of the library at build time

2. make it more clear that internal function is internal

and the requirement for clean up is to actually reduce code complexity and to not introduce more overhead than before. If to fix an real issue more overhead is necessary that's acceptable to a certain degree. However, none of the real issues requires initializing the library handle after the package compilation time so what this PR is doing is not needed for either.

One example of doing clean up would be to use `cglobal` as I mentioned above.

yuyichao:
Case in point, CUDA.jl has a [`trap` function](https://github.com/JuliaGPU/CUDA.jl/blob/463a41295bfede5125c584e6be9c51a4b9074e12/src/device/llvm.jl#L3). It's basically defined to crash and is intended to be used to detect inconsistency in the compiler etc. It's not an improvement to "fix" it so that ppl decided to randomly call every function they see get no crashes. It's an extreme example of course but still, using internal functions (or even external ones FWIW) in ways that they aren't intended to be used as isn't something to "improve on", at least not just because there's a crash. If you have a good reason to call `read_const` from another package that's whole different story and that's exactly what I asked first but it seems like the answer was no.

jmert:
> One example of doing clean up would be to use cglobal as I mentioned above.

The trivial change doesn't work because `sym` isn't a literal in the `cglobal` call:
```diff
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -66,8 +66,7 @@ primitive type Hmpih64 <: Hmpih 64 end # OpenMPI C: pointers (mostly 64 bit)
 
 # Function to extract exported library constants
 # Kudos to the library developers for making these available this way!
-const libhdf5handle = Libdl.dlopen(libhdf5)
-read_const(sym::Symbol) = unsafe_load(convert(Ptr{Hid}, Libdl.dlsym(libhdf5handle, sym)))
+read_const(sym::Symbol) = unsafe_load(cglobal((sym, libhdf5), Hid))
```
```julia
julia> using HDF5
ERROR: LoadError: TypeError: in cglobal: first argument not a pointer or valid constant expression, expected Ptr, got a value of type Tuple{Symbol,String}
Stacktrace:
...
```
so, does that leave only a macro to do the job? Would you prefer this?
```diff
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -66,8 +66,9 @@ primitive type Hmpih64 <: Hmpih 64 end # OpenMPI C: pointers (mostly 64 bit)
 
 # Function to extract exported library constants
 # Kudos to the library developers for making these available this way!
-const libhdf5handle = Libdl.dlopen(libhdf5)
-read_const(sym::Symbol) = unsafe_load(convert(Ptr{Hid}, Libdl.dlsym(libhdf5handle, sym)))
+macro read_const(sym)
+    return :(unsafe_load(cglobal(($sym, libhdf5), Hid)))
+end
```
(and corresponding changes to all `read_const` -> `@read_const`).

> ... It's not an improvement to "fix" it so that ppl decided to randomly call every function they see get no crashes.

Of course not — the purpose of functions like that and `abort()` are to crash, so they're working as intended.

---

I don't have any real vested in this PR — I thought it was an opportunity for a trivial mitigation, but I'm fine closing this as well, leaving any modifications to a PR with proper, systemic changes.

yuyichao:
> so, does that leave only a macro to do the job? Would you prefer this?

Yes that should be OK, but isn't cheaper so probably not.

> so they're working as intended.

So is `read_const`. It's for reading value at compile time and it works as intended. If you want an example for a function that doesn't aways crash, there's `unsafe_load`. If you call it on the wrong pointer of even at the wrong time it'll also crash. Adding a guard to that is possible but is not the right thing to do.

jmert:
> > Would you prefer this?
>
> Yes that should be OK, but isn't cheaper so probably not.

OK

musm:
@yuyichao  Thanks for your input. 

As I understand, using `@read_const` is not better / worse that what we are currently doing since the are still stored in the precompiled image.

Regarding the macro:

Why do you say that using the `@read_const` macro is not cheaper?  

```
julia> @time H5P_OBJECT_CREATE = read_const(:H5P_CLS_OBJECT_CREATE_ID_g)
  0.000020 seconds (3 allocations: 48 bytes)
648518346341351425

julia> @time H5P_OBJECT_CREATE = @read_const(:H5P_CLS_OBJECT_CREATE_ID_g)
  0.000001 seconds (1 allocation: 16 bytes)
648518346341351425
```

In any case, I'm not sure why the timings would matter since these are all compile time values.
In other words, even if it is slower, why would this matter since they are precompile constants.

In addition, it does seem a lot cleaner to use the macro that expands the cglobal.

musm:
> and here I'm assuming these return values are actually constant and doesn't change from run to run. If that's not the case then a much more involved fix would be necessary and this PR is still wrong. From a quick look I don't think any change to the use of read_const is needed.

Right, I think technically    the correct way is to initialize them in `__init__`, but I'm not sure it really matters for these constants, since they shouldn't ever change.

yuyichao:
> Why do you say that using the `@read_const` macro is not cheaper?

Because you are not measuring the compilation time.

> In any case, I'm not sure why the timings would matter since these are all compile time values.

Because small places like this is exactly what get accumulated to bad precompilation and startup time.

> In addition, it does seem a lot cleaner to use the macro that expands the cglobal.

Compared to what this PR was doing, yes for sure. Compared to what it was, not really. A function is usually preferred over a macro when both can achieve the same goal.

musm:
Makes sense.  

Ok so, here are the timing differences using the macro `@read_const`

master
```
julia> @time using HDF5
  3.570521 seconds (1.65 M allocations: 99.453 MiB, 0.68% gc time)....

julia> @time using HDF5
  1.161993 seconds (1.56 M allocations: 94.079 MiB, 2.05% gc time)
```


Using `@read_const`

```
julia> @time using HDF5
  3.914678 seconds (1.65 M allocations: 99.426 MiB, 0.68% gc time)

....
julia> @time using HDF5
  1.215145 seconds (1.56 M allocations: 94.047 MiB, 2.04% gc time)
```

So we are looking at (very roughly, due to noise)  ~8%  slower to precompile and  ~3% to load

musm:
Arguably, however, the `@read_const` version is cleaner and the 8% precompilation hit, isn't that significant? Alternatively, we could just manually expand out the macro directly.

yuyichao:
> the `@read_const` version is cleaner

As I said, that's what I don't agree with. Using a macro when a function can do is almost always a minus point.

PR #664, opened 2020-09-15
Explicitly generate ccall wrapper functions

jmert:
This is an alternative to #646.

This PR replaces the three separate `for` loops that have been defining a whole swath of `ccall` wrapper functions with a file of explicitly-generated functions. The new `src/api.jl` file is generated from a list of API definitions contained in `gen/api_defs.jl` and processed by a macro defined in `gen/bind_generator.jl`. Those two files alone could be used together to do the function generation on-demand, but a third file — `gen/gen_wrappers.jl` — exists to expand all the definitions explicitly and save the results to a source file: this is how `src/api.jl` has been created.

The first commit does the direct translation of the `for` loops into the described generator infrastructure.

The second commit is a bug fix — two wrapper functions misnamed the HDF5 library functions.

The third commit renames a few of the wrapper functions to make them more consistent with the rest of the package's naming choices. Deprecation bindings have been added to aid any external user in this transition.


jmert:
I just added a fourth commit which improves the error handling for various return types a bit — in particular, there are two `typedef`-ed unsigned integers in the HDF5 headers (here, `Haddr` and `Hsize`) that never compare as `< 0`. This changes the generator to special-case these two return types, and instead look for the value of `-1 % Haddr` and `-1 % Hsize`, respectively, as the error condition.

It ends up being that the only impacted function was `h5d_get_offset`, which makes the error check in `readmmap` redundant (though it changes the error message).

There are still two instances of ineffective error checks — `h5t_get_size` and `h5t_get_member_offset` — that I left alone because `h5t_get_member_offset` has no error case indicated by its return value. In contrast, `h5t_get_size` with a return of `0` indicates an error, so I've left that unchanged for the moment.

musm (approved):
Looks great, just a couple of minor comments.

musm:
This also cuts precomp time by 20% percent  (down to 2.8 from 3.6 s)

gen/bind_generator.jl
```diff
@@ -0,0 +1,127 @@
...
+with "H5DO" or "H5TB" then the latter library is used.
+"""
+macro bind(sig::Expr, err::Union{String,Expr})
+    sig.head === :(::) || error("return type required on function signature")
+
+    # Pull apart return-type and rest of function declaration
+    rettype = sig.args[2]::Union{Symbol,Expr}
+    funcsig = sig.args[1]
+    isexpr(funcsig, :call) || error("expected function-like expression, found `", funcsig, "`")
+    funcsig = funcsig::Expr
```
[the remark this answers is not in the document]
jmert: If the expression is really malformed (e.g. a Symbol), it'll type-assert error prior to the helpful `error()` call, but yeah, it's probably a bit too far on the defensive side.

gen/bind_generator.jl
```diff
@@ -0,0 +1,127 @@
...
+
+    # Extract function name and argument list
+    jlfuncname = funcsig.args[1]::Symbol
+    funcargs = funcsig.args[2:end]
+
+    # Pull apart argument names and types
+    args = Vector{Symbol}()
+    argt = Vector{Any}()
```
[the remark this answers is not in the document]
jmert: Probably. I remember I had `Vector{Symbol}` on both at the start and that wasn't compatible with `Ptr` types since those are expressions, so I just widened to `Any`. I could definitely try narrowing back to a `Union`.

gen/bind_generator.jl
```diff
@@ -0,0 +1,127 @@
...
+        # Returns a value
+        returnexpr = :(return $statsym)
+    end
+
+    # Then assemble the pieces. Doing it through explicit Expr() objects
+    # avoids inserting the line number nodes for the macro --- the call site
+    # is instead explicitly injected into the function body via __source__.
+    jlfuncsig = Expr(:call, jlfuncname, args...)
+    jlfuncbody = Expr(:block,
+                      __source__,
```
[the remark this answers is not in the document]
jmert: Yeah, the idea was to allow an on-demand version to still give some useful source information, even though the pre-generated version just strips that back out. (As I said in the other PR's comment, this started as an exploration of working with macros a bit more and teaching myself some of the constructs, so I could remove this if desired.)

gen/bind_generator.jl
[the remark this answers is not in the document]
musm: oh right there's the error call on line 64

gen/bind_generator.jl
```diff
@@ -0,0 +1,147 @@
...
+        # turn e.g. h5f_close into H5Fclose
+        prefix, rest = split(string(jlfuncname), "_", limit = 2)
+        cfuncname = Symbol(uppercase(prefix), rest)
+    end
+
+    # Determine the underlying C library to call
+    lib = startswith(string(cfuncname), r"H5(DO|TB)") ? :libhdf5_hl : :libhdf5
+
+    # Now start building up the full expression:
+    statsym = Symbol("#status#") # not using gensym() to have stable naming
```
musm, line 106: Does this really need to be `var"#status"`, why not just leave it as, say, `r` or `ret` ?

jmert: The idea was to make a naming clash with the function arguments unlikely, since we're not actually relying on macro hygiene to ensure the internal and externally-provided symbols are unique.

musm: Hmm, I see. It would be pretty unlikely someone do name an argument `ret`, but I suppose it "could" happen.  I guess I'm just a little apprehensive since it isn't very idiomatic.

jmert: We could change it to something more normal yet still exotic — `__status_` or something.

musm: I'm on the fence, we could just keep as is  or name it `status` where it would be pretty straightforward  to catch errors if people commit the wrong symbol names.

gen/bind_generator.jl
```diff
@@ -0,0 +1,147 @@
...
+2. If `ReturnType === :Haddr` or `ReturnType === :Hsize`, an error is raised when the
+   return value is equivalent to `-1 % Haddr` and `-1 % Hsize`, respectively.
+
+3. For all other return types, it is assumed a negative value indicates error.
+
+It is assumed that the HDF library names are given in global constants named `libhdf5`
+and `libhdf5_hl`. The former is used for all `ccall`s, except if the C library name begins
+with "H5DO" or "H5TB" then the latter library is used.
+"""
+macro bind(sig::Expr, err::Union{String,Expr})
```
musm, line 68: Come to think of it, wouldn't have been  better if the `@bind` generator converted the `C` function name to the  Julia name as opposed to the opposite process ?  This is in some sense similar to the `@ccall` macro.

musm: I guess there's the issue of the exception list, which complicates that a bit. I assume that's probably why you took this approach instead.

jmert: Yeah, I'd originally done it the other way around, but the MPIO wrappers which all share the same C call but have unique Julia names made that complicated since the Dict key wasn't unique (without having to do something more complicated like using the full type signature as the key).

PR #670, opened 2020-09-17
Allow no-error case in binding generator and add error checking of Ptr returns

jmert:
This is a small follow-up to #664.

The first change is to add a way to not emit an error check into the wrapper function since I found two current cases where the functions are documented to not have return values that indicate error. (In both cases, they return unsigned data types, so the current `< 0` check is eventually elided by LLVM's dead code elimination since the branch can never be true).
* `H5Pget_driver_info` will return a pointer to a driver-specific data structure which may include `NULL` if the driver has no such data structure.
* `H5Tget_member_offset` will not error, and the docs say that error states are checked instead through first checking if `H5Tget_member_class` errors or not.

I discovered these in trying to make sure that all stated return types are handled, and it was clear that for `Ptr` and `Csize_t` the `< 0` check didn't make sense.

The second change is currently unused (since `H5Pget_driver_info` is marked as not erroring), but I added in an error handler for `Ptr` return values that considers `C_NULL` return to be an error.


src/api.jl
```diff
@@ -485,7 +485,6 @@ end
 
 function h5p_get_driver_info(plist_id)
     var"#status#" = ccall((:H5Pget_driver_info, libhdf5), Ptr{Cvoid}, (Hid,), plist_id)
-    var"#status#" < 0 && error("Error getting driver info")
```
musm, line 488: Is the CNull error handling here not being picked up?

jmert: No, that's on purpose — [`H5Pget_driver_info`](https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pget_driver_info.htm) doesn't error because it's valid for the driver to not have extra info. That's the reason I had to allow for a no-error case in the generator (setting `err = nothing` by default in the macro definition).

musm: Oh I see err has to be set and the return type has to be Ptr for that case to be generated

PR #675, opened 2020-09-18
More robust tests and cleanup

musm:
Sometimes the downloads fail, so I decided to commit them to the repo so the tests are more robust.


test/plain.jl
```diff
@@ -399,14 +399,15 @@ if !isempty(HDF5.libhdf5_hl)
 end
 
 # Test that switching time tracking off results in identical files
-h5open("tt1.h5", "w") do f
+fn1 = tempname(); fn2 = tempname()
```
musm, line 402: I guess this is not a valid usage of the track times function? cc @jmert

jmert: I've been noticing that that test is flaky, but I haven't figured out why yet. Mostly I see it in the CI results, but I did see it once locally as well (though of course not reproducibly). This might be a better starting point, though, since I/we could augment the test locally and not delete the files if the crc's don't match — then comparing the files might show why it's sometimes failing and sometimes not.

musm: Yeah that sounds like a good way to debug this. I'm going to go ahead with this change in this case for now.

PR #676, opened 2020-09-18
fix dtype printing for closed datatypes

### LINKED ISSUE #671: Printing of closed datatype throws error
For other wrapper data types like `HDF5File`, `HDF5Dataset`, etc, printing the types after closing indicates that the object is no longer valid:
```julia
julia> hfile = h5open("test.h5", "r")
       dset = hfile["v1"]
       close(hfile)
       show(dset)
HFD5 dataset (invalid)
```

This is not true for the `HDF5Datatype`, though:

```julia
julia> dtype = HDF5.HDF5Datatype(HDF5.h5t_copy(HDF5.H5T_C_S1))
       show(dtype)
HDF5 datatype: H5T_STRING {
      STRSIZE 1;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_ASCII;
      CTYPE H5T_C_S1;
   }
julia> close(dtype)
       show(dtype)
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 0:
  #000: H5T.c line 1876 in H5Tget_class(): not a datatype
    major: Invalid arguments to routine
    minor: Inappropriate type
ERROR: Error getting dtype text representation
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
...
```

The unfortunate complicating factor is that it appears all of the built-in data types always return `false` to `isvalid`/`h5i_is_valid` calls:
```julia
julia> native_int = datatype(Int);

julia> isvalid(native_int)
false
```
so if you change the corresponding `show` function to run `isvalid()` checks like the other wrappers do, you get the incorrect/unhelpful statement:
```julia
julia> show(native_int)
HDF5 datatype (invalid)
```

There _might_ be a workaround — if at initialization, the reference counts are manually incremented for all of these types, the types then do show as valid:
```julia
julia> ccall((:H5Iinc_ref, HDF5.libhdf5), Cint, (HDF5.Hid,), native_int.id)
1

julia> isvalid(native_int)
true

julia> show(native_int)
HDF5 datatype: H5T_STD_I64LE
```
Since `datatype()` sets `toclose = false` for all of the native types, the ref count _should_ be maintained if all downstream uses are correct, but that's maybe a big if.

kleinhenz:
Closes #671 

Note we have to use return value from H5LTdtype_to_text to detect closed datatypes
because `H5Iis_valid` does not work for basic atomic types (H5T_NATIVE_INT etc.) since
these types do not have their reference count incremented. We disable automatic error
printing during this probe to prevent spurious error messages. Error reporting is restored
to its previous state at function exit.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -2057,11 +2057,24 @@ function h5l_get_info(link_loc_id::Hid, link_name::String, lapl_id::Hid)
 end
 
 function h5lt_dtype_to_text(dtype_id)
-    len = Ref{Csize_t}()
-    h5lt_dtype_to_text(dtype_id, C_NULL, 0, len)
-    buf = Base.StringVector(len[] - 1)
-    h5lt_dtype_to_text(dtype_id, buf, 0, len)
-    return String(buf)
+    # Note we can't use h5i_is_valid to check if dtype is valid because
+    # ref counts aren't incremented for basic atomic types (e.g. H5T_NATIVE_INT).
+    # Instead just temporarily turn off error printing and try call to probe if dtype is valid.
+    old_func = Ref{Ptr{Cvoid}}()
+    old_client_data = Ref{Ptr{Cvoid}}()
+    h5e_get_auto(H5E_DEFAULT, old_func, old_client_data)
+    h5e_set_auto(H5E_DEFAULT, C_NULL, C_NULL)
+    try
+        len = Ref{Csize_t}()
+        h5lt_dtype_to_text(dtype_id, C_NULL, 0, len)
+        buf = Base.StringVector(len[] - 1)
+        h5lt_dtype_to_text(dtype_id, buf, 0, len)
+        return String(buf)
+    catch
+        return "(invalid)"
+    finally
+        h5e_set_auto(H5E_DEFAULT, old_func[], old_client_data[])
+    end
 end
 
 function h5s_get_simple_extent_dims(space_id::Hid)
--- a/test/plain.jl
+++ b/test/plain.jl
@@ -559,10 +559,8 @@ dspace = dataspace((1,))
 close(dspace)
 @test sprint(show, dspace) == "HDF5Dataspace(-1)"
 
-# TODO: Test HDF5Datatypes once printing of built-in datatypes (i.e. H5T_NATIVE_INT8 and
-# the like) is sorted out. See #671. Commented out because libhdf5 prints.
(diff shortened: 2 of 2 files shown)


kleinhenz:
travis failures are from the flaky crc32c test.

jmert:
I just pushed PR #677 which is going to partly conflict with this.

It's unfortunate that the only way seems to be to possibly invoke and then check for error status, but unfortunately it appears from peeking at the [`H5LTdtype_to_text` source code](https://github.com/HDFGroup/hdf5/blob/9eaeac2e1bb660d9b19a766e38764a6895ae22c4/hl/src/H5LT.c#L2562-L2643) that they just resort to comparing the datatype ID against every built-in before starting the generic handling so there isn't an internal API we could maybe abuse.

One change is that the `show` test in `test/plain.jl` should be uncommented and modified to match the output, and another line should probably be added to test representation of a built-in type, too. (The test uses a copy so that it can be closed and tested for invalid types as well.)

If my PR is merged first, the binding will have to have its error path disabled so that the return value can actually be used here, and if this PR is merged first, I'll make sure to preserve this behavior.

musm:
Hmm, yeah this is kind of an unfortunate hack to work around things. Should we just go ahead and do this until we can implement a more proper fix (a la ref counts) as proposed in the original issue, which would hopefully properly fix it? 

Perhaps add a `FIXME` note in the source?

kleinhenz:
rebased on #677. Yeah this is definitely a bit of a hack but I think it's probably better than the alternative. It seems like a pain to increment everything during the initialization and I would be worried about missing some cases and about changing some expected global state for the library.

gen/bind_generator.jl
```diff
@@ -10,6 +10,7 @@ push!(bind_exceptions, :h5d_create => :H5Dcreate2)
 push!(bind_exceptions, :h5d_open => :H5Dopen2)
 push!(bind_exceptions, :h5e_get_auto => :H5Eget_auto2)
 push!(bind_exceptions, :h5e_set_auto => :H5Eset_auto2)
+push!(bind_exceptions, :h5e_get_auto => :H5Eget_auto2)
```
jmert, line 13:
```suggestion
```
This key is already added two lines above.

musm:
Merging for now. We can explore more 'permanent' or explore other ways we can fix this without having to resort to modifying the error state, which is pretty hacky.

PR #677, opened 2020-09-18
Wrap more of the HDF5 API with generated bindings

jmert:
This wraps more direct uses of `ccall`s with bindings made with the generator.

There are a few notable exceptions due to either (a) memory handling which the binding generator cannot (yet?) deal with, and (b) a case where a call to the underlying `H5Dwrite` is relying on being able to declare the buffer to type `Cstring` for its data validation features.

One more interesting change is to the `h5t_get_member_name` and `h5t_get_tag` functions — this changes to using `h5_free_memory` from `Libc.free` as noted in the [`h5t_get_member_name` documentation](https://portal.hdfgroup.org/display/HDF5/H5T_GET_MEMBER_NAME) (which says it's more important on Windows).

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -41,6 +41,7 @@ include("datafile.jl")
 
 function h5_get_libversion()
     majnum, minnum, relnum = Ref{Cuint}(), Ref{Cuint}(), Ref{Cuint}()
+    # Note: direct ccall because api.jl hasn't been sourced yet.
     status = ccall((:H5get_libversion, libhdf5), Cint, (Ptr{Cuint}, Ptr{Cuint}, Ptr{Cuint}), majnum, minnum, relnum)
     status < 0 && error("Error getting HDF5 library version")
     VersionNumber(majnum[], minnum[], relnum[])
@@ -1965,14 +1966,6 @@ h5l_exists(loc_id::Hid, name::String) = h5l_exists(loc_id, name, H5P_DEFAULT)
 h5o_open(obj_id::Hid, name::String) = h5o_open(obj_id, name, H5P_DEFAULT)
 #h5s_get_simple_extent_ndims(space_id::Hid) = h5s_get_simple_extent_ndims(space_id, C_NULL, C_NULL)
 h5t_get_native_type(type_id::Hid) = h5t_get_native_type(type_id, H5T_DIR_ASCEND)
-function h5r_dereference(obj_id::Hid, ref_type::Integer, pointee::HDF5ReferenceObj)
-    ret = ccall((:H5Rdereference1, libhdf5), Hid, (Hid, Cint, Ref{HDF5ReferenceObj}),
-                obj_id, ref_type, pointee)
-    if ret < 0
-        error("Error dereferencing object")
-    end
-    ret
-end
 
 # Core API ccall wrappers
 include("api.jl")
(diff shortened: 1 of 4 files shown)


src/HDF5.jl
```diff
@@ -41,6 +41,7 @@ include("datafile.jl")
 
 function h5_get_libversion()
```
musm, line 42: just move this to the end of the file

src/HDF5.jl
```diff
@@ -2023,36 +2011,34 @@ function h5s_get_simple_extent_dims(space_id::Hid)
     h5s_get_simple_extent_dims(space_id, dims, maxdims)
     return tuple(reverse!(dims)...), tuple(reverse!(maxdims)...)
 end
+# Note: The following two functions implement direct ccalls because the binding generator
+# cannot (yet) do the string wrapping and memory freeing.
 function h5t_get_member_name(type_id::Hid, index::Integer)
-    pn = ccall((:H5Tget_member_name, libhdf5), Ptr{UInt8}, (Hid, Cuint), type_id, index)
+    pn = ccall((:H5Tget_member_name, libhdf5), Ptr{Cchar}, (Hid, Cuint), type_id, index)
```
musm, line 2017: Can we keep this as Ptr{UInt8} to be consistent with all other  string arguments we call from hdf5 with type `char *`

musm, line 2017:
```suggestion
    pn = ccall((:H5Tget_member_name, libhdf5), Ptr{UInt8}, (Hid, Cuint), type_id, index)
```

src/HDF5.jl
```diff
@@ -2023,36 +2011,34 @@ function h5s_get_simple_extent_dims(space_id::Hid)
...
     s = unsafe_string(pn)
-    Libc.free(pn)
-    s
+    h5_free_memory(pn)
+    return s
 end
 function h5t_get_tag(type_id::Hid)
     pc = ccall((:H5Tget_tag, libhdf5),
-                   Ptr{UInt8},
+                   Ptr{Cchar},
```
musm, line 2027: same comment as above regarding the otuput type

src/HDF5.jl
```diff
@@ -2023,36 +2011,34 @@ function h5s_get_simple_extent_dims(space_id::Hid)
...
 
 function h5f_get_obj_ids(file_id::Hid, types::Integer)
-    sz = ccall((:H5Fget_obj_count, libhdf5), Int, (Hid, UInt32),
-               file_id, types)
-    sz >= 0 || error("error getting object count")
-    hids = Vector{Hid}(undef,sz)
-    sz2 = ccall((:H5Fget_obj_ids, libhdf5), Int, (Hid, UInt32, UInt, Ptr{Hid}),
-          file_id, types, sz, hids)
-    sz2 >= 0 || error("error getting objects")
+    sz = h5f_get_obj_count(file_id, type)
```
musm, line 2039:
```suggestion
    sz = h5f_get_obj_count(file_id, types)
```

musm:
LGTM and a good clean up, just some minor comments!

src/HDF5.jl
jmert: The `read_const` calls are actually relying on this having already been run to set on line 50 where `const libversion = h5_get_libversion()`. That's part of the initialization mess that needs to get sorted out.

(The alternative is to move this lower in the file and instead just call `h5_open()` directly before defining and using `read_const()`.)

src/HDF5.jl
jmert: Sure.

jmert:
I added a second commit that moves the definition of `h5_get_libversion()` and the corresponding use to set the `libversion` constant. That definition was being used to implicitly initialize the library, so to keep the `read_const` function working requires an explicit `ccall` to initialize the library.

It's a bit ugly, but that should maybe just be taken as motivation to work on sorting out the initialization process.

src/HDF5.jl
jmert: Note to self: this must not be tested for this typo to have been made.

jmert:
Hmmm... MacOS nightly failed in both Travis and Github Actions, apparently with `libhdf5` not being defined, which should be imported from HDF5_jll and works as expected everywhere else (and this PR didn't change that).

musm:
Strange, let's drop that commit for now so we can get the rest of the changes in?

jmert:
> Strange, let's drop that commit for now so we can get the rest of the changes in?

I've backed off this branch by one commit. Let's see if MacOS CI is still unhappy or not.

jmert:
Something needs to be fixed in HDF5_jll — I made a branch that just prints a bunch from `deps/deps.jl`, and it's showing that there is no library on MacOS with Julia latest: https://github.com/jmert/HDF5.jl/runs/1138325384?check_suite_focus=true

musm:
Thanks for detecting this bug. Let's see if we can get that fixed.  So I guess this PR discovered a bug in the HDF_jll wrapper which is good.

musm:
osx on master should now be fixed

musm:
want to try that second commit now?

PR #678, opened 2020-09-19
Rename HDF5 constants to C definitions for consistency

musm:
I'm not sure what others think about this, but I thought I'd open up a PR anyways. I never understood the previous convention for these types, which is to capitalize the constant and remove the `_t` postfix. I think `Clang.jl` directly translates the names, so I'm hoping this is at least consistent with that. In any case, we don't apply any special name normalization to other API constants, so I don't see why we should treat these any differently.

### DIFF
--- a/gen/api_defs.jl
+++ b/gen/api_defs.jl
@@ -18,248 +18,248 @@
 ### HDF5 General library functions
 ###
 
-@bind h5_close()::Herr "Error closing the HDF5 resources"
-@bind h5_dont_atexit()::Herr "Error calling dont_atexit"
-@bind h5_free_memory(buf::Ptr{Cvoid})::Herr "Error freeing memory"
-@bind h5_garbage_collect()::Herr "Error on garbage collect"
-@bind h5_get_libversion(majnum::Ref{Cuint}, minnum::Ref{Cuint}, relnum::Ref{Cuint})::Herr "Error getting HDF5 library version"
-@bind h5_open()::Herr "Error initializing the HDF5 library"
-@bind h5_set_free_list_limits(reg_global_lim::Cint, reg_list_lim::Cint, arr_global_lim::Cint, arr_list_lim::Cint, blk_global_lim::Cint, blk_list_lim::Cint)::Herr "Error setting limits on free lists"
+@bind h5_close()::herr_t "Error closing the HDF5 resources"
+@bind h5_dont_atexit()::herr_t "Error calling dont_atexit"
+@bind h5_free_memory(buf::Ptr{Cvoid})::herr_t "Error freeing memory"
+@bind h5_garbage_collect()::herr_t "Error on garbage collect"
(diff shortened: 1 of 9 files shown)


jmert:
I'm OK with this, though I also don't have strong feelings one way or the other.

While you're at it, though, `Hvl_t` should probably be changed to match as `hvl_t` as well.

musm:
Good catch.

Yes part of my motivation is that the naming scheme we have isin't even consistent. As you see we have `Hvl_t`, why did this retain the `_t` postfix and the others not? I'd rather we have a consistent naming convention for everything. Either we directly translate the names as I have done here or change them all to: 1. capitalize the first letter and remove the `_t` postfix. I'm not sure I'm a fan of that, but it's an option.

musm:
For an example of a type that doesn't quite work with the convention of (1) capitalize and (2) remove `_t` post fix is:
```
H5Z_filter_t
```
changing it to 
```
H5z_filter
```
is confusing since you don't really know if it's a type or not since we have removed the post-fix

gen/api_defs.jl
```diff
@@ -18,248 +18,248 @@
...
+@bind h5f_create(pathname::Ptr{UInt8}, flags::Cuint, fcpl_id::hid_t, fapl_id::hid_t)::hid_t error("Error creating file ", pathname)
+@bind h5f_flush(object_id::hid_t, scope::Cint)::herr_t "Error flushing object to file"
+@bind h5f_get_access_plist(file_id::hid_t)::hid_t "Error getting file access property list"
+@bind h5f_get_create_plist(file_id::hid_t)::hid_t "Error getting file create property list"
+@bind h5f_get_intent(file_id::hid_t, intent::Ptr{Cuint})::herr_t "Error getting file intent"
+@bind h5f_get_name(obj_id::hid_t, buf::Ptr{UInt8}, buf_size::Csize_t)::Cssize_t "Error getting file name"
+@bind h5f_get_obj_count(file_id::hid_t, types::Cuint)::Cssize_t "Error getting object count"
+@bind h5f_get_obj_ids(file_id::hid_t, types::Cuint, max_objs::Csize_t, obj_id_list::Ptr{hid_t})::Cssize_t "Error getting objects"
+@bind h5f_get_vfd_handle(file_id::hid_t, fapl_id::hid_t, file_handle::Ptr{Ptr{Cint}})::herr_t "Error getting VFD handle"
+@bind h5f_is_hdf5(pathname::Cstring)::htri_i error("Cannot access file ", pathname)
```
jmert, line 91:
```suggestion
@bind h5f_is_hdf5(pathname::Cstring)::htri_t error("Cannot access file ", pathname)
```
Looks like typo here and in a couple of other places.

musm:
This is going to be a pain to rebase, so I'll just wait until we make an official decision to update and merge.

Option 1:
1. Keep names as defined in HDF5 C lib.

Option 2:
2. Capitalize first word.
3. Remove `_t` postfix

jmert:
I'm generally in favor of option 1 — duplicate the C names.

PR #679, opened 2020-09-19
Fix the flaky CRC32 test through actually setting the track_times property

jmert:
The CRC32 test was flaky because the test depended on how quickly or
slowly the two files could be written to disk. By inserting a small
sleep, we guarantee that the file-creation times will be different
between the two data sets. This was sufficient to force the test to
fail reliably.

The fix then comes actually setting the track_times property. The
implementation of `p_create()` was skipping over the property because
the `H5P_DATASET_CREATE != H5P_OBJECT_CREATE`. By special-casing the
`H5P_OBJECT_CREATE` value in the dictionary, we can recognize it as
acting upon any of the three file/dataset/group creation property
classes.

An unrelated change is to inline the call to the property setter
function since it avoids repeating the dictionary lookup done by
`setindex!(::HDF5Properties, ...)` when using the `p[k] = v` syntax.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -980,9 +980,14 @@ a_create(parent::Union{HDF5File, HDF5Object}, name::String, dtype::HDF5Datatype,
 function p_create(class; pv...)
   p = HDF5Properties(h5p_create(class), class)
   for (k,v) in pairs(pv)
-    if hdf5_prop_get_set[k][3] == class
-      p[k] = v
+    funcget, funcset, classprop = hdf5_prop_get_set[k]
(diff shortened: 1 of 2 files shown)


musm:
Out of curiosity, does the HDF5 documentation say anything about this?

jmert:
What specifically? The fact that `:track_times` as an H5P_OBJECT_CREATE is applicable to files, datasets, and groups? That I just discovered empirically by trying out various property types since the docs didn't say exactly what is valid.

musm:
Yeah. I see, ok sounds good, I was just curious, because I couldn't find anything in the docs.

PR #680, opened 2020-09-20
Rename constant reading and use api for hdf5 lib version

musm:
@jmert  I went ahead with the commit you dropped

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -55,12 +55,12 @@ abstract  type Hmpih end
 primitive type Hmpih32 <: Hmpih 32 end # MPICH C/Fortran, OpenMPI Fortran: 32 bit handles
 primitive type Hmpih64 <: Hmpih 64 end # OpenMPI C: pointers (mostly 64 bit)
 
-# Private function to extract exported library constants
-# Need to call h5open to ensure library is initalized before reading these constants
-# Altough these are runtime initalized constants, in practice their values are constant so we can precompile for improved latency
+# Private function to extract exported global library constants.
+# Need to call H5open to ensure library is initalized before reading these constants.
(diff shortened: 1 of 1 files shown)


src/HDF5.jl
```diff
@@ -65,10 +55,12 @@ abstract  type Hmpih end
...
-# Function to extract exported library constants
-# Kudos to the library developers for making these available this way!
-const libhdf5handle = Libdl.dlopen(libhdf5)
-read_const(sym::Symbol) = unsafe_load(convert(Ptr{Hid}, Libdl.dlsym(libhdf5handle, sym)))
+# Private function to extract exported library constants
+# Need to call h5open to ensure library is initalized before reading these constants
+# Altough these are runtime initalized constants, in practice their values are constant so we can precompile for improved latency
+let libhdf5handle = Libdl.dlopen(libhdf5)
+    global _read_const(sym::Symbol) = (ccall((:H5open, libhdf5), Herr, ()); unsafe_load(convert(Ptr{Hid}, Libdl.dlsym(libhdf5handle, sym))))
+end
```
jmert, line 63:
```suggestion
let libhdf5handle = Ref(Libdl.dlopen(libhdf5))
    ccall(Libdl.dlsym(libhdf5handle[], :H5open), Herr, ())
    global read_const(sym::Symbol) = unsafe_load(cglobal(Libdl.dlsym(libhdf5handle[], sym), Hid))
end
```

The library only needs to be initialized once, so there should be no point in having the `H5open` call at each invocation.

Then using a `Ref` serializes the null pointer into the precompiled function which protects against future invalid use.

Testing this construction against master, I get essentially unchanged times to precompile and then to load the precompiled image — on my machine, about 3.13s to precompile, and then on average ~0.7s to load after that.

musm: Thanks!

PR #681, opened 2020-09-20
Speed up (pre)compile and load times

jmert:
This PR does two primary things with the goal of reducing load time:

1. The three global `Dict`s that are rehashed on init are eliminated, replaced with `if-else` chains at their use or in new internal helper functions. This was motivated by looking at the output of `@snoopi using HDF5` (where I started Julia with `julia --compiled-modules=no`) and noticed that many of the methods near the end with largest times were `Dict` methods.

One minor advantage of doing this is that property names can now be valid for multiple HDF5 classes. For instance, the `__init__` code had to call `h5p_set_char_encoding` for `ASCII_ATTRIBUTE_PROPERTIES` and `UTF8_ATTRIBUTE_PROPERTIES` because `:char_encoding` was already mapped to `H5P_LINK_CREATE` so couldn't be used for these two properties which are `H5P_ATTRIBUTE_CREATE` properties. After replacing the `Dict` lookup, though, the `_prop_set!` function can have a `:char_encoding` case for `H5P_ATTRIBUTE_CREATE` without any conflicts.

To evaluate the effectiveness of this change, I've timed loading (`using HDF5`) with a missing precompiled cache, loading with a valid precompiled cache, and the time to run the entire test suite (with the precompiled cache active). The times comparing current master versus the first commit from this PR (run with near-latest Julia, average and std over 10 runs) show:

```
 master precompile: 3.145 ± 0.0295
no dict precompile: 1.927 ± 0.0164

 master pkg load: 0.706 ± 0.0084
no dict pkg load: 0.564 ± 0.0158

 master pkg test: 32.016 ± 0.9132
no dict pkg test: 31.306 ± 0.4119
```

2. The second commit enables a lower optimization level for the entire module in Julia 1.5+. The motivation here is that a lot of functions do not type infer, so we might as well tell the compiler to not try too hard.

Running the same timing tests again:
```
 master precompile: 3.145 ± 0.0295
no dict precompile: 1.927 ± 0.0164
low opt precompile: 1.860 ± 0.0156

 master pkg load: 0.706 ± 0.0084
no dict pkg load: 0.564 ± 0.0158
low opt pkg load: 0.502 ± 0.0123

 master pkg test: 32.016 ± 0.9132
no dict pkg test: 31.306 ± 0.4119
low opt pkg test: 27.085 ± 0.4015
```

---

* I've noticed at least one case where type inference improves despite the lower optimization level: `getindex(::Union{HDF5File, HDF5Group}, ::String)` used to infer as `Any` but with this PR is inferred returning an `HDF5Object`.

* My hacked together timing script can be found in [this gist](https://gist.github.com/jmert/c757ba90f98d10b9eb7ece12117da6e7)

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -270,18 +270,27 @@ const HDF5Scalar = Union{HDF5BitsKind, HDF5ReferenceObj}
 const ScalarOrString = Union{HDF5Scalar, String}
 
 # It's not safe to use particular id codes because these can change, so we use characteristics of the type.
-const hdf5_type_map = Dict(
-    (H5T_INTEGER, H5T_SGN_2, Csize_t(1)) => Int8,
-    (H5T_INTEGER, H5T_SGN_2, Csize_t(2)) => Int16,
-    (H5T_INTEGER, H5T_SGN_2, Csize_t(4)) => Int32,
-    (H5T_INTEGER, H5T_SGN_2, Csize_t(8)) => Int64,
-    (H5T_INTEGER, H5T_SGN_NONE, Csize_t(1)) => UInt8,
-    (H5T_INTEGER, H5T_SGN_NONE, Csize_t(2)) => UInt16,
-    (H5T_INTEGER, H5T_SGN_NONE, Csize_t(4)) => UInt32,
-    (H5T_INTEGER, H5T_SGN_NONE, Csize_t(8)) => UInt64,
-    (H5T_FLOAT, nothing, Csize_t(4)) => Float32,
-    (H5T_FLOAT, nothing, Csize_t(8)) => Float64,
-)
+function _hdf5_type_map(class_id, is_signed, native_size)
+    if class_id == H5T_INTEGER
+        if is_signed == H5T_SGN_2
+            return native_size === Csize_t(1) ? Int8 :
+                   native_size === Csize_t(2) ? Int16 :
+                   native_size === Csize_t(4) ? Int32 :
+                   native_size === Csize_t(8) ? Int64 :
+                   throw(KeyError(class_id, is_signed, native_size))
+        else
+            return native_size === Csize_t(1) ? UInt8 :
+                   native_size === Csize_t(2) ? UInt16 :
+                   native_size === Csize_t(4) ? UInt32 :
+                   native_size === Csize_t(8) ? UInt64 :
+                   throw(KeyError(class_id, is_signed, native_size))
+        end
+    else
+        return native_size === Csize_t(4) ? Float32 :
+               native_size === Csize_t(8) ? Float64 :
+               throw(KeyError(class_id, is_signed, native_size))
+    end
+end
 
 hdf5_type_id(::Type{S}) where {S<:AbstractString}  = H5T_C_S1
 
@@ -889,16 +898,19 @@ root(obj::Union{HDF5Group, HDF5Dataset}) = g_open(file(obj), "/")
 getindex(dset::HDF5Dataset, name::String) = a_open(dset, name)
 getindex(x::HDF5Attributes, name::String) = a_open(x.parent, name)
 
-const hdf5_obj_open = Dict(
-    H5I_DATASET   => (d_open, H5P_DATASET_ACCESS, H5P_DATASET_XFER),
-    H5I_DATATYPE  => (t_open, H5P_DATATYPE_ACCESS),
-    H5I_GROUP     => (g_open, H5P_GROUP_ACCESS),
-)
 function getindex(parent::Union{HDF5File, HDF5Group}, path::String; pv...)
     objtype = gettype(parent, path)
-    fun, propids = hdf5_obj_open[objtype]
-    props = [p_create(prop; pv...) for prop in propids]
-    obj = fun(parent, path, props...)
+    if objtype == H5I_DATASET
+        dapl = p_create(H5P_DATASET_ACCESS; pv...)
+        dxpl = p_create(H5P_DATASET_XFER; pv...)
+        return d_open(parent, path, dapl, dxpl)
+    elseif objtype == H5I_GROUP
+        gapl = p_create(H5P_GROUP_ACCESS; pv...)
+        return g_open(parent, path, gapl)
+    else#if objtype == H5I_DATATYPE # only remaining choice
+        tapl = p_create(H5P_DATATYPE_ACCESS; pv...)
+        return t_open(parent, path, tapl)
+    end
 end
 
 # Path manipulation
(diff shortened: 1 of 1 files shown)


src/HDF5.jl
```diff
@@ -277,18 +281,27 @@ const HDF5Scalar = Union{HDF5BitsKind, HDF5ReferenceObj}
...
-    (H5T_FLOAT, nothing, Csize_t(8)) => Float64,
-)
+function _hdf5_type_map(class_id, is_signed, native_size)
+    if class_id == H5T_INTEGER
+        if is_signed == H5T_SGN_2
+            native_size === Csize_t(1) ? (return Int8) :
+            native_size === Csize_t(2) ? (return Int16) :
+            native_size === Csize_t(4) ? (return Int32) :
+            native_size === Csize_t(8) ? (return Int64) :
+            throw(KeyError(class_id, is_signed, native_size))
```
jmert, line 291: I'm not sure this line (and similar in the other branches) is actually necessary — the final check size-8 case in principle could just be the else-case of the size-4 ternary. Thoughts?

src/HDF5.jl
```diff
@@ -277,18 +281,27 @@ const HDF5Scalar = Union{HDF5BitsKind, HDF5ReferenceObj}
...
-    (H5T_INTEGER, H5T_SGN_2, Csize_t(4)) => Int32,
-    (H5T_INTEGER, H5T_SGN_2, Csize_t(8)) => Int64,
-    (H5T_INTEGER, H5T_SGN_NONE, Csize_t(1)) => UInt8,
-    (H5T_INTEGER, H5T_SGN_NONE, Csize_t(2)) => UInt16,
-    (H5T_INTEGER, H5T_SGN_NONE, Csize_t(4)) => UInt32,
-    (H5T_INTEGER, H5T_SGN_NONE, Csize_t(8)) => UInt64,
-    (H5T_FLOAT, nothing, Csize_t(4)) => Float32,
-    (H5T_FLOAT, nothing, Csize_t(8)) => Float64,
-)
+function _hdf5_type_map(class_id, is_signed, native_size)
```
musm, line 284: 👍

src/HDF5.jl
```diff
@@ -896,16 +909,19 @@ root(obj::Union{HDF5Group, HDF5Dataset}) = g_open(file(obj), "/")
...
-    H5I_DATASET   => (d_open, H5P_DATASET_ACCESS, H5P_DATASET_XFER),
-    H5I_DATATYPE  => (t_open, H5P_DATATYPE_ACCESS),
-    H5I_GROUP     => (g_open, H5P_GROUP_ACCESS),
-)
 function getindex(parent::Union{HDF5File, HDF5Group}, path::String; pv...)
     objtype = gettype(parent, path)
-    fun, propids = hdf5_obj_open[objtype]
-    props = [p_create(prop; pv...) for prop in propids]
-    obj = fun(parent, path, props...)
+    if objtype == H5I_DATASET
```
musm, line 914: 👍 
I also bet this is just simply far faster

src/HDF5.jl
```diff
@@ -2216,20 +2250,12 @@ function __init__()
...
-    h5p_set_create_intermediate_group(ASCII_LINK_PROPERTIES[].id, 1)
-    UTF8_LINK_PROPERTIES[] = p_create(H5P_LINK_CREATE)
-    h5p_set_char_encoding(UTF8_LINK_PROPERTIES[].id, H5T_CSET_UTF8)
-    h5p_set_create_intermediate_group(UTF8_LINK_PROPERTIES[].id, 1)
-    ASCII_ATTRIBUTE_PROPERTIES[] = p_create(H5P_ATTRIBUTE_CREATE)
-    h5p_set_char_encoding(ASCII_ATTRIBUTE_PROPERTIES[].id, H5T_CSET_ASCII)
-    UTF8_ATTRIBUTE_PROPERTIES[] = p_create(H5P_ATTRIBUTE_CREATE)
-    h5p_set_char_encoding(UTF8_ATTRIBUTE_PROPERTIES[].id, H5T_CSET_UTF8)
-
-    rehash!(hdf5_type_map, length(hdf5_type_map.keys))
```
musm, line 2230: yay, glad we can get rid of this now

musm:
> I've noticed at least one case where type inference improves despite the lower optimization level: getindex(::Union{HDF5File, HDF5Group}, ::String) used to infer as Any but with this PR is inferred returning an HDF5Object.

It might then just be worth it to add type annotations where necessary to help the compiler with inference in these cases, instead of (or in addition to)   ` @eval Base.Experimental.@optlevel 1` . 

I kind of think adding the line
```
    @eval Base.Experimental.@optlevel 1
```
Is not that elegant, given it is current experimental status, however the precompilation gains (10% faster) and improved inference are very nice gains.

jmert:
> > I've noticed at least one case where type inference improves despite the lower optimization level: getindex(::Union{HDF5File, HDF5Group}, ::String) used to infer as Any but with this PR is inferred returning an HDF5Object.
> 
> It might then just be worth it to add type annotations where necessary to help the compiler with inference in these cases, instead of (or in addition to) ` @eval Base.Experimental.@optlevel 1` .

Yeah, doing some more systematic review of types and inference's understanding of the code is something worth doing but wasn't as easy of low-hanging fruit. And just because I was maybe a bit unclear, the inference change of `getindex` is due to rewriting it to not use a dictionary and is independent of the `@optlevel 1` addition (i.e. we'll get it even if the second commit is dropped).

src/HDF5.jl
```diff
@@ -978,19 +994,62 @@ end
 t_commit(parent::Union{HDF5File, HDF5Group}, path::String, dtype::HDF5Datatype) = t_commit(parent, path, dtype, p_create(H5P_LINK_CREATE))
 
 a_create(parent::Union{HDF5File, HDF5Object}, name::String, dtype::HDF5Datatype, dspace::HDF5Dataspace) = HDF5Attribute(h5a_create(checkvalid(parent).id, name, dtype.id, dspace.id), file(parent))
+
+function _prop_set!(p::HDF5Properties, name::Symbol, val, check::Bool = true)
```
musm, line 998: Is this `check` arg really needed? I'm not sure, but it seems a bit superfluous?   I think the rationale is that we can squeeze a bit of perf without having to  `error` check, but perhaps the compiler is good enough that we don't need such fine grained control?

musm (approved):

src/HDF5.jl
```diff
@@ -1,5 +1,9 @@
 module HDF5
 
+if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@optlevel"))
```
musm, line 3: Perhaps we could save this for another PR ?

musm: The only reason I say this is that it is not directly related to the rest of the changes in the PR. Perhaps it also should have some further discussion.

src/HDF5.jl
musm: EDIT:

Actually NVM I don't think it's that feasible given the way the code is structured as is, and would remove some important error checking.

jmert: > Is this check arg really needed, It seems a bit superfluous?

The reason I added it was to maintain the current behavior — `p_create` does not error even if you pass it properties that are not appropriate for a particular property class, while `setindex!` did (by way of just throwing when the underlying `h5p_set_*` function errored). The two functions now set `false` and `true`, respectively, to maintain that.

src/HDF5.jl
jmert: Done - #682

src/HDF5.jl
```diff
@@ -971,19 +983,62 @@ end
...
+               name === :track_times ? h5p_set_obj_track_times(p, val...) : # H5P_OBJECT_CREATE
+               check ? error("unknown dataset create property ", name) : nothing
+    end
+
+    if class == H5P_ATTRIBUTE_CREATE
+        return name === :char_encoding ? h5p_set_char_encoding(p, val...) :
+               check ? error("unknown attribute create property ", name) : nothing
+    end
+
+    check ? error("unknown property class ", class) : return nothing
```
musm, line 1033:
```suggestion
    return check ? error("unknown property class ", class) : nothing
```

I guess these could also just be a series of if .. ifelse.. although that doesn't matter

musm:
Awesome !

musm:
I kinda wonder if we should just disable the integration tests, it's pretty annoying to see red X on PRs that pass.

PR #684, opened 2020-09-22
Fix #683, MPI properties

### LINKED ISSUE #683: Parallel HDF5 is broken by commit d1e542a
Hi - I'm not sure if you are doing a series of commits as a work in progress, but just so you know - commit d1e542a breaks loading HDF5 with MPI (to get parallel hdf5).  [mpio.jl](https://github.com/JuliaIO/HDF5.jl/blob/master/src/mpio.jl) uses `hdf5_prop_get_set` (see line 57). That dictionary was removed in commit d1e542a (was in HDF5.jl around line 2138).

The result of this change is that 
```
using MPI
using HDF5
``` 
spews a warning about the missing dictionary and it seems that one MPI rank that I make fails to do anything. 

I hope you're planning to update `mpio.jl` accordingly in a subsequent release. Let me know if you need help testing.

jmert:
This should fix the load/use error that @lyon-fnal reported, but I haven't tested because I don't have an MPI-capable HDF5 library on my machine.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -995,6 +995,7 @@ function _prop_set!(p::HDF5Properties, name::Symbol, val, check::Bool = true)
 
     if class == H5P_FILE_ACCESS
         return name === :alignment     ? h5p_set_alignment(p, val...) :
+               name === :fapl_mpio     ? h5p_set_fapl_mpio(p, val...) :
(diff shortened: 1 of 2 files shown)


lyon-fnal:
As I wrote in #683, this PR seems to work fine.

PR #685, opened 2020-09-30
Define `getindex` for properties and rearrange low-level API helpers

jmert:
The first commit adds the missing symmetry from #681 wherein `setindex!` was defined for `HDF5Properties` but the corresponding `getindex` method was not. This adds the `getindex` method and thereby also recovers the information that was lost when the dictionary which contained a list of the accessor functions (but which were unused) was deleted. I've also added a few property accessors which weren't originally defined.

In doing so, I've started the process of stripping out more code from `src/HDF5.jl` and putting it in another file. The methods are moved/added to a new `src/api_helpers.jl` file, with the contents primarily being simple wrappers that turn C output argument pointers to `Ref`/`Array`s into (tuples of) return values.

The second commit just continues the trend with other low-level helpers outside of the property interface — more functions with out-arguments are wrapped up and placed in `api_helpers.jl`.

One of the advantages of systematically working through these methods is that I've changed a number of `Vector{UInt8}` arrays to `StringVector`s, which as I understand it should avoid a copy in modern Julia versions. I also took the opportunity to change some of the definitions in `gen/api_defs.jl` to more closely match the C types and variable names and changed some types from `Ptr` to `Ref` to signal that the arguments should be to Julia-owned data.

TODO:
- [x] Tests for `getindex(::HDF5Properties, ::Symbol)` calls
- [x] Audit out-argument uses in `src/blosc_filter.jl`.
- [ ] Figure out what to do about the many overloads of the low-level API (e.g. https://github.com/JuliaIO/HDF5.jl/blob/f0d6d626a6742944dd5f9ffbb673078ce6e11c1a/src/HDF5.jl#L1975-L2026)


musm:
Sorry for butchering this PR by merging the massive rename,  but this will need to be rebased

jmert:
That's OK — I knew that was a possibility. I should be able to get pretty far with just simple `s/Hid/hid_t/g`-like regexs in `vim`.

jmert:
Regarding the TODO item in the top comment:
> Figure out what to do about the many overloads of the low-level API ...

I'm inclined to leave that to another PR — my suspicion is that most of those should be overloads of e.g. `a_write` instead of `h5a_write` and similar, but I haven't taken the time yet to find where these methods are being called from to see if there's a good reason for keeping the status quo. In any case, I think that should be separate from this PR which aims to just get `getindex(::HDF5Properties, ...)` added, and some extra code organization is (I hope) a nice plus.

musm:
Most of this looks good, I haven't reviewed in detail yet, but I think it's good cleanup and re-org. I need to double check the `Cstring` vs `Vector{UInt8}` usage.

> I'm inclined to leave that to another PR — my suspicion is that most of those should be overloads of e.g. a_write instead of h5a_write and similar, but I haven't taken the time yet to find where these methods are being called from to see if there's a good reason for keeping the status quo. In any case, I think that should be separate from this PR which aims to just get getindex(::HDF5Properties, ...) added, and some extra code organization is (I hope) a nice plus.

I'm in agreement.

jmert:
> Most of this looks good, I haven't reviewed in detail yet, but I think it's good cleanup and re-org. I need to double check the Cstring vs Vector{UInt8} usage.

As far as I understand it, since it's not an output argument where there could be an uninitialized argument, the type can be "strengthened" to `Cstring` to signal that it's expected to have a valid C-string. I did test calling `h5a_get_name_by_idx` with the second argument being a non-trivial Unicode name, and it still worked, and adding an embedded null to the string caused Julia to throw an error.

I suppose that extra check does have some runtime cost, so translating the C prototype to the most informative might not be desired.

src/HDF5.jl
```diff
@@ -1062,6 +1115,7 @@ o_copy(src_obj::HDF5Object, dst_parent::Union{HDF5File, HDF5Group}, dst_path::St
 setindex!(dset::HDF5Dataset, val, name::String) = a_write(dset, name, val)
 setindex!(x::HDF5Attributes, val, name::String) = a_write(x.parent, name, val)
 # Getting and setting properties: p[:chunk] = dims, p[:compress] = 6
+getindex(p::HDF5Properties, name::Symbol) = _prop_get(p, name)
```
musm, line 1119: do we need to check `p` is valid?

jmert: Yeah, good catch.

src/HDF5.jl
```diff
@@ -984,6 +984,59 @@ t_commit(parent::Union{HDF5File, HDF5Group}, path::String, dtype::HDF5Datatype)
...
+    if class == H5P_LINK_CREATE
+        return name === :char_encoding ? h5p_get_char_encoding(p) :
+               name === :create_intermediate_group ? h5p_get_create_intermediate_group(p) :
+               error("unknown link create property ", name)
+    end
+
+    if class == H5P_DATASET_CREATE
+        return name === :alloc_time  ? h5p_get_alloc_time(p) :
+               name === :chunk       ? get_chunk(p) :
+               #name === :external    ? h5p_get_external(p) :
```
musm, line 1022: any reason this is commented out?

jmert: Yes — I haven't figured out how to actually correctly implement the get part of the external properties. The corresponding C calls aren't actually defined yet, and the original Dict just had `nothing` for the getter part. It looks like there's probably something useful to return, but I don't understand the external interface yet, so I just left that for a follow-up to not slow me down. (I actually have defining the corresponding bindings in a `git stash` just waiting the motivation to figure it out.)

src/HDF5.jl
```diff
@@ -2153,23 +2200,12 @@ function get_chunk(dset::HDF5Dataset)
...
-    out[]
-end
-function get_libver_bounds(p::HDF5Properties)
-    out1 = Ref{Cint}()
-    out2 = Ref{Cint}()
-    h5p_get_libver_bounds(p.id, out1, out2)
-    out1[], out2[]
-end
+
+get_alignment(p::HDF5Properties) = h5p_get_alignment(checkvalid(p).id)
```
musm, line 2083: I guess previously these didn't `checkvalid`, I guess neither do the setters. I'm kind of wondering if we need a more thorough audit.

jmert: There definitely needs to be an audit at some point. I added them here because `get_alignment` in my head is "higher level" than `h5p_get_alignment`, but that was a semi-arbitrary decision.

musm: I agree that if we add them they should be in this function. I was noting the behavioral difference, that this now checks them where as we didn't' before. The setters, also seem to be missing checks?

jmert: I'm about to add a check to the `setindex!(::HDF5Properties, ...)` that's missing, so that'll handle most uses where properties are set through the keyword interface. It does look like, yes, there are still some stragglers like `set_chunk` that should probably gain a `checkvalid`, though.

src/HDF5.jl
```diff
@@ -1535,11 +1563,11 @@ function readmmap(obj::HDF5Dataset, ::Type{Array{T}}) where {T}
     end
     if !Sys.iswindows()
         local fdint
-        prop = h5d_get_access_plist(checkvalid(obj).id)
+        prop = h5d_get_access_plist(obj.id)
```
jmert, line 1566: I removed this use of `checkvalid` here because calling `size(obj)` already checks for validity.

src/HDF5.jl
musm: So since `getindex` does `checkvalid`, and the `get_*` functions also do the checking, this will end up double checking everything. 

Perhaps we should only check at the highest level function which would be `getindex`? Of course this would then mean independent calls to the getters would not check, which could be problematic.

src/HDF5.jl
```diff
@@ -130,6 +130,7 @@ const H5I_DATASPACE    = 4
 const H5I_DATASET      = 5
 const H5I_ATTR         = 6
 const H5I_REFERENCE    = 7
+const H5I_VFL          = 8
```
musm, line 133: Should we just round out this enum and add

```
H5I_GENPROP_CLS = 8 
H5I_GENPROP_LST = 9  
H5I_ERROR_CLASS = 10 
H5I_ERROR_MSG = 11 
H5I_ERROR_STACK = 12 
H5I_NTYPES = 13 
``` 
?

src/HDF5.jl
```diff
@@ -1107,27 +1162,11 @@ exists(parent::Union{HDF5File, HDF5Group}, path::String) = exists(parent, path,
...
-    H5GINFO_TEMP_ARRAY[]
-end
-
-const H5OINFO_TEMP_ARRAY = Ref{H5Oinfo}()
-function objinfo(obj::Union{HDF5File, HDF5Object})
-    h5o_get_info(obj.id, H5OINFO_TEMP_ARRAY)
-    H5OINFO_TEMP_ARRAY[]
-end
-
-const LENGTH_TEMP_ARRAY = Ref{UInt64}()
```
musm, line 1122: Good change 👍

src/HDF5.jl
jmert: No — for all properties except `:chunk`, `_prop_get` calls the `h5p_get_*` functions, so the check won't be repeated. (`get_chunk` is used instead of `h5p_get_chunk` because the former is what handles reversing the order of dimensions.)

src/HDF5.jl
jmert: Sure, might as well.

src/HDF5.jl
musm: Oh that's right

src/HDF5.jl
jmert: It looks like the enum changed in HDF5 v1.12, and my computer's header doesn't give the same numbering, and I assume you've got an off-by-one in the list you added, with the values supposed to be starting at 9? That would seem to match up with [this handy diff](https://fossies.org/diffs/hdf5/1.10.6_vs_1.12.0/src/H5Ipublic.h-diff.html) I found online

src/HDF5.jl
```diff
@@ -1973,9 +1996,8 @@ function get_mem_compatible_jl_type(objtype::HDF5Datatype)
             return NamedTuple{Symbol.(membernames), Tuple{membertypes...}}
         end
     elseif class_id == H5T_ARRAY
-        nd = h5t_get_array_ndims(objtype.id)
-        dims = Vector{hsize_t}(undef,nd)
-        h5t_get_array_dims(objtype.id, dims)
+        dims = h5t_get_array_dims(objtype.id)
```
musm, line 1999: I guess we could also just have `h5t_get_array_dims` return a tuple of `(dims,ndims)` but that may not be worth it?

src/HDF5.jl
musm: They changed the enum in v1.12, well, let's not change this then until we sync up the BinaryBuilder hdf5 library.

src/HDF5.jl
jmert: That's an interesting idea — I'd been thinking of these helper functions doing just the minimum amount of work required to turn out-arguments into return values. Since there are several cases like this one, though, where the implementation is a sequence of two calls that both have meaning, I suppose you could argue that it should return all relevant return/out values. In that case, I guess I'd naturally sequence it as `(ndims, dims)` since the `ndims` call is the "first return", but I'm not set on that interpretation necessarily.

musm: Yeah I prefer the reverse ordering, since it's usually the second call that does the 'important' call. I'm not dead set on changing it, just thought I'd mention that this effectively does more work than necessary since the internal already calculated `ndims`

src/api_helpers.jl
```diff
@@ -0,0 +1,297 @@
...
+
+###
+### Optimized Functions Interface
+###
+
+###
+### HDF5 Lite Interface
+###
+
+function h5lt_dtype_to_text(dtype_id)
```
musm, line 271: unrelated, but lol this is such a hack, there's got to be a better way of doing this somehow.

src/HDF5.jl
jmert: Sorry, I'm not quite sure what you're stuggesting — are you saying not to add just the extra ones you suggested, or you'd like to not add the `H5I_VFL` constant as well? I added just the one value because I was trying to test the `fapl[:driver]` property, and the actual return value is an even-harder-to-access constant than just verifying that it has the right type. (Though I suppose any non-erroring call is valid, so the test could just be `@test fapl[:driver] isa hid_t` and it'd probably be fine.)

musm: >  are you saying not to add just the extra ones you suggested,

Yes to this. Ignore my suggestion to add the extra ones.

src/HDF5.jl
musm: ordering `(dims, ndims)` would also kinda align it with `h5s_get_simple_extent_dims` which returns `(dims, maxdims)`

jmert: What about the other similar cases (that don't lead to making a `String`) — `h5f_get_obj_ids` and `h5p_get_chunk`? Also, would `h5s_get_simple_extent_dims` in this scheme actually gain a third return value so it'd be `(dims, maxdims, ndims)` for consistency?

musm:
Just reviewed these changes and they look good to me.

src/HDF5.jl
musm: If we decide to go with this (reverse) ordering, I would argue (in addition to consistency) that there isn't any harm in including them as outputs from a performance an usage standpoint. In most cases, we probably don't need them, anyways, but they are available if needed.

This is also why I don't feel very strongly about changing them to include all outputs, but it does seem 'nice' to use this data as it _is_ after all available.

musm (approved):

src/HDF5.jl
jmert: I'm fine with that reasoning. Make the changes in this PR, or another one?

musm: Either is fine with me. Is this PR still WIP/RFC?

jmert: No, not still in progress. I removed the WIP flag but forgot to change the title. I can make the couple changes for this, though, and push for a final run through CI.

jmert: Ah, test suites! The naive thing to do actually does break a test as written. So maybe it will be better to follow-up in another PR.

musm:
>  Figure out what to do about the many overloads of the low-level API (e.g.

yeah not sure about this either. 



So what we currently have is 
api  --- low level 'dumb' wrappers
api_helpers -- higher level 'smart' wrappers (that handle type marshaling etc.) 

It might not make too much sense to try to pigeon hole anything more into 'api_helpers' other than it's  intended purpose as a  
smarter  wrapper over those in `api`,  and to leave every other function in the top-level ?

jmert:
> It might not make too much sense to try to pigeon hole anything more into 'api_helpers' other than it's intended purpose as a
> smarter wrapper over those in `api`, and to leave every other function in the top-level ?

Yeah, my working idea has been that `api`, `api_helpers`, and an upcoming `api_types` are the three files necessary to basically define the low-level interface. Then everything else is used to build up the high-level interface.

In that regard, it makes the overloading of `h5*_` functions kind of break the assumption that the `h5*_` are the low-level API and the `?_{create,read,write}` are the high-level API.

PR #686, opened 2020-09-30
Re-add `read(dset, String)` form

### LINKED ISSUE #669: Regression in typed `read`
The following behavior to use a read statement like `read(dataset, Type)` has regressed for `String`s in current master versus v0.13.6:

With v0.13.6
```julia
(test) pkg> add HDF5@0.13.6
...

julia> using HDF5

julia> h5open("test.h5", "w") do hfile
           hfile["dset"] = "string value"
       end
"string value"

julia> h5open("test.h5", "r") do hfile
           read(hfile["dset"], String)
       end
"string value"
```

Then on current master:
```julia
(test) pkg> add HDF5#master
...

julia> using HDF5

julia> h5open("test.h5", "w") do hfile
           hfile["dset"] = "string value"
       end
"string value"

julia> h5open("test.h5", "r") do hfile
           read(hfile["dset"], String)
       end
ERROR: Type String does not have a definite size.
Stacktrace:
 [1] sizeof
   @ ./essentials.jl:457 [inlined]
 [2] read(::HDF5Dataset, ::Type{String})
   @ HDF5 ~/.julia/packages/HDF5/iXIFC/src/HDF5.jl:1324
...
```

jmert:
This would close #669. Adds a method overload for the generic `read()` function introduced in #652 that specializes on the `String` type to make reading as a specific type easier for the string case. As I suggested in https://github.com/JuliaIO/HDF5.jl/issues/669#issuecomment-698022510, the idea is that the type argument is a natural way to specify particular deserializations of datasets, and the HDF5 string types are deserialized to Julia `Strings`.

In the course of testing, I found that there's actually a way to manually specify the underlying `Cstring` or `FixedString` types on the generic read that caused it to read "junk" values by [forcibly] misinterpreting the returned data during normalization. I've added a check that prevents that from being possible.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1435,6 +1435,18 @@ function read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
         return out
     end
 end
+# `Type{String}` does not have a definite size, so the previous method does not accept
+# it even though it will return a `String`. This explicit overload allows that usage.
+function read(obj::DatasetOrAttribute, ::Type{String}, I...)
+    dtype = datatype(obj)
+    try
+        T = get_jl_type(dtype)
+        T <: Union{Cstring, FixedString} || error(name(obj), " cannot be read as type `String`")
+        return read(obj, T, I...)
+    finally
+        close(dtype)
+    end
+end
 
 # Read OPAQUE datasets and attributes
 function read(obj::DatasetOrAttribute, ::Type{HDF5Opaque})
--- a/test/plain.jl
+++ b/test/plain.jl
@@ -46,7 +46,9 @@ let
     HDF5.h5t_set_cset(dtype.id, HDF5.cset(typeof(salut)))
     dspace = HDF5.dataspace(salut)
     dset = HDF5.d_create(f, "salut-vlen", dtype, dspace)
-    HDF5.h5d_write(dset, dtype, HDF5.H5S_ALL, HDF5.H5S_ALL, HDF5.H5P_DEFAULT, [pointer(salut)])
+    GC.@preserve salut begin
+        HDF5.h5d_write(dset, dtype, HDF5.H5S_ALL, HDF5.H5S_ALL, HDF5.H5P_DEFAULT, [pointer(salut)])
+    end
 end
 # Arrays of strings
 salut_split = ["Hi", "there"]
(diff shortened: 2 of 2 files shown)


src/HDF5.jl
```diff
@@ -1387,6 +1387,14 @@ function read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
               sizeof($memtype) = $(h5t_get_size(memtype.id))
               """)
     end
+    if h5t_get_class(memtype.id) == HDF5.H5T_STRING
```
kleinhenz, line 1390: I'm not sure that these checks are a good idea here. I envisioned this function as being somewhat unsafe and allowing you to read into basically any datatype as long as the sizes match. I think having this kind of interface which is 'high level' in some sense but still lets you do what you want is useful to have. For example I could imagine wanting to read data from a fixed string field into something other than a `FixedString`.

kleinhenz:
Responding to your comment on #669: I agree that `read(::DatasetOrAttribute, ::Type{T})` is a nice way to define custom deserializations and that we should try to make sure that it works as conveniently as possible with the rest of the library. This is what I tried to do with the `name_type_pair::Pair{String, DataType}` methods. I guess I'm not sure if it is necessary to define for `String` since this type is already special cased in the internal machinery for mapping from hdf5 to julia. However, I can't really think of much downside to having this method so this seems fine to me.

src/HDF5.jl
jmert: Sure, I'd be fine dropping this check specifically. I only noticed it because `"variable"` happens to be 8 bytes, so it make it past the size check when you try to read the variable pointer as a fixed string. Just thought I'd point it out and see what other had to say about it — I can easily delete this and the corresponding tests.

jmert:
> This is what I tried to do with the `name_type_pair::Pair{String, DataType}` methods. I guess I'm not sure if it is necessary to define for `String` since this type is already special cased in the internal machinery for mapping from hdf5 to julia.

The status quo doesn't support `"name" => String`:

```julia
HDF5 data file: /tmp/test.h5

julia> datatype(hfile["test"])
HDF5 datatype: H5T_STRING {
      STRSIZE 11;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_UTF8;
      CTYPE H5T_C_S1;
   }

julia> read(hfile, "test" => String)
ERROR: Type String does not have a definite size.
Stacktrace:
 [1] sizeof
   @ ./essentials.jl:457 [inlined]
 [2] read(::HDF5Dataset, ::Type{String})
   @ HDF5 ~/.julia/dev/HDF5/src/HDF5.jl:1411
...

julia> read(hfile["test"])
"test string"
```
so that is why I'd like to add this particular specialization.

kleinhenz:
yeah that makes sense. LGTM besides the note about the check in the generic read method.

jmert:
> LGTM besides the note about the check in the generic read method

Rebased on master and with the check removed.

kleinhenz (approved):

musm (approved):

PR #688, opened 2020-10-03
Separate types required for low-level API from high-level functionality

jmert:
This PR does two major things:
1. It moves all of the types and constants required to call the low-level C functions into a new file `src/api_types.jl`.
2. A few more types are renamed to match the C names, as was done in #678.

This resolves the bulk of what I see as the outstanding reorganization aiming at separating the low-level and high-level interfaces. One of the outstanding items is the definition and use of `HDF5ReferenceObj`, which in this PR is left alone and still straddles divide. I've got a second PR I'll be putting up where I also take a stab at disentangling that, but since it's not a trivial operation, I thought I'd present them separately.


src/api_types.jl
```diff
@@ -0,0 +1,310 @@
...
+    fileno::Cuint
+    addr::haddr_t
+    otype::Cint # enum H5O_type_t
+    rc::Cuint
+    atime::C_time_t
+    mtime::C_time_t
+    ctime::C_time_t
+    btime::C_time_t
+    num_attrs::hsize_t
+    #{ inlined struct H5O_hdr_info_t named type
```
kleinhenz, line 48: I remember looking at this before and thinking that this inlining seems to directly contradict the [documentation](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/#Struct-Type-Correspondences). I know you didn't do this, but I'm just raising the point now since you added the nice comments and made me think of it again. Do we know if this actually works or if the alignment is messed up?

kleinhenz: This shouldn't be a blocker for merging or anything since this PR doesn't change anything. It just made me think of this question that I thought of but didn't have the motivation to figure out last time I saw it.

jmert: I think that should work — the structure has size and alignment equal to the members, so inlining them shouldn't change that. As a concrete check:
```julia
julia> using HDF5

julia> Int(fieldoffset(HDF5.H5O_info_t, fieldcount(HDF5.H5O_info_t)))
144
```
which can be compared against the `offsetof(H5O_info1_t, meta_size.attr);` computed in C: [godbolt](https://godbolt.org/z/jKsPfs) which calculates it as 144 as well.

jmert: Ah, just read the link you had to the docs. Maybe the saving grace here is that the sum of the inlined members' sizes correspond to an alignment offset already, so there's no padding to worry about?

jmert: I'm not opposed to making the fields a separate struct to follow the C API, though.

src/api_types.jl
```diff
@@ -23,11 +23,11 @@ end
...
 
 # For group information
-struct H5Ginfo
-    storage_type::Cint
+struct H5G_info_t
+    storage_type::Cint # enum H5G_storage_type_t
     nlinks::hsize_t
     max_corder::Int64
-    mounted::Cint
+    mounted::Cuint
```
musm, line 30: I think we should just add `const hbool_t = Cuint` and use that for the type of mounted.

src/api_types.jl
```diff
@@ -0,0 +1,307 @@
...
+    len::Csize_t
+    p::Ptr{Cvoid}
+end
+const HVL_SIZE = sizeof(hvl_t)
+
+# Object reference types
+struct HDF5ReferenceObj
+    r::UInt64 # Size must be H5R_OBJ_REF_BUF_SIZE
+end
+const HDF5ReferenceObj_NULL = HDF5ReferenceObj(UInt64(0))
```
musm, line 23: Unrelated by I think we should just get rid of this and and use 
```
HDF5ReferenceObj(0)
```
on line 
https://github.com/JuliaIO/HDF5.jl/blob/master/src/HDF5.jl#L1521

musm (approved):
LGTM! Good re-org, I just had one minor comment then I will merge after that.

src/api_types.jl
```diff
@@ -0,0 +1,310 @@
...
+struct H5_ih_info_t
+    index_size::hsize_t
+    heap_size::hsize_t
+end
+struct H5O_info_t # version 1 type H5O_info1_t
+    fileno::Cuint
+    addr::haddr_t
+    otype::Cint # enum H5O_type_t
+    rc::Cuint
+    atime::C_time_t
```
musm, line 43: `C_time_t` -> `Ctime_t` to match `Csize_t` et al.  in base julia?

src/api_types.jl
musm: nvm just saw your other PR lol

gen/api_defs.jl
```diff
@@ -155,13 +155,13 @@
 @bind h5p_get_fapl_mpio32(fapl_id::hid_t, comm::Ptr{Hmpih32}, info::Ptr{Hmpih32})::herr_t "Error getting MPIO properties"
 @bind h5p_get_fapl_mpio64(fapl_id::hid_t, comm::Ptr{Hmpih64}, info::Ptr{Hmpih64})::herr_t "Error getting MPIO properties"
 @bind h5p_get_fclose_degree(fapl_id::hid_t, fc_degree::Ref{Cint})::herr_t "Error getting close degree"
-@bind h5p_get_filter_by_id(plist_id::hid_t, filter_id::H5Z_filter_t, flags::Ref{Cuint}, cd_nelmts::Ref{Csize_t}, cd_values::Ptr{Cuint}, namelen::Csize_t, name::Ptr{UInt8}, filter_config::Ptr{Cuint})::herr_t "Error getting filter ID"
+@bind h5p_get_filter_by_id(plist_id::hid_t, filter_id::Cint, flags::Ref{Cuint}, cd_nelmts::Ref{Csize_t}, cd_values::Ptr{Cuint}, namelen::Csize_t, name::Ptr{UInt8}, filter_config::Ptr{Cuint})::herr_t "Error getting filter ID"
```
musm, line 158: Why delete the definition for `H5Z_filter_t` ?

src/api_types.jl
musm: > I'm not opposed to making the fields a separate struct to follow the C API, though.

Probably  a better idea. 

Looking at the docs, I don't think there's an alignment issue here. The result should be the same if we defined the fields in a separate (immutable) struct.  @kleinhenz was there something specific that lead you to believe there was an alignment issue to begin with?

kleinhenz: No I haven't noticed any problems, although I think this part of the code is not exercised in the tests much. I just noticed that this does what the julia documentation makes a point of saying not to do, but I am definitely not an expert on this kind of thing even in c.

musm: The documentation reads wrong to me?

This doesn't make sense to me

>  When mirroring a struct used by-value inside another struct in C, it is imperative that you do not attempt to manually copy the fields over, as this will not preserve the correct field alignment.

from the docs at https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/#Struct-Type-Correspondences

To me this reads as telling us that we should *not*  mirror  
```c
typedef struct {
    int a;
    double b;
} A;

typedef struct {
    A a;
    A b;
    int c;
} B;
```

But AFAIK it's perfectly correct to mirror this in Julia as follows:
```julia
struct A
    a::Cint
    b::Cdouble
end

struct B
    a::A
    b::A
    c::Cint
end

## or even

struct B
    a1::Cint
    b1::Cdouble
    a2::Cint
    b2::Cdouble
    c::Cint
end
```

and retain correct alignment

src/api_types.jl
jmert: :+1: I actually went back and forth a couple of times on whether to add that or not, so glad you think we should since I initially wanted to.

src/api_types.jl
jmert: Yeah, I think that'd be a good change.

gen/api_defs.jl
jmert: It struck me as out of place that this is one of the only places where a const name was given to a C typedef rather than just using the underlying C type directly (excluding the much-used `herr_t`, `hid_t`, etc.). I have no issue putting it back, especially if you think we should move in the direction of adding more type aliases (such as named constants for all of the enum types and/or just using `@enum` or `@cenum`).

src/api_types.jl
musm: For reference I've also been referred to : https://github.com/analytech-solutions/CBinding.jl   and https://github.com/JuliaLang/julia/issues/33120 regarding these issues.

musm:
LGTM, I think I will squash this if OK.

src/api_types.jl
jmert: Yeah, I came across `CBinding.jl` and `CBindingGen.jl` yesterday. My quick/inexpert thought is that it's maybe too heavy for what we want here — it generated a _large_ file of a bunch of macro definitions that took quite a while to precompile and load. Doing it manually seems to give much shorter load time.

That said, I've got a very rough use of `Clang.jl`'s generator running — not well enough that it can actually auto-generate the bindings for us, but enough that it can be a useful point of comparison (and very fast way to replace all the list of consts with enums if desired).

PR #689, opened 2020-10-03
Separate the low-level reference buffer from the high-level wrapper

jmert:
This PR is a companion and follow-up to #688 that specifically handles decoupling the high-level interface aspects of `HDF5ReferenceObj` from the requirement for the low-level API to be provided buffers of a particular size.

The diff is easier to read when only looked at on top of #688: https://github.com/jmert/HDF5.jl/compare/api_types...objref_highlow

There's a bit of code movement just to get things defined in the correct order, but the two important parts are:

1. Compared to #688, the definition of `HDF5ReferenceObj` is brought back into `src/HDF5.jl`, and it's only member changes from `UInt64` to the "opaque" `hobj_ref_t` type that corresponds to the similar type from `H5Rpublic.h`. This type then continues the role is did in the high-level constructs, being returned from reference reads and dispatched upon in other calls.

2. The low-level `h5r_*` calls are changed to take `Ptr{Cvoid}` arguments to match the C prototypes instead of `Ref{hobj_ref_t}` because (as far as I understand it...) there is a second valid opaque structure which is a difference size — `hdset_reg_ref_t` — that (I think) is the only appropriate type to be passed to calls to `h5r_get_region`. There's no high-level support for this type, and given that fact, I assume that means nobody is actually using this feature? _(I also added a commented out prototype for the new `H5R_ref_t` which unifies several reference types in the v1.12 HDF5 library and is operated upon through a new set of functions that would have to be added later.)_

3. Minor thing, but I took the opportunity to change the `h5r_dereference` call from using the underlying version 1 interface to version 2.

First, if anyone has experience with references, I'd appreciate a review of whether I've understood their use correctly. There's only a couple of tests of reference functionality in the test suite, so it's possible I've broken some assumption.

Second, the actual definitions of the opaque types are currently given as structs with a tuple member, but there's also commented-out implementations that just make use of raw primitive types. Those in some sense are kind of nice because they are _very_ opaque and have basically no support for actually operating on them by default, but the best method I could figure out for how to get a zero-initialized constant was to dig down into a use of `Core.Intrinsics.zext_int`, which I'm not keen on.


src/api_types.jl
```diff
@@ -0,0 +1,329 @@
...
+#primitive type hobj_ref_t 64 end # size is H5R_OBJ_REF_BUF_SIZE bytes
+#primitive type hdset_reg_ref_t 96 end # size is H5R_DSET_REG_REF_BUF_SIZE bytes
+#const HOBJ_REF_T_NULL = reinterpret(type64_t, UInt64(0))
+#const HDSET_REG_REF_T_NULL = Core.Intrinsics.zext_int(hdset_reg_ref_t, UInt(0))
+
+# TODO: when upgraded to using newer HDF5 v1.12 reference API, can replace both with just
+#=
+struct H5R_ref_t
+    buf::NTuple{64,UInt8} # H5R_REF_BUF_SIZE bytes
+end
```
kleinhenz, line 38: I remember looking at this a little bit before but then abandoning it because we weren't up to v1.12 yet. I remember being a little concerned about alignment issues since the c code uses a union to enforce 64 byte alignment which I wasn't sure how to mimic in julia.

```
typedef struct {
    union {
        uint8_t __data[H5R_REF_BUF_SIZE];       /* opaque data */
        int64_t align;                          /* ensures alignment */
    } u;
} H5R_ref_t;
```

Do you know if this is a real issue or does everything just work?

jmert: Good catch — I originally had it has 8 `Int64`s, but changed it later on having forgotten this point.

From the [Developer Docs](https://docs.julialang.org/en/v1/devdocs/object/#Object-allocation), you'll get pointer alignment on GC allocated resources, so it'll be correct by default on 64-bit systems, but 32-bit systems could be a problem. Also, my reading of that statement is that that only applies to GC resources, so I'm not sure[*] that carries forward to stack variables if the compiler can elide everything. To take a stab at empirically determining what happens, I've made up [this simple trial](https://gist.github.com/jmert/e60e6f0b49eed89b527a1cabddc4d23b) — I think it shows that you do end up with only 1-byte alignment if defined with `UInt8`s while `UInt64`s get you the desired 8-byte alignment.

Therefore I'll just change the struct definition to using `NTuple{8,UInt64}` instead.

_[*] (My maybe incorrect understanding of CPU architectures, though, is that x86 is fine loading unaligned 8-byte values, so optimizing at `Ref` which under GC would be pointer-aligned to an unaligned stack allocation is OK, but presumably Julia/LLVM would keep a larger alignment on architectures where alignment matters to maintain the semantics?? That doesn't change the conclusion that defining with `UInt64`s is still what you want to avoid issues on 32-bit systems ... just something I'm curious about.)_

src/api_types.jl
```diff
@@ -218,8 +238,8 @@ const H5P_LINK_ACCESS      = _read_const(:H5P_CLS_LINK_ACCESS_ID_g)
 # Reference constants
 const H5R_OBJECT         = 0
 const H5R_DATASET_REGION = 1
-const H5R_OBJ_REF_BUF_SIZE      = 8
-const H5R_DSET_REG_REF_BUF_SIZE = 12
+const H5R_OBJ_REF_BUF_SIZE      = 8  # == sizeof(hobj_ref_t)
```
musm, line 241: Am I looking at the right code
https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/src/H5Rpublic.h#29
?

musm:
I'm wondering if the existing code is properly closing objects 
https://confluence.hdfgroup.org/display/HDF5/H5R_DEREFERENCE2

> The object opened with this function should be closed when it is no longer needed so that resource leaks will not develop. Use the appropriate close function such as H5O_CLOSE or H5D_CLOSE for datasets.

src/api_types.jl
jmert: You want the v1.10 branch:

https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/src/H5Rpublic.h?at=refs%2Fheads%2Fhdf5_1_10#29-39

and

https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/src/H5Rpublic.h?at=refs%2Fheads%2Fhdf5_1_10#53-64

The comments are not a literal translation of how the C headers define the constants; they were meant to just clarify that those correspond to the type sizes.

musm: I'm a little confused on some of the changes here. 

Shouldn't `const hobj_ref_t = haddr_t`, i.e. `UInt64` but it seems like here it's instead being represented by  an 'equivalent' representation consistent of `NTuple{8,UInt8}` 

Also I'm having a hard time understanding what this means?
```
    # UInt64 and not UInt8 to get 8-byte alignment
```
Yet the comment is using the ntuple.

musm:
I'm ok merging this, but to be honest some of the logic I don't fully understand, I can tell this doesn't fundamentally change how the original code behaves, but improves upon it's organization and logic. Certainly it would be good to add some future tests, depending on how  this ends up. 

In particular, the user-facing interface for these function we may want to change to be more user-friendly ?

src/api_types.jl
jmert: The first one was just for symmetry with the other types — they're supposed to be opaque data structures, so the actual representation shouldn't matter. I could just change that back to `UInt64` if you prefer, though.

The second question just means that the element type of the tuple is `UInt64` rather than `UInt8` — [this gist](https://gist.github.com/jmert/e60e6f0b49eed89b527a1cabddc4d23b) I linked to in an earlier comment suggests LLVM doesn't maintain the full 8-byte alignment if the tuple is made of `UInt8`s while it does if the elements are `UInt64`s, so it's just to try to mirror the reason the C code uses a `union`.

jmert: I guess `UInt64` does impose an alignment, so that could be a semantic change caused by changing to a tuple of `UInt8`s.

musm: Running the code in your gist I obtain:
image

```
julia> versioninfo()
Julia Version 1.6.0-DEV.1117
Commit 36effbe10a (2020-10-02 17:38 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-10.0.1 (ORCJIT, icelake-client)
Environment:
  JULIA_EDITOR = code.cmd
  JULIA_NUM_THREADS = 8
```

musm: Oh I see, how do we know what's the correct alignment to use?

jmert: I think you just have to follow C semantics. Since that type is typedef'd to `haddr_t`, it'll have that alignment if e.g. a user were to embed it within a struct, so we should make sure to duplicate that property.

jmert:
> I'm ok merging this, but to be honest some of the logic I don't fully understand, I can tell this doesn't fundamentally change how the original code behaves, but improves upon it's organization and logic. Certainly it would be good to add some future tests, depending on how this ends up.

Admittedly I'm still getting up to speed on this package, but as far as I can tell, splitting up the definition of references into the underlying buffer type and a high-level wrapper shouldn't actually break any user of the high-level API. (The only complication is the `HDF5ReferenceObj_NULL` constant which can be deprecated but still isn't pretty — there should maybe just be an `isvalid` method added and used instead.)

> In particular, the user-facing interface for these function we may want to change to be more user-friendly ?

Yeah, I think there's incremental progress that could be made. In particular, HDF5 v1.12 appears to add a new API which adds a new type of reference and then more clearly delineates the different type of reference objects (e.g. different constructor functions for each reference type). That should provide a nicer interface to model/wrap and expose.

src/api_types.jl
musm: I see, yeah these alignment issues are interesting and I think good to resolve, but I don't think they will impact any user-facing code or internal usage, just yet.

jmert:
> I'm wondering if the existing code is properly closing objects
> https://confluence.hdfgroup.org/display/HDF5/H5R_DEREFERENCE2
> 
> > The object opened with this function should be closed when it is no longer needed so that resource leaks will not develop. Use the appropriate close function such as H5O_CLOSE or H5D_CLOSE for datasets.

I _think_ what the docs are trying to say is that dereferencing creates an HDF5 object which must be closed later, not that there are resources associated with the reference itself. The only use of `h5r_dereference` in this package is followed by a call to `h5object` which wraps the returned object in one of the high-level object types that all have associated finalizers. Therefore I think the current use is OK in terms of the high-level API.

The example linked in the docs don't have any resource closing associated with the references themselves either — just closing of the dereferenced objects.

musm:
Yeah, I think that's correct, w.r.t the resources freeing. 

In any case, the discussion regarding the alignment I don't think should have that big an affect on anything just  yet.

PR #690, opened 2020-10-06
Rename public exports to exclude HDF5 prefix

musm:


src/HDF5.jl
```diff
@@ -309,16 +309,16 @@ end
 eltype(::Type{VariableArray{T}}) where T = T
 
 # VLEN objects
-struct HDF5Vlen{T}
+struct Vlen{T}
```
jmert, line 312: Just a thought: should this be `VLen` to be closer to standard camel case?

src/HDF5.jl
```diff
@@ -158,117 +158,117 @@ function show(io::IO, fid::HDF5File)
...
-    parent::Union{HDF5File, HDF5Group, HDF5Dataset}
+mutable struct Attributes
+    parent::Union{File, Group, Dataset}
 end
-attrs(p::Union{HDF5File, HDF5Group, HDF5Dataset}) = HDF5Attributes(p)
+attrs(p::Union{File, Group, Dataset}) = Attributes(p)
 
 # Methods for reference types
-function HDF5ReferenceObj(parent::Union{HDF5File, HDF5Group, HDF5Dataset}, name::String)
+function HDF5ReferenceObj(parent::Union{File, Group, Dataset}, name::String)
```
jmert, line 271: Shouldn't `HDF5ReferenceObj` be renamed as well? The fact that v1.10 has two different types of references makes this less nice, but since we also don't support he dataset region references in the high-level API, this could just be `Reference`. Then once the library is upgraded to v1.12, the underlying storage type can be changed to `H5R_ref_t` which will support both of the existing object and dataset region reference types and then gets the new attribute reference for free.

musm: Agreed

src/HDF5.jl
musm: Agreed

src/deprecated.jl
```diff
@@ -193,3 +193,16 @@ end
 # - switch from H5Rdereference1 to H5Rdereference2
 @deprecate_binding HDF5ReferenceObj_NULL HDF5ReferenceObj(HOBJ_REF_T_NULL)
 @deprecate h5r_dereference(obj_id, ref_type, ref) h5r_dereference(obj_id, H5P_DEFAULT, ref_type, ref)
+
+### Changed in PR#689
+# - rename exported bindings
```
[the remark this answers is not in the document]
musm: Thanks

musm:
Review comments addressed.

musm:
What is the point of the `has` function that is exported...

jmert:
> What is the point of the `has` function that is exported...

It looks like really old legacy naming from very early Julia ?? — https://github.com/JuliaLang/julia/commit/3ca22bdfaf016bfbe08306cf39a681df995773d0

musm:
> It looks like really old legacy naming from very early Julia ?? — JuliaLang/julia@3ca22bd

Good digging, probably that explains this  😄

musm:
I don't think this PR is controversial. Looking at the update, the code is so much cleaner and better IMO.

musm:
I'll merge sans objections?

src/HDF5.jl
```diff
@@ -1381,8 +1380,8 @@ end
 # Create datasets and attributes with "native" types, but don't write the data.
 # The return syntax is: dset, dtype = d_create(parent, name, data; properties...)
 for (privatesym, fsym, ptype) in
-    ((:_d_create, :d_create, Union{HDF5File, HDF5Group}),
-     (:_a_create, :a_create, Union{HDF5File, HDF5Group, HDF5Dataset, HDF5Datatype}))
+    ((:_d_create, :d_create, Union{File,Group}),
```
musm, line 1383: We might want to unroll these to improve precompile speeds?

src/HDF5.jl
```diff
@@ -85,13 +84,13 @@ hdf5_type_id(::Type{Int64})      = H5T_NATIVE_INT64
 hdf5_type_id(::Type{UInt64})     = H5T_NATIVE_UINT64
 hdf5_type_id(::Type{Float32})    = H5T_NATIVE_FLOAT
 hdf5_type_id(::Type{Float64})    = H5T_NATIVE_DOUBLE
-hdf5_type_id(::Type{HDF5ReferenceObj}) = H5T_STD_REF_OBJ
+hdf5_type_id(::Type{Reference})  = H5T_STD_REF_OBJ
 
 hdf5_type_id(::Type{S}) where {S<:AbstractString}  = H5T_C_S1
 hdf5_type_id(::Type{C}) where {C<:CharType}        = H5T_C_S1
 
 const HDF5BitsKind = Union{Bool, Int8, UInt8, Int16, UInt16, Int32, UInt32, Int64, UInt64, Float32, Float64}
```
musm, line 92: These are used internally, but could also remove the `HDF5` prefix.

src/HDF5.jl
jmert: I agree, but I think it'd be OK to wait on for another PR.

jmert (approved):

test/extend_test.jl
```diff
@@ -51,8 +51,8 @@ dims, max_dims = HDF5.get_dims(b)
 @test dims == (UInt64(10000),)
 @test max_dims == (HDF5.H5S_UNLIMITED,)
 #println("b is size current $(map(int,HDF5.get_dims(b)[1])) max $(map(int,HDF5.get_dims(b)[2]))")
-# b[:] = [1:10000] # gave error no method lastindex(HDF5Dataset{PlainHDF5File},),
-# so I defined lastindex(dset::HDF5Dataset) = length(dset), and exported lastindex
+# b[:] = [1:10000] # gave error no method lastindex(HDF5.Dataset{PlainHDF5File},),
```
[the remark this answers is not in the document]
musm: No idea, I noticed this too `PlainHDF5File` doesn't exist, probably old legacy code.

src/HDF5.jl
musm: Definitely new PR but thought I'd mention it.

src/HDF5.jl
musm: There's a couple of others like this including `HDF5Scalar` that are internally used.

musm: @jmert  Should I save these changes in a new PR ?

musm: I'll do them in a separate PR since they are more internal utility constants, rather than user-facing ones, so logically it would make sense to be a separate PR.

PR #691, opened 2020-10-07
Add default null reference constructor

jmert:
I think this is a better solution for providing a null reference that can be compared against — trying to fix deprecations in JLD and MAT showed that something like this would be useful.

I know this will have to be rebased after #690 is merged.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -48,6 +48,7 @@ include("api_helpers.jl")
 struct Reference
     r::hobj_ref_t
 end
+Reference() = Reference(HOBJ_REF_T_NULL) # NULL reference to compare to
 Base.cconvert(::Type{Ptr{T}}, ref::Reference) where {T<:Union{Reference,hobj_ref_t,Cvoid}} = Ref(ref)
 
 # Single character types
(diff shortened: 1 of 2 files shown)


musm:
lgtm

musm:
This is going to conflict, do you want to me to merge this first or second?

jmert:
I can do it second. I just put up another PR which is going to cause either one of us to need to rebase, so you might as well get the larger rename done first.

src/HDF5.jl
```diff
@@ -1262,7 +1263,7 @@ unpad(s, pad::Integer) = unpad(String(s), pad)
 
 # Dereference
 function getindex(parent::Union{File, Group, Dataset}, r::Reference)
-    r.r == HOBJ_REF_T_NULL && error("Reference is null")
+    r == Reference() && error("Reference is null")
```
musm, line 1266: Is it normal for a no arg constructor to have by default a NULL reference? I suppose it makes sense for this case, since by definition if there is nothing to reference it will be a null reference.

jmert: I was thinking of it more in terms as providing a default zero-initialized data structure. It just so happens to be useful to compare against to indicate whether a reference is valid or not.

musm:
I don't see a reason against this.

PR #692, opened 2020-10-07
An assortment of minor changes

jmert:
Just a collection of small changes I've been collecting. Notable things include:

* Moving the datatype printing workaround from the low-level function to the `show` method itself. I've also added a "fast path" that avoids the hack if the datatype is recognized as valid. (Unfortunately it's not likely to be taken often since most results probably give the built-in types, but it's something.)
* I bound a few more low-level wrappers. This was motivated in part by seeing #341 just sitting there without resolution — here I've added the low-level methods that it was adding to be used.
* Using `reinterpret` within `_dataspace` actually creates a `ReinterpretArray` on modern Julia versions rather than an `Array` (but apparently that's not an issue when passed down to C). The broadcast expression is also just easier to read.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -229,10 +229,32 @@ mutable struct Datatype
     end
 end
 convert(::Type{hid_t}, dtype::Datatype) = dtype.id
-show(io::IO, dtype::Datatype) = print(io, "HDF5 datatype: ", h5lt_dtype_to_text(dtype.id))
 hash(dtype::Datatype, h::UInt) =
     (dtype.id % UInt + h) ^ (0xadaf9b66bc962084 % UInt)
 ==(dt1::Datatype, dt2::Datatype) = h5t_equal(dt1, dt2) > 0
+function show(io::IO, dtype::Datatype)
+    print(io, "HDF5 datatype: ")
+    if isvalid(dtype)
+        print(io, h5lt_dtype_to_text(dtype.id))
+    else
+        # Note that h5i_is_valid returns `false` on the built-in datatypes (e.g.
+        # H5T_NATIVE_INT), apparently because they have refcounts of 0 yet are always
+        # valid. Just temporarily turn off error printing and try the call to probe if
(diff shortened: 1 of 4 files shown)


musm:
Really nice

musm (approved):

src/HDF5.jl
```diff
@@ -283,7 +283,7 @@ end
 convert(::Type{hid_t}, attr::Attribute) = attr.id
 show(io::IO, attr::Attribute) = isvalid(attr) ? print(io, "HDF5 attribute: ", name(attr)) : print(io, "HDF5 attribute (invalid)")
 
-mutable struct Attributes
```
musm, line 264: Is this the only `mutable struct` -> `struct` audited?

jmert: I didn't check thoroughly — I noticed it while browsing through your rename PR and this one was the only one that stood out to me as not needing to be mutable since it doesn't add a finalizer.

jmert:
It looks like the only candidate, though:
```
$ git grep 'mutable struct'
src/HDF5.jl:mutable struct File <: DataFile
src/HDF5.jl:mutable struct Group <: DataFile
src/HDF5.jl:mutable struct Properties
src/HDF5.jl:mutable struct Dataset
src/HDF5.jl:mutable struct Datatype
src/HDF5.jl:mutable struct Dataspace
src/HDF5.jl:mutable struct Attribute
```

src/HDF5.jl
jmert: It looks like the only candidate, though:
```
$ git grep 'mutable struct'
src/HDF5.jl:mutable struct File <: DataFile
src/HDF5.jl:mutable struct Group <: DataFile
src/HDF5.jl:mutable struct Properties
src/HDF5.jl:mutable struct Dataset
src/HDF5.jl:mutable struct Datatype
src/HDF5.jl:mutable struct Dataspace
src/HDF5.jl:mutable struct Attribute
```

PR #694, opened 2020-10-07
Fix rename by removing has exported binding to haskey

musm:


src/HDF5.jl
```diff
@@ -922,7 +922,7 @@ end
 exists(attr::Attributes, path::String) = h5a_exists(checkvalid(attr.parent).id, path)
 exists(dset::Union{Dataset,Datatype}, path::String) = h5a_exists(checkvalid(dset).id, path)
 exists(parent::Union{File,Group}, path::String) = exists(parent, path, Properties())
-has(parent::Union{File,Group,Dataset}, path::String) = exists(parent, path)
+Base.haskey(parent::Union{File,Group,Dataset}, path::String) = exists(parent, path)
```
jmert, line 925:
```suggestion
Base.haskey(parent::Union{File,Object,Attributes}, path::String) = exists(parent, path)
```
Looks like all those types are covered by `exists`, though maybe `exists` should also be deprecated in favor of just `haskey`, the existing methods renamed, and this one removed.

src/deprecated.jl
```diff
@@ -213,3 +213,7 @@ export HDF5Attribute, HDF5File, HDF5Group, HDF5Dataset, HDF5Datatype,
 ### Changed in PR#691
 # - Make Reference() construct a null reference
 @deprecate_binding HDF5ReferenceObj_NULL HDF5.Reference()
+
+### Changed in PR#694
+export has
+@deprecate_binding has Base.haskey
```
jmert, line 219:
```suggestion
@deprecate has(parent::Union{File,Group,Dataset}, path::String) Base.haskey(parent, path)
```
Not completely sure of the semantics, but my understanding is that just deprecating the binding attaches `has` to `Base.haskey` which has a far wider type signature that we shouldn't be duplicating into `has`.

src/HDF5.jl
musm: I was wondering about that. Good to know you are also thinking the same. 

This also aligns closer with the `dict` like interface we want to transition towards.

musm: I'm going to go ahead with this bigger change, might as well have it in a single coherent PR.

src/deprecated.jl
musm: good catch

musm:
Updated.


I kinda dont' get the purpose of `https://confluence.hdfgroup.org/display/HDF5/H5A_EXISTS_BY_NAME` compared to `https://confluence.hdfgroup.org/display/HDF5/H5A_EXISTS`

src/HDF5.jl
```diff
@@ -675,17 +675,13 @@ function getindex(parent::Union{File,Group}, path::String; pv...)
...
+    if ind == 1 # matches root group
+        return "/", path[2:end]
     else
-        if off == 1
-            # Matches the root group
-            return "/", path[2:end]
-        else
-            return path[1:prevind(path, off)], path[nextind(path, off):end]
-        end
+        return path[1:ind-1], path[ind+1:end] # better to use begin:ind-1, but only avaialble on v1.5
```
jmert, line 684:
```suggestion
        indm1, indp1 = prevind(path, ind), nextind(path, ind)
        return path[1:indm1], path[indp1:end] # better to use begin:indm1, but only available on v1.5
```

test/plain.jl
```diff
@@ -629,3 +629,56 @@ close(hfile)
...
 end # show tests
+
+@testset "split1" begin
+
+@test HDF5.split1("a") == ("a", nothing)
+@test HDF5.split1("/a/b/c") == ("/", "a/b/c")
+@test HDF5.split1("a/b/c") == ("a", "b/c")
+@test HDF5.split1(GenericString("a")) == ("a", nothing)
+@test HDF5.split1(GenericString("/a/b/c")) == ("/", "a/b/c")
+@test HDF5.split1(GenericString("a/b/c")) == ("a", "b/c")
```
jmert, line 640:
```suggestion
@test HDF5.split1(GenericString("a/b/c")) == ("a", "b/c")

# The following to paths have the same graphemes but different code unit structures:
# the first one is
#     <latin small letter a with circumflex> "/" <greek small leter alpha>
# while the second one is
#     "a" <combining circumflex accent> "/" <greek small letter alpha>
circa = "â" # <latin small leter a with circumflex>
acomb = "â" # "a" + <combining circumflex accent>
path1 = circa * "/α"
path2 = acomb * "/α"
# Sanity checks that the two strings are different but equivalent under normalization
@test path1 != path2
@test Base.Unicode.normalize(path1, :NFC) == Base.Unicode.normalize(path2, :NFC)
# Check split1 operates correctly
@test HDF5.split1(path1) == (circa, "α")
@test HDF5.split1(path2) == (acomb, "α")
@test HDF5.split1("/" * path1) == ("/", path1)
@test HDF5.split1("/" * path2) == ("/", path2)
```

test/runtests.jl
```diff
@@ -4,6 +4,23 @@ using Pkg
 
 println("HDF5 version ", HDF5.h5_get_libversion())
 
+# Copied over from JuliaLang/Test
```
jmert, line 7: Why copy this? As far as I can tell, `Test` has exported `GenericString` since v0.5:
```bash
$ grep -ne 'export.*GenericString' stdlib/Test/src/Test.jl
23:export GenericString, GenericSet, GenericDict, GenericArray, GenericOrder
$ git log --oneline --no-patch -L 23,23:stdlib/Test/src/Test.jl
744ead41cb fix maximum on generic orders (#30441)
66ef347631 Base.Test: add GenericArray type and use it in test/random.jl (#22790)
c6799e6437 enable rand(::Union{AbstractSet,Associative}) (#22228)
0a61406e5a GenericString: move into Base.Test and export; rm CharString.
$ contrib/commit-name.sh 0a61406e5a
0.5.0-dev+4653
```

jmert (changes requested):

test/runtests.jl
musm: Thanks, didint realize it was exported

jmert:
This will also need some updates to the docs: https://juliaio.github.io/HDF5.jl/dev/#Getting-information

jmert:
> I kinda dont' get the purpose of `https://confluence.hdfgroup.org/display/HDF5/H5A_EXISTS_BY_NAME` compared to `https://confluence.hdfgroup.org/display/HDF5/H5A_EXISTS`

I'm guessing there's some internal library overhead which can be avoided if with the two-names version provided by `H5A_EXISTS_BY_NAME` versus the one-name `H5A_EXISTS` since it is maybe skipping some processing when you don't first have to create a valid dataset handle that can be returned to the user.

musm:
Many thanks for the review

PR #696, opened 2020-10-07
Fix straggling uses of HDF5Opaque

jmert:
I know I approved #695, but there were two straggling uses to be replaced yet — I guess that's what I get for being lazy and not checking out the PR :grimacing:

After:
```
$ git grep HDF5Opaque
src/deprecated.jl:@deprecate_binding HDF5Opaque HDF5.Opaque
```


musm:
whops

musm (approved):

PR #699, opened 2020-10-09
Rename public functions and stop exporting several public API methods

musm:
Rename public functions and stop exporting most public API

Open to re-exporting API people think should be exported.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -32,6 +32,13 @@ The mode can be any one of the following:
 |"cw"  |read-write, create file if not existing, preserve existing contents|
 |"w"   |read-write, destroying any existing contents (if any)              |
 
+For example
+```@repl main
+using HDF5
+fname = tempname(); # temporary file
+fid = h5open(fname, "w")
+```
+
 
 This produces an object of type `HDF5File`, a subtype of the abstract type `DataFile`. This file will have no elements (groups, datasets, or attributes) that are not explicitly created by the user.
 
@@ -43,25 +50,54 @@ close(fid)
 
 Closing a file also closes any other open objects (e.g., datasets, groups) in that file. In general, you need to close an HDF5 file to "release" it for use by other applications.
 
+## Writing a group or dataset
+
+Groups can be created as follows:
+```@repl main
+create_group(fid, "mygroup")
+```
+
+We can write the `"mydataset"` by:
+```@repl main
+fid["mydataset"] = rand()
+```
+Or
+```@repl main
+create_dataset(fid, "myvector", rand(10))
+```
+
+Writing to a dataset to a group is as simple as:
+```@repl main
+g = fid["mygroup"]
+g["mydataset"] = "Hello World!"
+```
+
+
+The `do` syntax is also supported, which will automatically take care of closing the file handle:
+```@repl main
+h5open("example2.h5", "w") do fid
+    create_group(fid, "mygroup")
+end
+```
 
 ## Opening and closing objects
 
-If you have a file object `fid`, and this has a group or dataset called `"myobject"` at the top level of a file, you can open it in the following way:
+If you have a file object `fid`, and this has a group or dataset called `"mygroup"` at the top level of a file, you can open it in the following way:
 
-```julia
-obj = fid["myobject"]
+```@repl main
+obj = fid["mygroup"]
 ```
 
 This does not read any data or attributes associated with the object, it's simply a handle for further manipulations. For example:
 
-```julia
+```@repl main
 g = fid["mygroup"]
 dset = g["mydataset"]
 ```
 
 or simply
 
-```julia
+```@repl main
 dset = fid["mygroup/mydataset"]
 ```
 
@@ -102,7 +138,7 @@ appending them at the end of function calls as keyword arguments.
 g["A"] = A  # basic
 g["A", chunk=(5,5)] = A # add chunks
 
-B=h5read(fn,"mygroup/B", # two parameters
+B = h5read(fn,"mygroup/B", # two parameters
   fapl_mpio=(ccomm,cinfo), # if parameter requires multiple args use tuples
   dxpl_mpio=HDF5.H5FD_MPIO_COLLECTIVE )
 ```
@@ -154,7 +190,7 @@ useful to incrementally save to very large datasets you don't want to keep in
 memory. For example,
 
 ```julia
-dset = d_create(g, "B", datatype(Float64), dataspace(1000,100,10), chunk=(100,100,1))
+dset = create_dataset(g, "B", datatype(Float64), dataspace(1000,100,10), chunk=(100,100,1))
 dset[:,1,1] = rand(1000)
 ```
 
@@ -190,7 +226,7 @@ a newly added `ismmappable` dataset can only be memory mapped after it has been 
 to.
 The following fails:
 ```julia
-vec_dset = d_create(g, "v", datatype(Float64), dataspace(10_000,1))
+vec_dset = create_dataset(g, "v", datatype(Float64), dataspace(10_000,1))
 ismmappable(vec_dset)    # == true
 vec = readmmap(vec_dset) # throws ErrorException("Error mmapping array")
 ```
@@ -202,10 +238,10 @@ vec_dset[1,1] = 0.0      # force allocation of /g/v within the file
 vec = readmmap(vec_dset) # and now the memory mapping can succeed
 ```
 
-Alternatlively, the policy can be set so that the space is allocated immediately upon
+Alternatively, the policy can be set so that the space is allocated immediately upon
 creation of the data set with the `alloc_time` keyword:
 ```julia
-mtx_dset = d_create(g, "M", datatype(Float64), dataspace(100, 1000),
+mtx_dset = create_dataset(g, "M", datatype(Float64), dataspace(100, 1000),
                     alloc_time = HDF5.H5D_ALLOC_TIME_EARLY)
 mtx = readmmap(mtx_dset) # succeeds immediately
 ```
@@ -215,7 +251,7 @@ mtx = readmmap(mtx_dset) # succeeds immediately
 `HDF5.jl` knows how to store values of the following types: signed and unsigned integers of 8, 16, 32, and 64 bits, `Float32`, `Float64`; `Complex` versions of these numeric types; `Array`s of these numeric types (including complex versions); `ASCIIString` and `UTF8String`; and `Array`s of these two string types.
 `Array`s of strings are supported using HDF5's variable-length-strings facility.
 By default `Complex` numbers are stored as compound types with `r` and `i` fields following the `h5py` convention.
-When reading data, compound types with matching field names will be loaded as the corresponding `Complex` julia type.
+When reading data, compound types with matching field names will be loaded as the corresponding `Complex` Julia type.
 These field names are configurable with the `HDF5.set_complex_field_names(real::AbstractString, imag::AbstractString)` function and complex support can be completely enabled/disabled with `HDF5.enable/disable_complex_support()`.
 
 For `Array`s, note that the array dimensionality is preserved, including 0-length
@@ -249,7 +285,7 @@ This module also supports HDF5's VLEN, OPAQUE, and REFERENCE types, which can be
 Create a new group in the following way:
 
 ```julia
-g = g_create(parent, name)
+g = create_group(parent, name)
 ```
 
 The named group will be created as a child of the parent.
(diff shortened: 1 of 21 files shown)


jmert:
Should the modifying functions terminate with `!` to better follow typical Julia naming?

musm:
A lot of these feel like we should just use multiple dispatch for, e.g.:
```
create_* could all just be
create(HDF5.group, ....) et al. 
```
although the dispatch is messy because the function signatures vary quite a bit between them all.

musm:
> Should the modifying functions terminate with ! to better follow typical Julia naming? IO functions don't, but Dict functions do, so not sure which should be followed.

That's a good point, I don't have a very strong opinion on this either, but it might make sense to adopt the Dict like notion  for some of the functions. In particular `delete`, which AFAIK  doesn't have a IO function equivalent.

musm:
> although the dispatch is messy because the function signatures vary quite a bit between them all.

Eek, this in general might just be too confusing.

musm:
Note so I don't forget:  figure out what to do with `name`, `filename`

musm:
I do think we should probably not export anything and make the public interface  module scoped.

musm:
If we are renaming these we should really decide what to unexport, otherwise it's going to be basically impossible to unexport the names if we want to do so in a deprecation friendly manner.

musm:
One option is to basically unexport everything other than h5* functions.  Then if we decide to export module scoped public API functions, it will at least be possible to do so.

musm:
Another question, which form should we adopt:
(1) `group_write` or (2) `write_group` 

(2)reads more naturally and is the scheme adopted by, e.g. h5py, and more consistent in general with Julia. OTOH, (1) allows you to hit `<Tab>`  on `group_` and easily discover all available methods.

src/show.jl
```diff
@@ -118,44 +118,44 @@ _tree_icon(obj::Attributes) = _tree_icon(obj.parent)
 _tree_head(io::IO, obj) = print(io, _tree_icon(obj), " ", obj)
 _tree_head(io::IO, obj::Datatype) = print(io, _tree_icon(obj), " HDF5.Datatype: ", name(obj))
 
-function _tree_children(parent::Union{File, Group}, attributes::Bool)
+function _tree_children(parent::Union{File, Group}, show_attributes::Bool)
```
musm, line 121: @jmert  attributes was clashing with the `attributes` method so I had to rename it.

src/HDF5.jl
```diff
@@ -2,34 +2,39 @@ module HDF5
...
+using Compat
 
 import Libdl
 import Mmap
 
-# needed for filter(f, tuple) in julia 1.3
-using Compat
-
 export
 @read, @write,
```
musm, line 12: We should change this to `@load` and `@save` to align with JLD and JLD2.

musm: Probably follow up work.

src/HDF5.jl
```diff
@@ -2,34 +2,39 @@ module HDF5
...
+# create_attribute, open_attribute. delete_attribute, read_attribute, write_attribute, attributes,
+# create_dataset, create_external_dataset, open_dataset, read_dataset, write_dataset,
+# create_group, open_group,
+# copy_object, open_object, delete_object,
+# create_datatype, commit_datatype,
+# create_property
+# group_info, object_info
+
+# file, filename, name,
+# dataspace, datatype,
```
musm, line 27: I'm open to exports lines 18-27. Although, personally I feel like we should keep the list of exports minimal.

musm:
This should be ready for review now.

src/deprecated.jl
```diff
@@ -57,19 +57,19 @@ function h5read(filename, name::AbstractString, indices::Tuple{Vararg{Union{Abst
...
 function d_create(parent::Union{File,Group}, path::AbstractString, dtype::Type, dspace, prop1::AbstractString, val1, pv...)
-    depwarn("d_create with string key and value argument pairs is deprecated. Use keywords instead.", :d_create)
+    depwarn("d_create with string key and value argument pairs is deprecated. Use create_dataset with keywords instead.", :create_dataset)
     props = (prop1, val1, pv...)
-    return d_create(parent, path, dtype, dspace; [Symbol(props[i]) => props[i+1] for i in 1:2:length(props)]...)
+    return create_dataset(parent, path, dtype, dspace; [Symbol(props[i]) => props[i+1] for i in 1:2:length(props)]...)
 end
 function p_create(class, prop1::AbstractString, val1, pv...)
-    depwarn("p_create with string key and value argument pairs is deprecated. Use keywords instead.", :p_create)
+    depwarn("create_property with string key and value argument pairs is deprecated. Use create_property keywords instead.", :create_property)
```
jmert, line 70:
```suggestion
    depwarn("p_create with string key and value argument pairs is deprecated. Use create_property keywords instead.", :create_property)
```

docs/src/index.md
```diff
@@ -336,15 +372,15 @@ attr = a_open(parent, name)
 t = t_open(parent, name)
 ```
 
-These open the named group, dataset, attribute, and committed datatype, respectively. For datasets, `apl` stands for "access parameter list" and provides opportunities for more sophisticated control (see the [HDF5](https://www.hdfgroup.org/solutions/hdf5/) documentation).
+These open the named group, dataset, attribute, and committed HDF5.datatype, respectively. For datasets, `apl` stands for "access parameter list" and provides opportunities for more sophisticated control (see the [HDF5](https://www.hdfgroup.org/solutions/hdf5/) documentation).
```
jmert, line 375:
```suggestion
These open the named group, dataset, attribute, and committed datatype, respectively. For datasets, `apl` stands for "access parameter list" and provides opportunities for more sophisticated control (see the [HDF5](https://www.hdfgroup.org/solutions/hdf5/) documentation).
```

Extra find/replace?

src/HDF5.jl
```diff
@@ -2,34 +2,39 @@ module HDF5
...
-# Note: Public API requires module scoping
-### Public Functions:
-# g_info, o_info
-### Types:
+h5open, h5read, h5write, h5rewrite, h5writeattr, h5readattr
+
+### PUBLIC API ### (requires module scoping)
+
+### Public Functions
+# create_attribute, open_attribute. delete_attribute, read_attribute, write_attribute, attributes,
```
jmert, line 18:
```suggestion
# create_attribute, open_attribute, delete_attribute, read_attribute, write_attribute, attributes,
```

src/HDF5.jl
```diff
@@ -2,34 +2,39 @@ module HDF5
...
+# create_dataset, create_external_dataset, open_dataset, read_dataset, write_dataset,
+# create_group, open_group,
+# copy_object, open_object, delete_object,
+# create_datatype, commit_datatype,
```
jmert, line 22:
```suggestion
# create_datatype, commit_datatype, open_datatype
```
?

src/HDF5.jl
```diff
@@ -645,10 +650,10 @@ fd(obj::Object) = h5i_get_file_id(checkvalid(obj).id)
 Base.flush(f::Union{Object,Attribute,Datatype,File}, scope = H5F_SCOPE_GLOBAL) = h5f_flush(checkvalid(f).id, scope)
 
 # Open objects
-g_open(parent::Union{File,Group}, name::AbstractString, apl::Properties=DEFAULT_PROPERTIES) = Group(h5g_open(checkvalid(parent), name, apl), file(parent))
-d_open(parent::Union{File,Group}, name::AbstractString, apl::Properties=DEFAULT_PROPERTIES, xpl::Properties=DEFAULT_PROPERTIES) = Dataset(h5d_open(checkvalid(parent), name, apl), file(parent), xpl)
+open_group(parent::Union{File,Group}, name::AbstractString, apl::Properties=DEFAULT_PROPERTIES) = Group(h5g_open(checkvalid(parent), name, apl), file(parent))
+open_dataset(parent::Union{File,Group}, name::AbstractString, apl::Properties=DEFAULT_PROPERTIES, xpl::Properties=DEFAULT_PROPERTIES) = Dataset(h5d_open(checkvalid(parent), name, apl), file(parent), xpl)
 t_open(parent::Union{File,Group}, name::AbstractString, apl::Properties=DEFAULT_PROPERTIES) = Datatype(h5t_open(checkvalid(parent), name, apl), file(parent))
```
jmert, line 655:
```suggestion
open_datatype(parent::Union{File,Group}, name::AbstractString, apl::Properties=DEFAULT_PROPERTIES) = Datatype(h5t_open(checkvalid(parent), name, apl), file(parent))
```
?

docs/src/index.md
```diff
@@ -336,15 +372,15 @@ attr = a_open(parent, name)
 t = t_open(parent, name)
```
jmert, line 372:
```suggestion
t = open_datatype(parent, name)
```
(and similar for previous three lines. For some reason I can't get Github to let me add a suggestion for those lines...)

docs/src/index.md
```diff
@@ -336,15 +372,15 @@ attr = a_open(parent, name)
...
 
 New objects can be created in the following ways:
 ```julia
 g = g_create(parent, name[, lcpl, dcpl])
-dset = d_create(parent, name, data[, lcpl, dcpl, dapl])
+dset = HDF5.create_dataset(parent, name, data[, lcpl, dcpl, dapl])
 attr = a_create(parent, name, data)
 ```
-creates groups, datasets, and attributes without writing any data to them. You can then use `write(obj, data)` to store the data. The optional property lists allow even more fine-grained control. This syntax uses `data` to infer the object's "datatype" and "dataspace"; for the most explicit control, `data` can be replaced with `dtype, dspace`, where `dtype` is an `HDF5Datatype` and `dspace` is an `HDF5Dataspace`.
+creates groups, datasets, and attributes without writing any data to them. You can then use `write(obj, data)` to store the data. The optional property lists allow even more fine-grained control. This syntax uses `data` to infer the object's "HDF5.datatype" and "HDF5.dataspace"; for the most explicit control, `data` can be repl mainaced with `dtype, dspace`, where `dtype` is an `HDF5Datatype` and `dspace` is an `HDF5Dataspace`.
```
jmert, line 383:
```suggestion
creates groups, datasets, and attributes without writing any data to them. You can then use `write(obj, data)` to store the data. The optional property lists allow even more fine-grained control. This syntax uses `data` to infer the object's "HDF5.datatype" and "HDF5.dataspace"; for the most explicit control, `data` can be replaced with `dtype, dspace`, where `dtype` is an `HDF5Datatype` and `dspace` is an `HDF5Dataspace`.
```

jmert (changes requested):

src/HDF5.jl
jmert: I think I'd tend toward exporting those. A lot of the high-level interface is already accessible with only `import HDF5` because it's implemented as overloads on many `Base` functions, so `using` vs `import` alone is already pretty good at cleaning up a namespace if necessary. I think the `using` case can afford to have a few more exports.

jmert:
> In a follow up I propose we remove/refactor `h5rewrite, h5writeattr, h5readattr`

:+1:

musm:
Thanks for the review. Addressed the comments and fixed up the remaining issues. 

Do you want to sign off on the updated export list?

jmert:
> (Should we export ishdf5?)

I'm agnostic — I think there's no harm in exporting it, but I haven't actually personally found it necessary either. (I've been using `FileIO.query` on files and dispatching on `FileIO.File{format"HDF5"}`.)

jmert:
Just as a note, I found I temporarily needed:
```diff
diff --git a/src/deprecated.jl b/src/deprecated.jl
index 04e204a..fa138f7 100644
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -412,18 +412,18 @@ end
 @deprecate objinfo(obj::Union{File,Object}) object_info(obj)
 
 # - rename bindings
-@deprecate a_open    open_attribute
-@deprecate a_read    read_attribute
-@deprecate a_write   write_attribute
-@deprecate a_create  create_attribute
-@deprecate a_delete  delete_attribute
-@deprecate attrs     attributes
+@deprecate a_open                    open_attribute
+@deprecate a_read                    read_attribute
+@deprecate a_write(args...; kws...)  write_attribute(args...; kws...)
+@deprecate a_create(args...; kws...) create_attribute(args...; kws...)
+@deprecate a_delete                  delete_attribute
+@deprecate attrs                     attributes
 
-@deprecate d_open            open_dataset
-@deprecate d_read            read_dataset
-@deprecate d_write           write_dataset
-@deprecate d_create          create_dataset
-@deprecate d_create_external HDF5.create_external_dataset
+@deprecate d_open                    open_dataset
+@deprecate d_read                    read_dataset
+@deprecate d_write(args...; kws...)  write_dataset(args...; kws...)
+@deprecate d_create(args...; kws...) create_dataset(args...; kws...)
+@deprecate d_create_external         HDF5.create_external_dataset
 
 @deprecate g_open   open_group
 @deprecate g_create create_group
@@ -432,7 +432,7 @@ end
 @deprecate o_copy   copy_object
 @deprecate o_delete delete_object
 
-@deprecate p_create create_property
+@deprecate p_create(args...; kws...) create_property(args...; kws...)
 
 @deprecate t_create create_datatype
 @deprecate t_open   open_datatype
```
to work through MAT.jl's update to this PR because it was already updated to property syntax and the bare `@deprecate` doesn't add keywords automatically. I _don't_ necessarily think this should be added, though, because it's not relevant for upgrading from the tagged v0.13.6 — rather just wanted to point it out as an interesting consequence of multiple rounds of deprecated changes.

jmert (approved):

musm:
Ok great, thanks again!

PR #703, opened 2020-10-12
RFC: Initialize runtime globals during __init__

jmert:
This isn't necessarily a real proposal — I wanted to explore how far we could get with actually doing runtime initialization of the in-principle dynamic global "constants" from the HDF5 library while simultaneously not adding a burden on users to know whether a value is a true constant or a runtime-constant.

As it turns out, I was able to hide the difference from users by using (abusing?) `getproperty` defined on structs. The basic idea is:

* An overload of `getproperty` can evaluate an expression for each use of `name.symbol`, so a large `if-elseif` block can be used to hide whether a value is a true constant (and therefore return `symbolname`) or a dynamic runtime value stored in a `Ref` (and therefore return `symbolname[]`).
* `getproperty` cannot be defined on a module, so I've "abused" a singleton `struct` for method dispatch. The runtime values are
actually stored in a [lightly] name-mangled bare module.
  - Storing the values within a module and not just reusing the existing top-level names means deprecation bindings can be appropriately set since the representation of using the constants directly does have to change for the `Ref` containers.
* Likewise, overloading `setindex!` allows easily initializing the value during `__init__` no matter the internal implementation.

Pros:
* This is the "correct" thing to do given runtime-constant values.
* It puts something like defining a HDF5 data type for `Float16` on similar footings to all the built-in data types — the datatype is constructed at initialization and just fills in its `Ref` container. No need for runtime `@eval` which is otherwise necessary to avoid requiring users to do the Ref-indexing.
  - Note that the last commit is insufficient for `Float16` to actually work — there's some changes to uses of `h5t_get_native_type` in #341 that I haven't replicated here, and not including those changes means the generic `read()` doesn't work due to size mismatch on the actual Float16 datatype (2 bytes) versus it's native-ified datatype (4 bytes). I've just included the last commit to show that adding a new "constant" is actually very easy with this scheme.
* The constants are namespaced, so they can actually be more conveniently access by just exporting the top-level "module" names. An overload of `propertynames` also means you get very convenient tab completion as well.

Cons:
* This adds a non-trivial macro to the precompile path. The macro definitely makes defining this "too clever" hack easier, but it must surely add to the precompile time.
* Likewise, runtime initializing all of the `Ref` containers at `__init__` must add a bit of load time on normal uses as well.
* I've been playing around with this in Julia v1.6 master where Julia seems to constant-propagate the `getproperty`s to just the relevant expression, but I haven't checked prior versions for whether this is actually adding a large runtime if-else chain to every use.
* It's in general maybe just too clever for its own good :-P


musm:
Nice, I haven't fully looked at the PR in details, but I think we should go for something like this. 

I actually prototyped a similar idea before hand over at https://github.com/musm/HDF5.jl/tree/refs, however I used constant global refs, which where initialized at runtime. I never fully pushed it through, but this PR got be interested to compare how the different approaches effect package precompilation and load times. 

Here are the timings I get (using your gist for timings)
```
master precompile: 2.271 ± 0.0576
master pkg load: 0.597 ± 0.0195

# https://github.com/musm/HDF5.jl/tree/refs
refs precompile: 2.313 ± 0.0670
refs pkg load: 0.626 ± 0.0389

# this pr
init precompile: 3.116 ± 0.0706
init pkg load: 0.688 ± 0.0316
```

I'm not too concerned with the precompile times, since the PR here does more codegen . Comparing the precompile times, it seems that the hit is really not too bad.  Using const refs, does seem to offer improved precompilation times, which I am assuming come from the fact that they are global constants and that this helps compilation.  In summary, the pkg-load time hit using this PR is 15%, and using constant refs 5%.  The constant ref approach, does make using the constants less ergonomic, which is why I think I held off from it, since I couldn't think of a clean interface for the constants. 


I think in principle it should be possible to segfault master HDF5, by opening multiple HDF5.jl instances from different Julia processes, which would demonstrate why we need such a PR that makes these constants runtime initialized. Although, I haven't been able to successfully to do so.

src/api_types.jl
```diff
@@ -104,222 +104,316 @@ abstract  type Hmpih end
...
+    H5T.NATIVE_INT32    = read_const(:H5T_NATIVE_INT32_g)
+    H5T.NATIVE_UINT32   = read_const(:H5T_NATIVE_UINT32_g)
+    H5T.NATIVE_INT64    = read_const(:H5T_NATIVE_INT64_g)
+    H5T.NATIVE_UINT64   = read_const(:H5T_NATIVE_UINT64_g)
+    H5T.NATIVE_FLOAT    = read_const(:H5T_NATIVE_FLOAT_g)
+    H5T.NATIVE_DOUBLE   = read_const(:H5T_NATIVE_DOUBLE_g)
+
+    nothing
+end
+__init_globals__()
```
musm, line 419:
```suggestion
 ```
Typo here?

Comment (jmert) [reply src/api_types.jl:419]: Not a typo exactly — just left-over cruft. At some point while rewriting, it was necessary explicitly initialize everything while the module was being defined, but that's no longer the case.

**Comment by jmert** [comment]
I just added a new commit which brings down both the precompile and load time just a little bit. The issue was that I was defining the `getproperty`/`setproperty!` overloads on the singleton types, but that leads to a huge number of invalidations in `Base`. By instead defining a singleton instances of the type and then defining the functions on the instance, the invalidations are eliminated, which gives a small speed-up.

The SnoopCompile setup:
```julia
julia> using SnoopCompileCore

julia> invs = @snoopr using HDF5;

julia> using SnoopCompile
```

**Before:** *(commit b6ab18b using singleton types)*
```julia
julia> trees = invalidation_trees(invs)
1-element Vector{SnoopCompile.MethodInvalidations}:
 inserting getproperty(::Type{H5E}, sym::Symbol) in HDF5 invalidated:
   backedges: 1: superseding getproperty(x::Type, f::Symbol) in Base at Base.jl:28 with MethodInstance for getproperty(::DataType, ::Symbol) (1220 children)
   3 mt_cache
```

**After:** *(commit f09463a using singleton instances)*
```
julia> trees = invalidation_trees(invs)
SnoopCompile.MethodInvalidations[]
```

And timing precompile and load from master and the before/after singleton commits:
```
 master precompile: 2.085 ± 0.0499
 master pkg load: 0.594 ± 0.0299

# using singleton types
runtime precompile: 2.807 ± 0.0309
runtime pkg load: 0.650 ± 0.0287

# using singleton instances
runtime precompile: 2.542 ± 0.0290
runtime pkg load: 0.637 ± 0.0226
```

src/macros.jl
```diff
@@ -0,0 +1,100 @@
...
+    pushfirst!(defbody, Expr(:import, imports...))
+
+    prefix   = esc(prefix)
+    innermod = esc(innermod)
+    block = quote
+        Base.@__doc__(struct $prefix end)
+        baremodule $innermod
+            $(defbody...)
+        end
+        function Base.propertynames(::Type{$prefix})
```
musm: In this commit how would this internal module know about the outside struct definition?

musm: Oh nvm I see how this sets things up now.  It creates a struct, then  a baremodule with only the constants and then defines  `get,set,properties` in the main module scope.

musm:
So in the new change `H5P` is just an alias for the singleton type `_H5P.H5P()` ?

jmert:
> So in the new change H5P is just an alias for the singleton type _H5P.H5P() ?

Yes.

src/macros.jl
```diff
@@ -0,0 +1,120 @@
...
+
+        isruntime = isexpr(line, :(=)) ? false :
+                    isexpr(line, :(::)) ? true :
+                    error("Unexpected statement: ", repr(line))
+
+        # Get the name and type pair
+        nametype = isruntime ? line : line.args[1]
+        isexpr(nametype, :(::)) || error("Expected `name::type`, got ", repr(nametype))
+        name = nametype.args[1]::Symbol
+        type = nametype.args[2]
```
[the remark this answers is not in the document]
jmert: As used now, it could be, but it'd preclude the possibility of setting up any non-trivial types. For example, `:(Ptr{Int})` is an `Expr(:curly, :Ptr, :Int)`.

src/macros.jl
```diff
@@ -0,0 +1,120 @@
...
+assigned at runtime from an `__init__()` function:
+```julia
+function __init__()
+    H5PREFIX.DYANMIC_CONSTANT = my_runtime_value()
+end
+```
+"""
+macro defconstants(prefix::Symbol, expr::Expr)
+    isexpr(expr, :block) || error("Expected block expression")
+    stmts = expr.args
```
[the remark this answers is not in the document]
jmert: Possibly — I don't know which is faster to execute. I was guessing the `isa`-`continue` might be since it wouldn't have to modify any arrays.

musm:
Yeah I think this is pretty cool.  In general,  I'm comfortable with pushing through with this, because in principle it is the  correct thing to do (runtime initialization) and the ergonomics of using these types is enhanced, albeit behind some chicanery, which I don't think is all that different from, say factorization using `(get/set)property` in `LinearAlgebra`. The only  'mischievous' thing here is that we are overloading a singleton instance.   Could we piggyback off of this and use these instances for functions as well? I.e. `h5r_create` -> `H5R.create`? It feels like since we are going this route perhaps we should just commit to it for the interal api as well. (we could probably piggy back of the `genbinding` macro.  

I wonder if we can block redefinitions of truly constant values or error them (without a runtime or compile time cost), because right now for non-refs things like `H5S.runtimeconstant = 3` won't error , but is allowed, which probably should error.

src/macros.jl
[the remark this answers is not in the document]
jmert: `::Union{Symbol,Expr}` would match what the `imports` array that it's pushed into is, though.

src/HDF5.jl
```diff
@@ -1928,17 +1933,26 @@ function __init__()
         ENV["HDF5_USE_FILE_LOCKING"] = "FALSE"
     end
 
+    __init_globals__()
     register_blosc()
 
+    # Generate the Float16 datatype:
```
musm, line 1939: I would move this to `__init_globals()` and add some comments that this is added to support IEEE julia compatible Float16 values. We could also think about adding types of other types of Float16 dataypes that the Julia ecosystem supports.

src/api_types.jl
```diff
@@ -253,8 +253,8 @@ end
 # Dataspace constants
 @defconstants H5S begin
     # atomic data types
-    ALL::hid_t       = 0
-    UNLIMITED::hid_t = typemax(hsize_t)
+    ALL::hsize_t       = 0
+    UNLIMITED::hsize_t = typemax(hsize_t)
```
musm, line 257: nice catch

src/HDF5.jl
jmert: This has to happen after the `include("api.jl")` line which happens after this file.

src/api_types.jl
jmert: It's actually a bit ambiguous if you read `H5Spublic.h`:
```
#define H5S_ALL         (hid_t)0
#define H5S_UNLIMITED   HSIZE_UNDEF
```
As far as I can tell, though, `hsize_t` makes as much or more sense, though.

src/HDF5.jl
musm: Perhaps then we can move the to a function elsewhere or in this file and then call that here. Something like `_init_custom_globals__()` ?

jmert: Sure, I'd be OK with that.

jmert:
>  Could we piggyback off of this and use these instances for functions as well? I.e. h5r_create -> H5R.create? It feels like since we are going this route perhaps we should just commit to it for the interal api as well. (we could probably piggy back of the genbinding macro.

It's definitely possible, but I'm a little worried that at some point the long chain of `if-elseif` statements are going to not optimize well and performance will suffer. It's something that could be looked at, though. (Maybe a two-level `getproperty` chain would resolve most of that, though — keep constants defined explicitly in `getproperty` and chain-call to a hidden module and let normal name resolution take place??)

> I wonder if we can block redefinitions of truly constant values or error them (without a runtime or compile time cost), because right now for non-refs things like H5S.runtimeconstant = 3 won't error , but is allowed, which probably should error.
Ah, yes, I should definitely add an error path to both `getproperty` and `setproperty!`. They used to have one by default by way of chain-calling the type's `getfield`/`setfield!`, but since those went away after switching to using an instance, I should add back explicit errors.

src/HDF5.jl
jmert: Oh wait, that was probably again caused by the temporary need to run `__init_globals__()` immediately during development. It should probably now work that actually running the function doesn't happen until `__init__`.

musm:
> It's definitely possible, but I'm a little worried that at some point the long chain of if-elseif statements are going to not optimize well and performance will suffer. It's something that could be looked at, though. (Maybe a two-level getproperty chain would resolve most of that, though — keep constants defined explicitly in getproperty and chain-call to a hidden module and let normal name resolution take place??)

I haven't thought about the implementation in to great detail, but that sounds like it could work. I guess that means we would have to hide all of the lower-level function in the same module.


Should we also fully commit to the hack and define `show` methods :trollface: ? 
```
Base.show(io::IO, ::typeof(H5)) = print("H5")
```

src/api_types.jl
musm: I guess you're right, seems like it was changed from 1.18  https://github.com/HDFGroup/hdf5/blob/1.8/master/src/H5Spublic.h#L24 
https://github.com/HDFGroup/hdf5/blob/1.10/master/src/H5Spublic.h#L25

jmert: We don't have to — what I meant by ambiguous was that I couldn't quickly/easily tell (from headers nor documentation) whether `H5S_ALL` and `H5S_UNLIMITED` are supposed to be essentially two possible values for the same type of thing or if the former actually has multiple meanings. I'm fine with just going with the literal translation of the header, though.

jmert:
> Should we also fully commit to the hack and define `show` methods :trollface: ?
> 
> ```
> Base.show(io::IO, ::typeof(H5)) = print("H5")
> ```

I'd been wondering the same thing.

musm:
I guess we should make a decision on the type of 'public' interface we want to go ahead with in the future. 

If we commit to this style of using 'dot' to prefix the various HDF5 modules, instead of our current convention of using underscore, then this change is very desirable. If not, this change is still desirable in the sense that it improves initialization of global and makes them much more ergonomic to use instead of trying to figure out which is a runtime global and having to use `x[]` vs `x` to reference these constants, and we can tell people to use `_` for functions, which admittedly feels less nice, if the global are using `.` to prefix them. 

> I'd been wondering the same thing.

I'm kind of interested to see if any other packages have utilized  similar in sprit, 'hacks' (that feels a bit too strong a word  here, but I think you get my point)

src/macros.jl
```diff
@@ -0,0 +1,125 @@
...
+        function Base.getproperty(::$einnermod.$prefix, sym::Symbol)
+            $(getbody...)
+            error($(string(prefix) * " has no constant "), sym)
+        end
+    end
+    if !isempty(setbody)
+        setfn = quote
+            function Base.setproperty!(::$einnermod.$prefix, sym::Symbol, value)
+                $(setbody...)
+                error($(string(prefix) * "."), sym, " cannot be set")
```
musm, line 116: We should probably change the error message to be consistent with what is printed in Base. Either:

Assigning a variable to in an other module:
"ERROR: cannot assign variables in other modules"

Trying to redefine a constant:
"ERROR: cannot declare x constant; it already has a value"

src/macros.jl
```diff
@@ -0,0 +1,125 @@
...
+                    $fullname[] = value
+                    return value
+                end
+            end
+            append!(setbody, Base.remove_linenums!(setexpr).args)
+        end
+    end
+
+    # Build expressions to import all necessary types from the parent module.
+    push!(imports, :Ref) # explicitly add `Ref`
```
musm, line 91: `Ref` should be defined in `Core` so I don't think this is needed for the `baremodule`.

jmert: :+1: I hadn't realized it was part of `Core`.

src/macros.jl
jmert: I like the second one because it's still technically correct.

PR #706, opened 2020-10-13
Fix #705, distinguishing 0-size and null arrays

### LINKED ISSUE #705: Inconsistent writing and loading of 0x0 matrix
```julia-repl
julia> using HDF5

julia> h5write("_test2.h5", "arr", zeros(0,0))

julia> h5read("_test2.h5", "arr")
Float64[]

julia> zeros(0,0)
0×0 Array{Float64,2}
```

jmert:
HDF5 has the concept of NULL arrays which are distinct from 0-size arrays. As shown in #705, writing any 0-size array was being translated to the NULL object during writing, but having 0-length axes is also perfectly valid. This PR preserves 0-sizes when writing arrays.

To represent the HDF5 NULL object, there was already the `HDF5.EmptyArray` type defined, but it appears to have been used only for internal dispatch on the generic read (which then returns a 0-length vector for NULL objects).

This PR reuses `HDF5.EmptyArray` and elevates it to a full subtype of `AbstractArray`, which then becomes the returned object for any NULL datasets/attributes which are read in. This array type carries an element type but has no contents. This is distinct from Julia's `Array{T,0}` 0-dimensional arrays because the latter _does_ have a single element. (I tried just doing duck typing instead of actually making it a subtype of `AbstractArray`, but you end up playing whack-a-mole with enough new method definitions that you might as well just implement the `AbstractArray` interface and take advantage of its fallbacks once the few methods are defined.)

Surprisingly, both MAT and JLD don't require any changes to handle this PR (beyond the deprecation-fixing private branches I've been maintaining...), so it appears this case is generally unused — only a couple of test cases in this package required fixes.

(One oddity is that Julia's 0-dimensional `Arrays` are turned into HDF5 scalars and will not be read back in as a 0-dimensional array, but (a) that is the existing behavior, and (b) I haven't thought of any other way to handle the case.)

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -317,6 +317,10 @@ Base.length(::EmptyArray) = 0
 # Required to avoid indexing during printing
 Base.show(io::IO, E::EmptyArray) = print(io, typeof(E), "()")
 Base.show(io::IO, ::MIME"text/plain", E::EmptyArray) = show(io, E)
+# FIXME: Concatenation doesn't work for this type (it's treated as a length-1 array like
+# Base's 0-dimensional arrays), so just forceably abort.
+Base.cat_size(::EmptyArray) = error("concatenation of HDF5.EmptyArray is unsupported")
+Base.cat_size(::EmptyArray, d) = error("concatenation of HDF5.EmptyArray is unsupported")
 
 # Stub types to encode fixed-size arrays for H5T_ARRAY
 struct FixedArray{T,D,L}
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -302,7 +302,21 @@ struct Opaque
 end
 
 # An empty array type
-struct EmptyArray{T} end
+struct EmptyArray{T} <: AbstractArray{T,0} end
+# Required AbstractArray interface
+Base.size(::EmptyArray) = ()
+Base.IndexStyle(::Type{<:EmptyArray}) = IndexLinear()
+Base.getindex(::EmptyArray, ::Int) = error("cannot index an `EmptyArray`")
+Base.setindex!(::EmptyArray, v, ::Int) = error("cannot assign to an `EmptyArray`")
+# Optional interface
+Base.similar(::EmptyArray{T}) where {T} = EmptyArray{T}()
+Base.similar(::EmptyArray, ::Type{S}) where {S} = EmptyArray{S}()
+Base.similar(::EmptyArray, ::Type{S}, dims::Dims) where {S} = Array{S}(undef, dims)
+# Override behavior for 0-dimensional Array
+Base.length(::EmptyArray) = 0
+# Required to avoid indexing during printing
+Base.show(io::IO, E::EmptyArray) = print(io, typeof(E), "()")
+Base.show(io::IO, ::MIME"text/plain", E::EmptyArray) = show(io, E)
 
 # Stub types to encode fixed-size arrays for H5T_ARRAY
 struct FixedArray{T,D,L}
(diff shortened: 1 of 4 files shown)


src/HDF5.jl
```diff
@@ -1040,34 +1054,26 @@ dataspace(attr::Attribute) = Dataspace(h5a_get_space(checkvalid(attr).id))
...
-dataspace(A::AbstractArray; max_dims::Union{Dims,Tuple{}} = ()) = _dataspace(size(A), max_dims)
-dataspace(str::String) = Dataspace(h5s_create(H5S_SCALAR))
+dataspace(A::AbstractArray{T,N}; max_dims::Union{Dims{N},Tuple{}} = ()) where {T,N} = _dataspace(size(A), max_dims)
+# special array types
 dataspace(v::VLen; max_dims::Union{Dims,Tuple{}}=()) = _dataspace(size(v.data), max_dims)
+dataspace(A::EmptyArray) = Dataspace(h5s_create(H5S_NULL))
 dataspace(n::Nothing) = Dataspace(h5s_create(H5S_NULL))
-dataspace(sz::Dims; max_dims::Union{Dims,Tuple{}}=()) = _dataspace(sz, max_dims)
+# for giving sizes explicitly
+dataspace(sz::Dims{N}; max_dims::Union{Dims{N},Tuple{}}=()) where {N} = _dataspace(sz, max_dims)
```
musm, line 1076: Is this and the parametric restriction above done for clarity/increased specificity (I don't think this will improve performance)?

musm (approved):

src/HDF5.jl
```diff
@@ -302,7 +302,21 @@ struct Opaque
...
 # An empty array type
-struct EmptyArray{T} end
+struct EmptyArray{T} <: AbstractArray{T,0} end
+# Required AbstractArray interface
+Base.size(::EmptyArray) = ()
+Base.IndexStyle(::EmptyArray) = IndexLinear()
+Base.getindex(::EmptyArray, ::Int) = error("cannot index an `EmptyArray`")
+Base.setindex!(::EmptyArray, v, ::Int) = error("cannot assign to an `EmptyArray`")
+# Optional interface
+Base.similar(::EmptyArray{T}) where {T} = EmptyArray{T}()
```
musm, line 312: Any specific reason to include these optional interface methods? I guess it doesn't hurt in case someone want's to use them, albeit the `EmptyArray` type is quite limited in utility.

musm:
Would be good to include documentation on this, either here or in another PR? Documentation could use a bigger overhaul (longer term goal)

src/HDF5.jl
jmert: Yes, just for exactness, with no intention that it actually makes this more performant. The only valid options are for default/empty `max_dims` or for `length(sz) == length(max_dims)`, so we might as well enforce that in the type signature.

src/HDF5.jl
jmert: I included them because some recent work I've been doing quite often relies on getting `similar` arrays, so it was on my mind as something useful. For instance, I could imagine making a type-compatible working buffer based on a loaded dataset, even if the "loaded" data is the empty set.

jmert:
Yeah, documentation is something I've had on my mind but haven't actually decided to take the plunge on starting. I can try to get some details on this added into this PR, though.

musm:
LGTM, feel free to merge

src/HDF5.jl
```diff
@@ -302,7 +302,21 @@ struct Opaque
...
+Base.similar(::EmptyArray, ::Type{S}) where {S} = EmptyArray{S}()
+Base.similar(::EmptyArray, ::Type{S}, dims::Dims) where {S} = Array{S}(undef, dims)
+# Override behavior for 0-dimensional Array
+Base.length(::EmptyArray) = 0
+# Required to avoid indexing during printing
+Base.show(io::IO, E::EmptyArray) = print(io, typeof(E), "()")
+Base.show(io::IO, ::MIME"text/plain", E::EmptyArray) = show(io, E)
```
musm, line 319: Isn't this already the default? 

From docs:

>   The default MIME type is MIME"text/plain". There is a fallback definition for text/plain output that calls show with 2 arguments, so it is not always
  necessary to add a method for that case.

jmert: There's a default implementation of 3-arg show for `<:AbstractArray`, so that has to be overridden — https://github.com/JuliaLang/julia/blob/44d44708f0f0f2456c623bcd999a89adc35fe8c8/base/arrayshow.jl#L323

musm: Oh right that makes sense. I guess we are still missing something w.r.t to printing

The matrix case ends up here printing junk values
image

jmert: The last one is `Matrix{Int32}`, so something weird went on with `hvcat`, I think?

musm: Hmm, interesting, didn't even realize the type got changes.

jmert: I think this is because Base's 0-dimensional arrays have shape `()` but have a size of `1` in any dimension when being concatenated with other arrays. We actually want an empty shape to have zero size, so there's a conflict with Base's assumption, and it's "overallocating" the output array.

This seems to be buried pretty deep in Base, so I'm not sure if there's a good fix on this end — various hacks I've tried end up exposing different unwanted behavior. The best option might be to just forceably error out within the concatenation path? I think an overload of `Base.cat_size(::HDF5.EmptyArray) = error("...unsupported...")` would do it.

musm: Hmm that seems annoying. I guess we can either just leave it or add the  overload you suggest iable option (although this might end up being wack-a-mole, if we discover other similar methods.).

jmert: The other option is to drop the `<: AbstractArray` inheritance, which then makes the concatenation framework treat the type as a scalar. The problem there is that again you have to play whack-a-mole defining a bunch of other methods which are assumed for any array-like type (and a bunch of methods — `write`, `d_write`, etc) have to have their types modified to accept a non-`AbstractArray` type explicitly... No good options.

musm: Yeah that sounds like the worse option.

PR #707, opened 2020-10-15
Explicitly define (d|a)_(write|create)

musm:
The only issue is I couldn't figure out the deprecations. @jmert any help here? 

I think not using the macro's for these make debugging etc far easier and the code simpler.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -91,7 +91,6 @@ hdf5_type_id(::Type{S}) where {S<:AbstractString} = H5T_C_S1
 
 const BitsType = Union{Bool,Int8,UInt8,Int16,UInt16,Int32,UInt32,Int64,UInt64,Float32,Float64}
 const ScalarType = Union{BitsType,Reference}
-const ScalarOrComplex = Union{ScalarType}
 
 # It's not safe to use particular id codes because these can change, so we use characteristics of the type.
 function _hdf5_type_map(class_id, is_signed, native_size)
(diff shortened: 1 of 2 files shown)


src/HDF5.jl
```diff
@@ -91,6 +91,7 @@ hdf5_type_id(::Type{S}) where {S<:AbstractString} = H5T_C_S1
 
 const BitsType = Union{Bool,Int8,UInt8,Int16,UInt16,Int32,UInt32,Int64,UInt64,Float32,Float64}
 const ScalarType = Union{BitsType,Reference}
+const ScalarOrComplex = Union{ScalarType}
```
musm, line 94:
```suggestion
```

Typo,

unrelated, but I think we have an error  with Complex in general. Right now we have `Complex{<:ScalarTypes}` everywhere, which includes `Reference` which I think needs to be disallowed. In, which case we could add `Complex{<:Real}` to `BitsType` ?

jmert:
> The only issue is I couldn't figure out the deprecations. @jmert any help here?

Sure, I'll take a look.

musm:
Many thanks, feel free to push to this branch

jmert:
I'm not sure what you're asking for help with — I saw the include of `deprecations.jl` was commented out, so I'd assumed there was just an implementation problem after this change, but when I uncommented it, I have no problems `using HDF5`.

What specifically were you asking for help with? Just un-`@eval`-ifying the similar definitions in `deprecations.jl`?

musm:
If you run the test suite, you'll run into ambiguity problems
image


I've had a hard time tracking how to resolve the conflicts...

jmert:
I think the solution was pretty simple — all the extra type annotations on `data` can just go away because the internal dispatches to `datatype` and `dataspace` (and later on, the actual underlying `writearray`/`h5d_write`) all have type restrictions which will error out for unsupported types.

Tests passed locally, but I wouldn't be surprised if there's an untested corner case that needs to be worried about. In particular, I'd like to consolidate all of the definitions into one place to make auditing the type overlaps a bit easier. (This could be more relevant if/when the `h5d_write` methods and similar are lifted to the mid-level API `d_write` methods...)

src/HDF5.jl
```diff
@@ -1427,7 +1421,7 @@ function d_write(parent::Union{File,Group}, name::String, data::Union{T,Abstract
     end
     nothing
 end
-function a_write(parent::Union{File,Object,Datatype}, name::String, data::Union{T,AbstractArray{T},VLen{S}}; pv...) where {T<:Union{ScalarType,String,Complex{<:ScalarType}},S<:Union{ScalarType,CharType}}
```
musm: Not sure why `Datatype` was included

musm:
Fantastic, that does make a whole bunch more sense... 👍

 

>  In particular, I'd like to consolidate all of the definitions into one place to make auditing the type overlaps a bit easier. (This could be more relevant if/when the h5d_write type-specialized methods and similar are lifted to be part of the mid-level API d_write methods...)

That sounds like a good idea. Do you want me to merge it as is and keep the following plans in another PR?

jmert:
> That sounds like a good idea. Do you want me to merge it as is and keep the following plans in another PR?

Yes, that can follow later. It's been on my mental TODO list for a while, but other items keep coming up first (and all presumably help make that step later finally easier :-) )

jmert:
Just FYI: This branch passed both of the MAT.jl and JLD.jl test suites on their own branches where I've been fixing them up for HDF5 v0.14 compatibility. Since especially JLD has a much larger test suite, that bodes well for my simplification not being a big problem.

musm:
Fantastic. Yes, looking back it probably never made much sense to restrict the function signatures for the high level interface  anyways, after all this would also restrict users implementing custom datatypes.

PR #709, opened 2020-10-16
Auto-generated list of library bindings for docs

jmert:
Very first step in updating docs — let's just auto-generate documentation listing the auto-generated bindings!

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -14,6 +14,7 @@ makedocs(;
     ),
     pages=[
         "Home" => "index.md",
+        "Low-level library bindings" => "api_bindings.md",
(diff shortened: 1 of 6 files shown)


musm (commented):
LGTM!

musm:
This looks great. I didn't have time to check, but I wonder if Documenter has a feature to automatically document all methods in a file.

jmert:
I had that thought, but I hadn't yet looked into it — I wanted this summary to contain the nominal C argument types, but the definitions in `src/api.jl` are all untyped function declarations. Therefore, just augmenting the macro generator was the better option.

PR #711, opened 2020-10-19
Use explicit extension of Base functions instead of manual imports

musm:
Major change:
1. Take advantage of defaults methods in `t_commit`

Unrelated changes:
1. Explicitly define method overloads (we should stick to this convention throughout the module, instead of imports as it can quickly be easy to forget where the methods are originally defined).
2. Misc whitespace fixups.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -330,7 +330,7 @@ struct FixedString{N,PAD}
     data::NTuple{N,UInt8}
 end
 Base.length(::Type{FixedString{N,PAD}}) where {N,PAD} = N
-Base.length(::FixedString) where {T<:FixedString} = length(T)
+Base.length(str::FixedString) = length(typeof(str))
 pad(::Type{FixedString{N,PAD}}) where {N, PAD} = PAD
 pad(x::T) where {T<:FixedString} = pad(T)
 
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -3,11 +3,6 @@ module HDF5
 using Base: unsafe_convert, StringVector
 using Requires: @require
 
-import Base:
-    close, convert, eltype, lastindex, flush, getindex, ==,
-    isempty, isvalid, length, ndims, parent, read,
-    setindex!, show, size, sizeof, write, isopen, iterate, eachindex, axes
-
 import Libdl
 import Mmap
 
@@ -58,12 +53,12 @@ abstract type CharType <: AbstractString end
 struct ASCIIChar <: CharType
     c::UInt8
 end
-length(c::ASCIIChar) = 1
+Base.length(c::ASCIIChar) = 1
 
 struct UTF8Char <: CharType
     c::UInt8
 end
-length(c::UTF8Char) = 1
+Base.length(c::UTF8Char) = 1
 
 chartype(::Type{String}) = ASCIIChar
 stringtype(::Type{ASCIIChar}) = String
@@ -146,8 +141,8 @@ mutable struct File <: DataFile
         f
     end
 end
-convert(::Type{hid_t}, f::File) = f.id
-function show(io::IO, fid::File)
+Base.convert(::Type{hid_t}, f::File) = f.id
+function Base.show(io::IO, fid::File)
     if isvalid(fid)
         print(io, "HDF5 data file: ", fid.filename)
     else
@@ -166,8 +161,8 @@ mutable struct Group <: DataFile
         g
     end
 end
-convert(::Type{hid_t}, g::Group) = g.id
-function show(io::IO, g::Group)
+Base.convert(::Type{hid_t}, g::Group) = g.id
+function Base.show(io::IO, g::Group)
     if isvalid(g)
         print(io, "HDF5 group: ", name(g), " (file: ", g.file.filename, ")")
     else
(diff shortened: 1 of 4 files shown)


src/HDF5.jl
```diff
@@ -327,15 +327,15 @@ struct FixedArray{T,D,L}
...
 eltype(::Type{FixedArray{T,D,L}}) where {T,D,L} = T
-eltype(x::T) where {T<:FixedArray} = eltype(T)
+eltype(x::FixedArray) = eltype(typeof(x))
 
 struct FixedString{N,PAD}
     data::NTuple{N,UInt8}
 end
 length(::Type{FixedString{N,PAD}}) where {N,PAD} = N
-length(x::T) where {T<:FixedString} = length(T)
+length(::FixedString) where {T<:FixedString} = length(T)
```
jmert, line 338:
```suggestion
length(str::FixedString) = length(typeof(str))
```

Apparently needs a test as well:
```julia
julia> fix = HDF5.FixedString{4,0}((b"test"...,))
HDF5.FixedString{4, 0}((0x74, 0x65, 0x73, 0x74))

julia> length(fix)
ERROR: UndefVarError: T not defined
Stacktrace:
 [1] length(#unused#::HDF5.FixedString{4, 0})
   @ HDF5 ~/.julia/dev/HDF5/src/HDF5.jl:338
 [2] top-level scope
   @ REPL[8]:1

julia> length(typeof(fix))
4
```

src/HDF5.jl
```diff
@@ -146,8 +146,8 @@ mutable struct File <: DataFile
         f
     end
 end
-convert(::Type{hid_t}, f::File) = f.id
-function show(io::IO, fid::File)
+Base.convert(::Type{hid_t}, f::File) = f.id
```
jmert, line 149: If all unqualified uses have been replaced with this and similar functions, the `import Base:` entries should be removed.

jmert (changes requested):

src/HDF5.jl
```diff
@@ -738,25 +738,12 @@ d_create(parent::Union{File,Group}, path::String, dtype::Type, dspace_dims; pv..
 
 # Note that H5Tcreate is very different; H5Tcommit is the analog of these others
 t_create(class_id, sz) = Datatype(h5t_create(class_id, sz))
-function t_commit(parent::Union{File,Group}, path::String, dtype::Datatype, lcpl::Properties, tcpl::Properties, tapl::Properties)
-    h5p_set_char_encoding(lcpl.id, cset(typeof(path)))
-    h5t_commit(checkvalid(parent).id, path, dtype.id, lcpl.id, tcpl.id, tapl.id)
+function t_commit(parent::Union{File,Group}, path::AbstractString, dtype::Datatype, lcpl::Properties=p_create(H5P_LINK_CREATE), tcpl::Properties=DEFAULT_PROPERTIES, tapl::Properties=DEFAULT_PROPERTIES)
```
musm, line 741: I'm going to split this change into the other PR, leave this PR to clean up imports.

src/HDF5.jl
musm: yikes, good catch, I swear I noticed this like two weeks ago and totally forgot about it.

src/HDF5.jl
musm: Thanks, will update the PR.

musm:
```julia
> git grep '::String' -- src/HDF5.jl  | wc -l

src/HDF5.jl:    filename::String
src/HDF5.jl:    tag::String
src/HDF5.jl:function h5write(filename, name::String, data; pv...)
src/HDF5.jl:function h5read(filename, name::String; pv...)
src/HDF5.jl:function h5read(filename, name::String, indices::Tuple{Vararg{Union{AbstractRange{Int},Int,Colon}}}; pv...)
src/HDF5.jl:function h5writeattr(filename, name::String, data::Dict)
src/HDF5.jl:function h5readattr(filename, name::String)
src/HDF5.jl:#getindex(parent::Union{HDF5File, HDF5Group}, path::String) = o_open(parent, path)
src/HDF5.jl:function setindex!(parent::Union{File,Group}, val, path::String; pv...)
src/HDF5.jl:function datatype(str::String)
src/HDF5.jl:dataspace(str::String) = Dataspace(h5s_create(H5S_SCALAR))
src/HDF5.jl:function unpad(s::String, pad::Integer)
src/HDF5.jl:read(attr::Attributes, name::String) = a_read(attr.parent, name)
src/HDF5.jl:function d_create_external(parent::Union{File,Group}, name::String, filepath::String, t, sz::Dims, offset::Integer)
src/HDF5.jl:d_create_external(parent::Union{File,Group}, name::String, filepath::String, t::Type, sz::Dims) = d_create_external(parent, name, filepath, t, sz, 0)
src/HDF5.jl:h5a_write(attr_id::hid_t, mem_type_id::hid_t, buf::String) = h5a_write(attr_id, mem_type_id, unsafe_wrap(Vector{UInt8}, pointer(buf), ncodeunits(buf)))
src/HDF5.jl:function h5d_write(dataset_id::hid_t, memtype_id::hid_t, str::String, xfer::hid_t=H5P_DEFAULT)

```

Down a lot more now. I only modified the deprecation file in cases where it lead to an ambiguity error.

musm:
All fixed up, ready for a final review.

jmert (approved):
Interestingly, this broke my JLD.jl update branch because it's been importing HDF5.jl's imports of Base. Getting rid of these explicit imports broke the chain to the `Base` methods. (Obviously not this package's problem and will be fixed at the source — I'm adding a new commit to my WIP branch to do just that.)

musm:
Sounds good and thanks for the review. Yeah, these rewrites are uncovering subtle bugs or incorrect usages, which is a good thing albeit annoying.

PR #712, opened 2020-10-22
Fixup cleanup function for a_write

musm:
Found this while working on another PR. Main issue is that 
`o_delete(obj)` when `obj` is attribute  doesn't work. 

Tests are not going to pass until the AbstractString clean up passes.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1756,7 +1756,13 @@ end
 ### Convenience wrappers ###
 # These supply default values where possible
 # See also the "special handling" section below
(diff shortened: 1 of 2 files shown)


jmert:
I hope you don't mind, but I rebased on master and added a commit that resolves https://github.com/JuliaIO/HDF5.jl/pull/710#discussion_r510272530.

src/HDF5.jl
```diff
@@ -1424,7 +1424,7 @@ function a_write(parent::Union{File,Object}, name::AbstractString, data; pv...)
     try
         writearray(obj, dtype.id, data)
     catch exc
-        o_delete(obj)
```
musm, line 1427: Thanks, but did it looks like you accidentally delete this? This is still problematic and `o_delete` won't work when `obj` is an Attribute.

musm: ops my bad looking at the wrong thing

musm:
Looks good and thanks!

PR #714, opened 2020-10-27
Cleanup id usage plus misc. house cleaning

musm:


src/HDF5.jl
```diff
@@ -1761,49 +1761,49 @@ end
...
         h5a_write(attr_id, mem_type_id, buf)
     end
 end
-function h5a_write(attr_id::hid_t, mem_type_id::hid_t, x::T) where {T<:Union{ScalarType,Complex{<:ScalarType}}}
+function h5a_write(attr_id, mem_type_id, x::T) where {T<:Union{ScalarType,Complex{<:ScalarType}}}
     tmp = Ref{T}(x)
     h5a_write(attr_id, mem_type_id, tmp)
 end
-function h5a_write(attr_id::hid_t, memtype_id::hid_t, strs::Array{S}) where {S<:String}
+function h5a_write(attr_id, memtype_id, strs::Array{S}) where {S<:String}
```
musm, line 1775: @jmert  I'm not sure if we need to fix this as well to accept Arrays of abstract strings.

jmert: It looks like changing it would work with `GenericString`, so I'd vote for yes.:
```julia
julia> using HDF5, Test

julia> hfile = h5open("/tmp/test.h5", "w");

julia> genstrs = GenericString["one", "purple"];

julia> a, dtype = a_create(hfile, "myattr", genstrs)
(HDF5 attribute: myattr, HDF5 datatype: H5T_STRING {
      STRSIZE H5T_VARIABLE;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_UTF8;
      CTYPE H5T_C_S1;
   })

julia> HDF5.h5a_write(a, dtype, Ref{Cstring}(genstrs))

julia> flush(hfile)

shell> h5dump -p /tmp/test.h5
HDF5 "/tmp/test.h5" {
GROUP "/" {
   ATTRIBUTE "myattr" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATA {
      (0): "one", "purple"
      }
   }
}
}
```

(I didn't locally modify this branch, so that's the reason for the manual call to `Ref{Cstring}(genstrs)`.)

src/HDF5.jl
```diff
@@ -901,7 +901,7 @@ function Base.setindex!(parent::Union{File,Group}, val, path::AbstractString; pv
 end
 
 # Check existence
-function Base.haskey(parent::Union{File,Group}, path::AbstractString, lapl::Properties=Properties())
```
musm, line 904: Right now we have
```julia
    function Properties(id, class::hid_t = H5P_DEFAULT)
        p = new(id, class)
        finalizer(close, p) #Essential, otherwise we get a memory leak, since closing file with CLOSE_STRONG is not doing it for us
        p
    end
 Properties() = Properties(H5P_DEFAULT)

const DEFAULT_PROPERTIES = Properties(H5P_DEFAULT, H5P_DEFAULT)

```

Should we change this to 
```julia
    function Properties(id::hid_t = H5P_DEFAULT, class::hid_t = H5P_DEFAULT)
        p = new(id, class)
        finalizer(close, p) #Essential, otherwise we get a memory leak, since closing file with CLOSE_STRONG is not doing it for us
        p
    end


```
I.e. just use the default constructor directly? I don't think it matters either way, just a design/consistency decision.

musm:
I think this is good to go on my end.

jmert (approved):

PR #717, opened 2020-10-28
Add support for dimension scales API bindings

### LINKED ISSUE #716: Add support for the HDF5 dimension scale functionality
As the title says. I get a lot of use from the HDF5 library's high level dimension scales functions (H5DS). I've added the functionality that should enable this and done some basic testing that shows that it seems to work (at least for my use case), so I'll be submitting a pull request for  this shortly.

fergu:
This adds bindings to the HDF5 dimension scales API functions (H5DS)

All of the H5DS functions are added, except for `H5DS_ITERATE_SCALES` which I could not figure out how to bind to Julia, and it seemed of dubious use anyway.

`bind_generator.jl` was changed to allow for the H5DS functions to be correctly linked to the high level library instead of the default library.

`gen_wrappers.jl` is changed to add references to the API on the HDF Group's website.

This will close #716

### DIFF
--- a/gen/api_defs.jl
+++ b/gen/api_defs.jl
@@ -259,6 +259,21 @@
 @bind h5do_write_chunk(dset_id::hid_t, dxpl_id::hid_t, filter_mask::Int32, offset::Ptr{hsize_t}, bufsize::Csize_t, buf::Ptr{Cvoid})::herr_t "Error writing chunk"
 
 ###
+### High Level Dimension Scale Interface
+###
+
+@bind h5ds_attach_scale(did::hid_t, dsid::hid_t, idx::Cuint)::herr_t "Unable to attach scale"
+@bind h5ds_detach_scale(did::hid_t, dsid::hid_t, idx::Cuint)::herr_t "Unable to detach scale"
+@bind h5ds_get_label(did::hid_t, idx::Cuint, label::Ptr{UInt8}, size::hsize_t)::herr_t "Unable to get label"
+@bind h5ds_get_num_scales(did::hid_t, idx::Cuint)::Cint "Error getting number of scales"
+@bind h5ds_get_scale_name(did::hid_t, name::Ptr{UInt8}, size::Csize_t)::Cssize_t "Unable to get scale name"
+@bind h5ds_is_attached(did::hid_t, dsid::hid_t, idx::Cuint)::htri_t "Unable to check if dimension is attached"
+@bind h5ds_is_scale(did::hid_t)::htri_t "Unable to check if dataset is scale"
+@bind h5ds_set_label(did::hid_t, idx::Cuint, label::Ref{UInt8})::herr_t "Unable to set label"
+@bind h5ds_set_scale(dsid::hid_t, dimname::Ptr{UInt8})::herr_t "Unable to set scale"
+
+
+###
 ### HDF5 Lite Interface
 ###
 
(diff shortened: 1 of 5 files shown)


gen/bind_generator.jl
```diff
@@ -115,7 +115,7 @@ macro bind(sig::Expr, err::Union{String,Expr,Nothing} = nothing)
     push!(funclist, string(funcsig))
 
     # Determine the underlying C library to call
-    lib = startswith(string(cfuncname), r"H5(DO|LT|TB)") ? :libhdf5_hl : :libhdf5
+    lib = startswith(string(cfuncname), r"H5(DO|LT|TB|DS)") ? :libhdf5_hl : :libhdf5
```
jmert, line 118:
```suggestion
    lib = startswith(string(cfuncname), r"H5(DO|DS|LT|TB)") ? :libhdf5_hl : :libhdf5
```

Keep in alphabetical order.

gen/api_defs.jl
```diff
@@ -275,3 +275,17 @@
 ###
 
 @bind h5z_register(filter_class::Ref{H5Z_class_t})::herr_t "Unable to register new filter"
+
```
jmert, line 278: Move up to below the H5DO section to keep alphabetical ordering.

jmert (commented):
Generally looks OK to me.

Having a high-level interface is _not_ a prerequisite, but out of curiosity, do you have some sort of wrapper or helper functions in mind? (I'm not at all familiar with what the H5DS module does, so just a quick glance at the docs isn't enough to have a suggestion.)

gen/gen_wrappers.jl
```diff
@@ -46,6 +46,7 @@ for (mod, desc, urltail) in (
     ("H5DO", "Optimized Functions Interface", "Optimizations"),
     ("H5LT", "Lite Interface", "Lite"),
     ("H5TB", "Table Interface", "Tables"),
+    ("H5DS", "Dimension Scale Interface", "Dimension+Scales"),
```
jmert, line 49: Move up two lines to keep alphabetical within the HL section.

jmert:
Oh, missed the obvious thing — you have to actually run `gen/gen_wrappers.jl` to get the bindings inserted.

fergu:
> Having a high-level interface is not a prerequisite, but out of curiosity, do you have some sort of wrapper or helper functions in mind?

I mainly wanted to add the ability to just use these functions from within Julia. The H5DS API functions, as far as I can tell, just wrap some of the other lower-level calls to create dimension scales according to their spec, linked [here](https://support.hdfgroup.org/HDF5/doc/HL/H5DS_Spec.pdf).

Basically all a dimension scale is is a dataset that you designate as a scale rather than data, and then attach that scale to a given dimension of other datasets. The H5DS API calls then give it a few special designations (listed in the spec) to link the scales and datasets/dimensions together.

As far as I know, a dimension scale doesn't change anything about the associated data sets within the file directly. It's more of a way to say "This dataset represents the coordinates of dimension N of dataset(s) A, B, and C. Or, if you look at dataset A (or B, or C), you'll see that the scale associated with dimension N is this dataset". Actually using those dimension scales is up to the person using the file - this is just a standard way of designating that some datasets represent axis coordinates rather than actual data, and a way for someone viewing the file to see that there is a relationship between them.

I do agree that it might be useful to add wrappers around these calls since they do have a lot of use for scientific datasets (Which is my use case)

musm:
@fergu  many thanks for the PR

Also feel free if you would like to add some example usage to the docs or some tests (as follow up)? It sounds like you have a good use-case for these functions and I think it might be helpful to demonstrate them for others.  This might also lead to a high-level interface, if need be.

PR #718, opened 2020-10-29
Fix a few ambiguities

jmert:
The first commit doesn't actually fix any ambiguities on master, but as I'm playing with rewriting `h5d_write` methods to `d_write`, I think these signatures are going to be relevant.

The other two commits' problems were identified by running
```julia
using HDF5, Test
Test.detect_ambiguities(HDF5)
```
I've added tests for the ambiguity in the `getindex(::Dataset, ::Reference)` method since that was obviously not being tested, but I haven't for the `d_create` ambiguity since I'm thinking the method should just be deprecated. It's not tested here, nor does it get used in MAT.jl or JLD.jl since none of those test suites have run into the ambiguity either.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1274,11 +1274,13 @@ end
 unpad(s, pad::Integer) = unpad(String(s), pad)
 
 # Dereference
-function Base.getindex(parent::Union{File,Group,Dataset}, r::Reference)
+function _deref(parent, r::Reference)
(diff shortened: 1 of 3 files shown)


musm:
LGTM

jmert:
This passed locally on branches of MAT and JLD which have been updated for all of the breaking changes here, so I'm merging this.

PR #719, opened 2020-10-29
Replace `vlenpack` with `cconvert`

jmert:
It's sufficient to have the implicit `cconvert` call at `ccall` time trigger the packing procedure. This is a step toward removing of the `h5d_write` specializations.

As I understand it, at the moment the type restrictions on the `VLen` constructors restrict the possible datasets to a much narrower range than the HDF5 library supports, but that should be OK since a wider type can always be added in the future.

(I think the narrow type range actually means GC handling is OK right now; in principle there should be calls to `cconvert` before the `unsafe_convert` to pointers, but skipping it for the small range of types allowed doesn't have any consequence. For more generic types, a `cconvert` is required, but then there's potentially the issue of rooting the returned objects through the `ccall`.)

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -349,22 +349,20 @@ Base.eltype(::Type{VariableArray{T}}) where T = T
 
 # VLEN objects
 struct VLen{T}
-    data
+    data::Array
 end
 VLen(strs::Array{S}) where {S<:String} = VLen{chartype(S)}(strs)
 VLen(A::Array{Array{T}}) where {T<:ScalarType} = VLen{T}(A)
 VLen(A::Array{Array{T,N}}) where {T<:ScalarType,N} = VLen{T}(A)
-
-t2p(::Type{T}) where {T<:ScalarType} = Ptr{T}
-t2p(::Type{C}) where {C<:CharType} = Ptr{UInt8}
-function vlenpack(v::VLen{T}) where {T<:Union{ScalarType,CharType}}
+function Base.cconvert(::Type{Ptr{Cvoid}}, v::VLen)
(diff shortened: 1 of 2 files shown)


musm:
lgtm

musm (approved):

musm:
Sorry about the conflict!

jmert:
> Sorry about the conflict!

Not actually a problem — I'd already had to fix the conflict when I made the merged base for #723, and thanks to enabling `git rerere` before doing that, the conflict was already resolved for me:
```bash
$ git rebase master
Auto-merging test/plain.jl
Auto-merging src/HDF5.jl
CONFLICT (content): Merge conflict in src/HDF5.jl
error: could not apply 1f7291b... Replace vlenpack with cconvert
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
Resolved 'src/HDF5.jl' using previous resolution.                                  # ← Notice here
Could not apply 1f7291b... Replace vlenpack with cconvert
$ git add src/HDF5.jl
$ git rebase --continue
[detached HEAD 4deda16] Replace vlenpack with cconvert
 2 files changed, 19 insertions(+), 17 deletions(-)
Successfully rebased and updated refs/heads/cconvert_vlen.
```
(This is basically the first time I've actually used the feature, so I don't understand why it didn't auto-continue since there wasn't actually anything left to do... :man_shrugging: but whatever.)

musm:
Wow that's pretty cool.

PR #722, opened 2020-10-29
Have deprecated `exists` export itself

### LINKED ISSUE #721: Removing exists() for Base.haskey() is a breaking change
Stumbled upon this since I switched to using the master branch over the version in the Julia repos to use the changes I submitted yesterday.

the `exists()` function was removed in PR #694. The reasoning makes perfect sense since it aligns better with the functionality one would expect from other Julia libraries. However, given that `exists()` has been the documented way to check if an object/group exists for a while now, would it be better to just pass `exists()` on to `haskey()`, perhaps with a printed warning that it's been deprecated and may be totally removed at some point, rather than jumping straight to removing it?

jmert:
#694 deprecated `exists`, but that PR left it within the explicit export list so the deprecation binding still worked with the `false` final argument. Then in #701 it was removed from the main export list, so the deprecation itself needs to do the [implicit] export for the deprecation to be useful to downstream users.

Fixes #721.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -221,4 +221,4 @@ import Base: names
 
 ### Changed in PR#694
 @deprecate has(parent::Union{File,Group,Dataset}, path::AbstractString) Base.haskey(parent, path)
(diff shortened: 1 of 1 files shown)


musm:
Many thanks!

Issue #666, opened 2020-09-16
v0.14 release tracker


musm:
~~I propose tagging a major release by the end of the week sans objections?~~

Checklist before release:
- [x] Decide what to un-export.
- [x] https://github.com/JuliaIO/HDF5.jl/pull/723
- [x] https://github.com/JuliaIO/HDF5.jl/pull/724
- [x] https://github.com/JuliaIO/HDF5.jl/pull/725 (pretty printing) (optional?)
- [x] https://github.com/JuliaIO/HDF5.jl/pull/699 (rename functions)

Save for next major release (note this can come soon after if we are ready)?
- [ ] https://github.com/JuliaIO/HDF5.jl/pull/703

musm:
After tagging, we can then fix JLD and MAT?

kleinhenz:
Do JLD and MAT have compat bounds so that our release won't break them?

musm:
Yes they do

jmert:
As the new issue #669 I just added hints to, I think there are still a few corner cases in the new `read` interface that need to be sorted out. In addition to that issue which I think is a true regression, the behavior surrounding fixed-length string arrays is maybe now a bit confused — 

writing on v0.13.6 — while not truly round-trippable — were at least similar in form for read and write:
```julia
julia> h5open("test.h5", "w") do hfile
           attrs(hfile)["attr"] = HDF5.HDF5Vlen(["string", "another"])
       end
HDF5Vlen{HDF5.ASCIIChar}(["string", "another"])

julia> h5open("test.h5", "r") do hfile
           read(attrs(hfile)["attr"])
       end
2-element Vector{String}:
 "string"
 "another"
```

but now on master, the form being read back is very different from that being written:
```julia
julia> h5open("test.h5", "w") do hfile
           attrs(hfile)["attr"] = HDF5.HDF5Vlen(["string", "another"])
       end
HDF5Vlen{HDF5.ASCIIChar}(["string", "another"])

julia> h5open("test.h5", "r") do hfile
           read(attrs(hfile)["attr"])
       end
2-element Vector{Vector{String}}:
 ["s", "t", "r", "i", "n", "g"]
 ["a", "n", "o", "t", "h", "e", "r"]
```

I haven't dug far enough yet to know the answer for sure, but I think this is due to special cases for `HDF5Vlen` strings that construct the "strange" length-1 string arrays that are plaguing MAT.jl right now on HDF5 master.

[...]

PR #723, opened 2020-10-29
Move low-level method specializations to the middle-level API

jmert:
This implements the API cleanup I advocated for in #666, moving some of the current method overloads on what should be the low-level API up to the middle-level API layer.

Marked as WIP because:
- [x] I've built this on top of merge commits of both #714 and #719. Those should be finished and merged first.
- [x] I haven't yet sorted out how to add all the necessary method deprecations.
- [x] There's one remaining `h5d_read` to be ported.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -711,8 +711,8 @@ end
 
 function g_create(parent::Union{File,Group}, path::AbstractString,
                   lcpl::Properties=_link_properties(path),
-                  dcpl::Properties=DEFAULT_PROPERTIES)
-    Group(h5g_create(checkvalid(parent), path, lcpl, dcpl, H5P_DEFAULT), file(parent))
+                  gcpl::Properties=DEFAULT_PROPERTIES)
+    Group(h5g_create(checkvalid(parent), path, lcpl, gcpl, H5P_DEFAULT), file(parent))
 end
 function g_create(f::Function, parent::Union{File,Group}, args...)
     g = g_create(parent, args...)
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1426,7 +1426,7 @@ end
 function d_write(parent::Union{File,Group}, name::AbstractString, data; pv...)
     obj, dtype = d_create(parent, name, data; pv...)
     try
-        writearray(obj, dtype.id, data)
+        d_write(obj, dtype, data)
     catch exc
         o_delete(obj)
         rethrow(exc)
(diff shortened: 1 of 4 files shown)


src/HDF5.jl
```diff
@@ -713,7 +713,7 @@ end
 function g_create(parent::Union{File,Group}, path::AbstractString,
                   lcpl::Properties=_link_properties(path),
                   dcpl::Properties=DEFAULT_PROPERTIES)
-    Group(h5g_create(checkvalid(parent), path, lcpl, dcpl), file(parent))
+    Group(h5g_create(checkvalid(parent), path, lcpl, dcpl, HDF5.H5P_DEFAULT), file(parent))
```
musm, line 716: `H5P_DEFAULT` (without `HDF5` prefix)

musm:
I noticed https://github.com/JuliaIO/HDF5.jl/pull/723/commits/cbb142fd6b360cb2caadeabbfc77b7976cc76dff this too in the docs when doing some cleanup work I think several weeks ago, and I couldn't figure out why the hdf5 docs are lying ?

src/deprecated.jl
```diff
@@ -235,3 +235,5 @@ import Base: names
 @deprecate h5g_open(loc_id, pathname) h5g_open(loc_id, pathname, HDF5.H5P_DEFAULT) false
 @deprecate h5l_exists(loc_id, pathname) h5l_exists(loc_id, pathname, HDF5.H5P_DEFAULT) false
 @deprecate h5o_open(loc_id, name) h5o_open(loc_id, name, HDF5.H5P_DEFAULT) false
+
+@deprecate writearray(obj::Attribute, type_id, x) a_write(obj, type_id, x) false
```
musm, line 239: I don't mind adding this, but I really hope no one was using this internal call.

musm:
Unrelated, but I think we should move the contents of `datafile.jl` at the top of the file `HDF5.jl` and consider renaming `DataFile` to `H5DataType` (matching JLD2.jl)

jmert:
> I noticed [cbb142f](https://github.com/JuliaIO/HDF5.jl/commit/cbb142fd6b360cb2caadeabbfc77b7976cc76dff) this too in the docs when doing some cleanup work I think several weeks ago, and I couldn't figure out why the hdf5 docs are lying ?

I'm guessing that the docs are trying to say that there's no behavioral change for any non-default property values, but apparently (presumably due to some shared/generic internal code) it saves the property values anyway.

src/deprecated.jl
jmert: JLD and MAT both do: https://juliahub.com/ui/CodeSearch?q=writearray&u=all&t=all

src/HDF5.jl
jmert: Hah, probably ended up that way from reading the deprecation message and just doing as it told me to :-P

jmert:
Even without adding deprecations for the `h5a_*` and `h5d_*` functions, the "Refine deprecations..." commit (e5af4b9) was sufficient to keep the JLD and MAT test suites passing (with deprecation warnings) — it was very nearly the case that only the `(read|write)array` function deprecations were necessary.

https://github.com/JuliaIO/MAT.jl/compare/master...jmert:HDF5_v0.14
https://github.com/JuliaIO/JLD.jl/compare/master...jmert:HDF5_v0.14

musm:
That's great to hear.  These deprecations look tricky. After the major release after next, it will make our lives a lot easier being able to  remove them 😄

jmert:
OK, I think I managed to add in the deprecations for the `h5(a|d)_(read|write)` methods, so in principle I think this is ready for a review.

In particular, I _think_ all of the type specializations change the data type of the passed through data object so that they don't intersect with a deprecated method; e.g. scalars are wrapped in a `Ref`, strings are passed through as `Ptr{UInt8}`, etc. I was initially afraid that for the `h5a_*` methods where the deprecated methods have the same number of arguments as the low-level `ccall` wrapper would result in cycles that would require ugly `invoke` calls to circumvent, but I think we got lucky and that's not actually an issue.

musm (approved):
Feel free to merge

PR #724, opened 2020-10-31
Grab bag of cleaning up

musm:
See commit list for exact changes:


musm:
I'm debating the first two commits. Commit renaming `DataFile` to `H5DataType` feels right, in the sense that other packages are meant to use that supertype, so having the `H5` prefix makes sense and is required if we want to use the name `DataType` (as it's reserved by `Base.Core`).

src/deprecated.jl
```diff
@@ -231,3 +231,6 @@ end
 
 @deprecate info(obj::Union{Group,File}) g_info(obj) false
 @deprecate objinfo(obj::Union{File,Object}) o_info(obj) false
+
+@deprecate_binding filename HDF5.filename
```
jmert, line 235: I don't think this will work — it's the same binding, so it should always throw a deprecation warning.

Testing confirms:
```
$ julia-dev --project --depwarn=yes
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.6.0-DEV.1397 (2020-10-30)
 _/ |\__'_|_|_|\__'_|  |  Commit 8d55e04b69* (2 days old master)
|__/                   |

julia> using HDF5

🗂️ HDF5.File: read only /tmp/test.h5
└─ 📄 complexf64

julia> HDF5.filename(hfile)
WARNING: HDF5.filename is deprecated, use HDF5.filename instead.
  likely near REPL[3]:1
"/tmp/test.h5"
```

src/HDF5.jl
```diff
@@ -20,7 +20,7 @@ iscontiguous, ishdf5, ismmappable, name,
 o_copy, o_delete, o_open, p_create,
 readmmap, @read, @write, root, set_dims!, t_create, t_commit
 ## Public Types:
-# Attribute, File, Group, Dataset, Datatype, Opaque
+# H5DataType, Attribute, File, Group, Dataset, Datatype, Opaque
```
jmert: I'm not crazy on `H5DataType` being the name because it sounds too close to `Datatype`. What about `H5DataStore`?

src/HDF5.jl
```diff
@@ -960,8 +962,8 @@ Base.haskey(attr::Attributes, path::AbstractString) = h5a_exists(checkvalid(attr
 Base.haskey(dset::Union{Dataset,Datatype}, path::AbstractString) = h5a_exists(checkvalid(dset), path)
 
 # Querying items in the file
-info(obj::Union{Group,File}) = h5g_get_info(checkvalid(obj))
-objinfo(obj::Union{File,Object}) = h5o_get_info(checkvalid(obj))
+g_info(obj::Union{Group,File}) = h5g_get_info(checkvalid(obj))
+o_info(obj::Union{File,Object}) = h5o_get_info(checkvalid(obj))
```
jmert, line 968: :+1:

src/HDF5.jl
musm: Yeah that sounds better. Are we ok with this having the `H5` prefix, which I think is appropriate since it's the supertype? In your opinion `H5DataStore` is better than `H5DataFile` (which seems not as accurate)

jmert: Yeah, I'm good with the H5 prefix since users which are subtyping on it are like to have done an `import`, so being explicit in its name seems fine/good to me. I don't have a strong opinion on `H5DataFile` vs `H5DataStore`, though the fact that `Group <: H5Data(Store|File)` makes me think `File` isn't the ideal suffix.

src/deprecated.jl
musm: Hmm I don't see a way to actually fix this.

musm: Removing the unexports from this PR.

src/deprecated.jl
```diff
@@ -222,3 +222,8 @@ import Base: names
 ### Changed in PR#694
 @deprecate has(parent::Union{File,Group,Dataset}, path::AbstractString) Base.haskey(parent, path)
 @deprecate exists(parent::Union{File,Group,Dataset,Datatype,Attributes}, path::AbstractString) Base.haskey(parent, path)
+
+function Base.read(f::Base.Callable, parent::DataFile, name::AbstractString...)
+    depwarn("Base.read(f::Base.Callable, parent::H5DataType, name::AbstractString...) is deprecated. Directly call `f` on the output from `read(parent, name...)`", :read)
+    f(read(parent, name...)...)
+end
```
jmert: I just noticed that there's a similar `g_create` method that should probably also be deprecated: https://github.com/JuliaIO/HDF5.jl/blob/1097227fcb66c65f38b916af11645c104316f6e0/src/HDF5.jl#L717-L724

musm: Good catch.

Seems like this is supposed to be used as follows

```
h5rewrite(outfile) do fid
    g_create(fid, "mygroup") do g
        write(g, "x", 3.3)
    end
end
```
which kinda strikes me as odd

musm:
@jmert updated the PR with your request. If everythign looks good I plan to merge

src/HDF5.jl
```diff
@@ -1127,7 +1160,7 @@ function a_read(parent::Union{File,Group,Dataset,Datatype}, name::AbstractString
     ret
 end
 
-# Datafile.jl defines generic read for multiple datasets, so we cannot simply add properties here.
+# H5DataStore defines generic read for multiple datasets, so we cannot simply add properties here.
```
jmert, line 1163: This comment isn't necessary anymore, right? With keywords rather than key-value argument pairs, there's not conflict.

jmert (approved):

src/HDF5.jl
musm: That's right. 

I think  these two function need to pass  the `pv`  keyword list down to the `read` function as well.
https://github.com/JuliaIO/HDF5.jl/pull/724/files#diff-3df1923ba1a5ee433e521ee7d3f75d066f7674b297db94ecda3ad89f7fc3e9a2R65 and https://github.com/JuliaIO/HDF5.jl/pull/724/files#diff-3df1923ba1a5ee433e521ee7d3f75d066f7674b297db94ecda3ad89f7fc3e9a2R70

musm: Another minor thing I think we should change is that the method here should probably be private and use an underscore prefix
https://github.com/JuliaIO/HDF5.jl/blob/00cf08352b0640b5b2a25ac10845c22fb167231c/src/HDF5.jl#L440

jmert: I'd say leave the potential change to `read(::H5DataStore, names...; pv...)` to a different PR — one could argue that as generic convenience functions, the interface doesn't need to accept any property values, and if you require property uses you should use the more specific functions that operate on only a single dataset at a time.

PR #561, opened 2019-07-29
WIP: add tree command

bjarthur:
still a work in progress as i haven't tested all corner cases.  please let me know if something doesn't work for you.

for example:

```
help?> tree
  tree(obj::Union{HDF5File,HDF5Group})

  Display the hierarchical structure of object obj, much like the bash tree command does for file
  system paths. Similar to h5dump on the linux command line but more concise as it omits the full
  contents of datasets.

julia> fid=h5open("plate_001/patch_attempt_0000_000/camera_clamp2_000/000/Clamp2.ma")

julia> tree(fid)
MetaArray: 2
data (80000, 3) Float64
└─ info
   ├─ _metaType_: list
   ├─ 0
   │  ├─ _metaType_: dict
   │  ├─ name: 'Channel'
   │  └─ cols
   │     ├─ _metaType_: list
   │     ├─ 0
   │     │  ├─ _metaType_: dict
   │     │  ├─ name: 'command'
   │     │  └─ units: 'V'
   │     ├─ 1
   │     │  ├─ _metaType_: dict
   │     │  ├─ name: 'primary'
   │     │  └─ units: 'A'
   │     └─ 2
   │        ├─ _metaType_: dict
   │        ├─ name: 'secondary'
   │        └─ units: 'V'
   ├─ 1
   │  ├─ _metaType_: dict
   │  ├─ name: 'Time'
   │  ├─ units: 's'
   │  └─ values (80000,) Float64
   └─ 2
      ├─ _metaType_: dict
      ├─ startTime: 1.5627830491764686e9
      ├─ ClampState
      │  ├─ LPFCutoff: 20000.0
      │  ├─ _metaType_: dict
      │  ├─ extCmdScale: 0.02
      │  ├─ holding: -0.07
      │  ├─ membraneCapacitance: 0.0
      │  ├─ mode: 'VC'
      │  ├─ primaryGain: 2.0
      │  ├─ primaryScaleFactor: 1.0e-9
      │  ├─ primarySignal: 'Membrane Current'
      │  ├─ primaryUnits: 'A'
      │  ├─ secondaryGain: 1.0
      │  ├─ secondaryScaleFactor: 0.1
      │  ├─ secondarySignal: 'Membrane Potential'
      │  ├─ secondaryUnits: 'V'
      │  └─ ClampParams
      │     ├─ BridgeBalEnable: 6004
      │     ├─ BridgeBalResist: 8.413395979806202e-42
      │     ├─ FastCompCap: 1.848093076142754e-12
[...]

PR #725, opened 2020-10-31
3-arg show with hiearchical tree printing

jmert:
This is an alternative to #561, adding a simpler version of the tree printing (no values or datatypes yet) but also hooks up the tree printing as the default 3-arg show for most of the high-level datatypes so that the tree printing is shown by default at the REPL.

For example, the output for an HDF5 file handle at the REPL now looks like:
```julia
julia> hfile
🗃️ HDF5 data file: /tmp/jl_mH1PGw
├─ 🏷️ creator
├─ 📑 dtype
├─ 📂 inner
│  ├─ 🏷️ dirty
│  └─ 🔢 data
│     └─ 🏷️ mode
└─ 🔢 version
```

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -92,8 +92,8 @@ function _tree_icon(obj)
         return obj isa Attribute ? "🏷️ " :
                obj isa Group ? "📂 " :
                obj isa Dataset ? "🔢 " :
-               obj isa Datatype ? "📑 " :
-               obj isa File ? "🗃️ " :
+               obj isa Datatype ? "📄 " :
+               obj isa File ? "🗂️ " :
                "❓ "
     else
         return obj isa Attribute ? "[A] " :
--- a/src/show.jl
+++ b/src/show.jl
@@ -39,6 +39,9 @@ function Base.show(io::IO, attr::Attribute)
         print(io, "HDF5.Attribute: (invalid)")
     end
 end
+function Base.show(io::IO, attr::Attributes)
+    print(io, "Attributes of ", attr.parent)
+end
 
 function Base.show(io::IO, dtype::Datatype)
     print(io, "HDF5.Datatype: ")
@@ -104,10 +107,10 @@ function _tree_icon(obj)
                "[?] "
     end
 end
+_tree_icon(obj::Attributes) = _tree_icon(obj.parent)
 
-_tree_head(io::IO, obj::Union{File, Group, Dataset, Attribute}) = println(io, _tree_icon(obj), obj)
-_tree_head(io::IO, obj::Datatype) = println(io, _tree_icon(obj), "HDF5 Datatype: ", name(obj))
-_tree_head(io::IO, obj::Attributes) = println(io, _tree_icon(obj.parent), "Attributes of ", obj.parent)
+_tree_head(io::IO, obj) = println(io, _tree_icon(obj), obj)
+_tree_head(io::IO, obj::Datatype) = println(io, _tree_icon(obj), "HDF5.Datatype: ", name(obj))
 
 function _tree_children(parent::Union{File, Group}, attributes::Bool)
     names = keys(parent)
--- a/src/show.jl
+++ b/src/show.jl
@@ -144,7 +144,7 @@ function _tree_children(parent::Union{Attribute, Datatype}, attributes::Bool)
     return (String[], Union{Object, Attribute}[])
 end
 
-function show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Attribute}, indent::String="";
+function _show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Attribute}, indent::String="";
                    attributes::Bool = true)
     isempty(indent) && _tree_head(io, obj)
     !isvalid(obj) && return
@@ -160,8 +160,14 @@ function show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Att
         println(io, indent, islast ? "└─ " : "├─ ", icon, name)
 
         nextindent = indent * (islast ? "   " : "│  ")
-        show_tree(io, child, nextindent; attributes = attributes)
+        _show_tree(io, child, nextindent; attributes = attributes)
     end
     return nothing
 end
+
 show_tree(obj; kws...) = show_tree(stdout, obj; kws...)
+function show_tree(io::IO, obj; kws...)
+    buf = IOBuffer()
+    _show_tree(IOContext(buf, io), obj; kws...)
+    print(io, String(take!(buf)))
+end
--- a/src/show.jl
+++ b/src/show.jl
@@ -64,14 +64,46 @@ function Base.show(io::IO, dtype::Datatype)
     end
 end
 
-Base.show(io::IO, ::MIME"text/plain", obj::Union{File,Group,Dataset,Attributes,Attribute}) = show_tree(io, obj)
+"""
+    SHOW_TREE = Ref{Bool}(true)
 
-_tree_icon(obj) = obj isa Attribute ? "🏷️ " :
-                  obj isa Group ? "📂 " :
-                  obj isa Dataset ? "🔢 " :
-                  obj isa Datatype ? "📑 " :
-                  obj isa File ? "🗃️ " :
-                  "❓ "
+Configurable option to control whether the default `show` for HDF5 objects is printed
+using `show_tree` or not.
+"""
+const SHOW_TREE = Ref{Bool}(true)
+"""
+    SHOW_TREE_ICONS = Ref{Bool}(true)
+
+Configurable option to control whether emoji icons (`true`) or a plain-text annotation
+(`false`) is used to indicate the object type by `show_tree`.
+"""
+const SHOW_TREE_ICONS = Ref{Bool}(true)
+
+function Base.show(io::IO, ::MIME"text/plain", obj::Union{File,Group,Dataset,Attributes,Attribute})
+    if SHOW_TREE[]
+        show_tree(io, obj)
+    else
+        show(io, obj)
+    end
+end
+
+function _tree_icon(obj)
+    if SHOW_TREE_ICONS[]
+        return obj isa Attribute ? "🏷️ " :
+               obj isa Group ? "📂 " :
+               obj isa Dataset ? "🔢 " :
+               obj isa Datatype ? "📑 " :
+               obj isa File ? "🗃️ " :
+               "❓ "
+    else
+        return obj isa Attribute ? "[A] " :
+               obj isa Group ? "[G]" :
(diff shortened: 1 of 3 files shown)


musm:
This is awesome, I really love the output.

musm:
Just some minor things as I was playing around with this:
- 📑  Is the pink that there supposed to highlight that this isn't just a document but rather the 'datatype tag', perhaps the simpler 📄 could also be  appropriate? 
- 🗃️ I totally get why it's a good idea to use this symbol for the file. OTOH perhaps just using  📂  would also work be visually simpler?

jmert:
> Just some minor things as I was playing around with this:
> 
> * 📑 Is the pink that there supposed to highlight that this isn't just a document but rather the 'datatype tag', perhaps the simpler :page_facing_up: could also be appropriate?

The choice was relatively arbitrary — I just browsed through the list of emoji at https://emojipedia.org/objects/ and picked one that I found quickly. Using 📄 is fine with me.


> * 🗃️  I totally get why it's a good idea to use this symbol for the file. OTOH perhaps just reusing the group symbol 📂 would be visually simpler, albeit confuses the two different objects (although the fact that the data file has an additional line would negate this confusion).

I'd vote to keep the printing distinct, but that's just my personal preference. I initially had them printing the same, but I thought it was useful to visually see the distinction between having a `File` object versus the root group.

musm:
I also just stumbled upon 🗂️      https://emojipedia.org/card-index-dividers/, which might be another (better?) option  for 🗃️.

```
julia> hfile
🗂️ HDF5 data file: /tmp/jl_mH1PGw
├─ 🏷️ creator
├─ 📄 dtype
├─ 📂 inner
│  ├─ 🏷️ dirty
│  └─ 🔢 data
│     └─ 🏷️ mode
└─ 🔢 version
```

```
julia> hfile
🗃️ HDF5 data file: /tmp/jl_mH1PGw
├─ 🏷️ creator
├─ 📄 dtype
├─ 📂 inner
│  ├─ 🏷️ dirty
│  └─ 🔢 data
│     └─ 🏷️ mode
└─ 🔢 version
```

src/show.jl
```diff
@@ -0,0 +1,164 @@
...
+    else
+        print(io, "HDF5 data file (closed): ", fid.filename)
+    end
+end
+
+function Base.show(io::IO, g::Group)
+    if isvalid(g)
+        print(io, "HDF5 group: ", name(g), " (file: ", g.file.filename, ")")
+    else
+        print(io, "HDF5 group (invalid)")
```
musm, line 13: Should we change these to `HDF5 Group` or `HDF5.Group`  (capitalize them to match the struct name?

jmert: `HDF5.Group` and similar sounds good to me.

musm:
One issue I had with this PR on my terminal is the spacing of some of the text is not aligned
image

(note the two spaces between the `dirty` attribute` and the `data` dataset.)

jmert:
> One issue I had with this PR on my terminal is the spacing of some of the text is not aligned
> 
> (note the two spaces between the `dirty` attribute`and the`data` dataset.)

I end up with the opposite problem — missing spaces :-\

musm:
Interesting, is this just a terminal rendering issue or a bug in Julia's printing? (I'm using the Windows Terminal)

jmert:
I'm pretty sure it's a terminal issue since copy-pasting to text editors and comments here works just fine.

musm:
Interesting, if I copy and paste the string  into my terminal,  everything looks good.
image


This suggests if we just build up the string buffer recursively and then print it out at the last call it should likely just work.

jmert:
> This suggests if we just build up the string buffer recursively and then print it out just once, at the last call it should likely just work.

Doing that doesn't help on my terminal:
```diff
diff --git a/src/show.jl b/src/show.jl
index 9d19017..6d526de 100644
--- a/src/show.jl
+++ b/src/show.jl
@@ -144,7 +144,7 @@ function _tree_children(parent::Union{Attribute, Datatype}, attributes::Bool)
     return (String[], Union{Object, Attribute}[])
 end
 
-function show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Attribute}, indent::String="";
+function _show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Attribute}, indent::String="";
                    attributes::Bool = true)
     isempty(indent) && _tree_head(io, obj)
     !isvalid(obj) && return
@@ -160,8 +160,14 @@ function show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Att
         println(io, indent, islast ? "└─ " : "├─ ", icon, name)
 
         nextindent = indent * (islast ? "   " : "│  ")
-        show_tree(io, child, nextindent; attributes = attributes)
+        _show_tree(io, child, nextindent; attributes = attributes)
     end
     return nothing
 end
+
 show_tree(obj; kws...) = show_tree(stdout, obj; kws...)
+function show_tree(io::IO, obj; kws...)
+    buf = IOBuffer()
+    _show_tree(IOContext(buf, io), obj; kws...)
+    print(io, String(take!(buf)))
+end
```
image

musm:
That fixed it on my terminal! 

image

src/show.jl
```diff
@@ -0,0 +1,174 @@
...
+    if SHOW_TREE[]
+        show_tree(io, obj)
+    else
+        show(io, obj)
+    end
+end
+
+function _tree_icon(obj)
+    if SHOW_TREE_ICONS[]
+        return obj isa Attribute ? "🏷️ " :
```
musm, line 96: Minor nit, might be better to remove the space after the emoji here and add in the `print` line  in the `_tree....` function.

src/show.jl
```diff
@@ -0,0 +1,174 @@
...
+
+    names, children = _tree_children(obj, attributes)
+    nchildren = length(children)
+    for ii in 1:nchildren
+        name = names[ii]
+        child  = children[ii]
+
+        islast = ii == nchildren
+        icon = _tree_icon(child)
+        println(io, indent, islast ? "└─ " : "├─ ", icon, name)
```
musm, line 161:
```suggestion
        print(io, "\n", indent, islast ? "└─ " : "├─ ", icon, name)
```

src/show.jl
```diff
@@ -0,0 +1,174 @@
...
+               obj isa Dataset ? "[D] " :
+               obj isa Datatype ? "[T] " :
+               obj isa File ? "[F] " :
+               "[?] "
+    end
+end
+_tree_icon(obj::Attributes) = _tree_icon(obj.parent)
+
+_tree_head(io::IO, obj) = println(io, _tree_icon(obj), obj)
+_tree_head(io::IO, obj::Datatype) = println(io, _tree_icon(obj), "HDF5.Datatype: ", name(obj))
```
musm, line 114:
```suggestion
_tree_head(io::IO, obj) = print(io, _tree_icon(obj), obj)
_tree_head(io::IO, obj::Datatype) = print(io, _tree_icon(obj), "HDF5.Datatype: ", name(obj))
```

musm (commented):
This fixes the extra newline printed.

src/show.jl
```diff
@@ -0,0 +1,174 @@
+function Base.show(io::IO, fid::File)
```
musm, line 1: It would be nice if there was an easy way to print the file mode `"r", "r+", "w"` etc... Not sure if there's an easy way to go about this since it doesn't seem to be captured in the File struct

jmert: Without storing the original mode choice, we can't have a distinction between, e.g. `"cw", "w", "r+"`, but we can get read-write vs read only:
```julia
julia> hfile = h5open("/tmp/test.h5", swmr = true)
🗂️ HDF5.File: read only (swmr) /tmp/test.h5
└─ 📄 complexf64

julia> close(hfile)

🗂️ HDF5.File: read only /tmp/test.h5
└─ 📄 complexf64
```

Added in the newest commit.

musm: That's awesome. Maybe we can put them all in parens. And also perhaps put them after the filename is printed?  (read only, swmr)  ? 
```
🗂️ HDF5.File: /tmp/test.h5 (read only, swmr)

```

jmert: I agree with all in parens, but I'm not as sure about after the filename; I've worked/working in an environment where we end up with very long filenames (both deep path and near-maximum file names ...), and to bury some useful information way at the end would be unfortunate.

musm: That makes sense. One minor thing, lets also add the dash between `(write|read)-only`

musm (approved):
This is such a massive improvement in UX

src/show.jl
```diff
@@ -0,0 +1,179 @@
...
+        print(io, "\n", indent, islast ? "└─ " : "├─ ", icon, " ", name)
+
+        nextindent = indent * (islast ? "   " : "│  ")
+        _show_tree(io, child, nextindent; attributes = attributes)
+    end
+    return nothing
+end
+
+show_tree(obj; kws...) = show_tree(stdout, obj; kws...)
+function show_tree(io::IO, obj; kws...)
```
musm, line 175: One final comment, are we happy with the name here or should we use `tree` ? We should probably also add this to the top of the `src.jl` module name as a  `public` but module scoped function as in here https://github.com/JuliaIO/HDF5.jl/pull/724/files

src/show.jl
```diff
@@ -0,0 +1,179 @@
...
+        icon = _tree_icon(child)
+        print(io, "\n", indent, islast ? "└─ " : "├─ ", icon, " ", name)
+
+        nextindent = indent * (islast ? "   " : "│  ")
+        _show_tree(io, child, nextindent; attributes = attributes)
+    end
+    return nothing
+end
+
+show_tree(obj; kws...) = show_tree(stdout, obj; kws...)
```
musm, line 174: One final comment, are we happy with the name here or should we use `tree` ? We should probably also add this to the top of the `src.jl` module name as a  `public` but module scoped function as in here https://github.com/JuliaIO/HDF5.jl/pull/724/files

jmert: I don't feel strongly about the name, so I'm fine with renaming it. A public way to get the printing no matter what the name of this function is is to call `show(stdout, MIME"text/plain"(), h5object)`.

musm: Yeah that's a good point, I don't mind keeping it as is, but was brainstorming given the previous PR used the name `tree`.

musm:
Is this PR ready to merge or do you plan to make any additional changes?

jmert:
It's ready. I think there are potential changes regarding adding extra contextual information, but that can happen in future work.

bjarthur:
is it dynamically expandable like REPLTreeViews?

jmert:
No, this is purely a static print. Some of the code elements should be reusable for an implementation tied in with `REPLTreeViews`, though.

PR #732, opened 2020-11-04
remove hdf5_to_julia

kleinhenz:
This performs some cleanup leftover from #652. 

Specifically it removes the `hdf5_to_julia` function in favor of the `get_jl_type` function. This function was just kept around because it was used in `setindex!` and `readmmap` and I didn't get around to changing those uses in the original PR. The main difference between `hdf5_to_julia` and `get_jl_type` is that `hdf5_to_julia` looks at the dataspace and returns `Array{T}` if the space is not scalar or null. IMHO this behavior doesn't make much sense because the `Array` type doesn't contain its size so if you actually want to do something you have to construct the dataspace again which seems wasteful. 

I also removed `hdf5_to_julia_eltype` which almost completely duplicates the functionality of `get_jl_type` except for a few differences in string handling that aren't relevant anymore after #652. 

I expect that this may break uses in `MAT.jl` or `JLD.jl` so I'm not sure if it would be better to just deprecate these functions for now.


src/HDF5.jl
```diff
@@ -1300,13 +1291,17 @@ function iscontiguous(obj::Dataset)
     end
 end
 
-ismmappable(::Type{Array{T}}) where {T<:ScalarType} = true
+ismmappable(::Type{T}) where {T<:ScalarType} = true
```
musm, line 1294:
```suggestion
ismmappable(::Type{<:ScalarTYpe}) = true
```

src/HDF5.jl
```diff
@@ -1465,53 +1460,60 @@ Base.axes(dset::Dataset) = map(Base.OneTo, size(dset))
...
-    if !(T <: Array)
-        error("Dataset indexing (hyperslab) is available only for arrays")
-    end
-    ET = eltype(T)
-    if !(ET <: Union{ScalarType, Complex{<:ScalarType}})
-        error("Dataset indexing (hyperslab) is available only for bits types")
-    end
-    if length(X) != prod(map(length, I))
-        error("number of elements in range and length of array must be equal")
+IndexType = Union{AbstractRange{Int},Int,Colon}
```
musm, line 1463:
```suggestion
const IndexType = Union{AbstractRange{Int},Int,Colon}
```

src/HDF5.jl
```diff
@@ -1465,53 +1460,60 @@ Base.axes(dset::Dataset) = map(Base.OneTo, size(dset))
...
+
+    # perform conversions for numeric types
+    if (U <: Number) && (T <: Number) && U !== T
+        X = convert(Array{U}, X)
     end
-    if eltype(X) != ET
-        X = convert(Array{ET}, X)
+
+    filetype = datatype(dset)
+    memtype = Datatype(h5t_get_native_type(filetype.id))  # padded layout in memory
```
musm, line 1474:
```suggestion
    memtype = Datatype(h5t_get_native_type(filetype))  # padded layout in memory
```

src/HDF5.jl
```diff
@@ -1465,53 +1460,60 @@ Base.axes(dset::Dataset) = map(Base.OneTo, size(dset))
...
+    dspace = dataspace(dset)
+    stype = h5s_get_simple_extent_type(dspace.id)
+    stype == H5S_NULL && error("attempting to write to null dataspace")
+
+    indices = Base.to_indices(dset, I)
+    dspace = hyperslab(dspace, indices...)
+
     memspace = dataspace(X)
+
+    if h5s_get_select_npoints(dspace.id) != h5s_get_select_npoints(memspace.id)
```
musm, line 1494:
```suggestion
    if h5s_get_select_npoints(dspace) != h5s_get_select_npoints(memspace)
```

src/HDF5.jl
```diff
@@ -1465,53 +1460,60 @@ Base.axes(dset::Dataset) = map(Base.OneTo, size(dset))
...
+        error("number of elements in src and dest arrays must be equal")
+    end
+
     try
-        h5d_write(dset, memtype, memspace, dsel_id, dset.xfer, X)
+        h5d_write(dset.id, memtype.id, memspace.id, dspace.id, dset.xfer.id, X)
```
musm, line 1499:
```suggestion
        h5d_write(dset, memtype, memspace, dspace, dset.xfer, X)
```

src/HDF5.jl
```diff
@@ -1300,13 +1291,17 @@ function iscontiguous(obj::Dataset)
     end
 end
 
-ismmappable(::Type{Array{T}}) where {T<:ScalarType} = true
+ismmappable(::Type{<:ScalarTYpe}) = true
```
musm, line 1294:
```suggestion
ismmappable(::Type{<:ScalarType}) = true
```

musm:
Looks like we're getting a type size mismatch on x86 on Windows

src/HDF5.jl
```diff
@@ -1465,53 +1460,60 @@ Base.axes(dset::Dataset) = map(Base.OneTo, size(dset))
...
+        X = convert(Array{U}, X)
     end
-    if eltype(X) != ET
-        X = convert(Array{ET}, X)
+
+    filetype = datatype(dset)
+    memtype = Datatype(h5t_get_native_type(filetype))  # padded layout in memory
+    close(filetype)
+
+    if sizeof(T) != h5t_get_size(memtype.id)
```
musm, line 1477:
```suggestion
    if sizeof(T) != h5t_get_size(memtype)
```

src/HDF5.jl
```diff
@@ -1465,53 +1460,60 @@ Base.axes(dset::Dataset) = map(Base.OneTo, size(dset))
...
+        error("""
+              Type size mismatch
+              sizeof($T) = $(sizeof(T))
+              sizeof($memtype) = $(h5t_get_size(memtype.id))
```
musm, line 1481:
```suggestion
              sizeof($memtype) = $(h5t_get_size(memtype))
```

src/HDF5.jl
```diff
@@ -1465,53 +1460,60 @@ Base.axes(dset::Dataset) = map(Base.OneTo, size(dset))
...
+              """)
     end
-    dsel_id = hyperslab(dset, I...)
-    memtype = datatype(X)
+
+    dspace = dataspace(dset)
+    stype = h5s_get_simple_extent_type(dspace.id)
```
musm, line 1486:
```suggestion
    stype = h5s_get_simple_extent_type(dspace)
```

musm:
For example take a look at https://github.com/JuliaIO/HDF5.jl/blob/master/test/extend_test.jl#L45 
```julia
julia> b[1:200] = ones(200)
ERROR: Type size mismatch
sizeof(Float64) = 8
sizeof(HDF5.Datatype: H5T_STD_I32LE) = 4


julia> b[1:200]
200-element Vector{Int32}:
   1
   2
   3
   4
   5
   6
   7

julia> ones(200)
200-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0

```

kleinhenz:
I'll fix x86 tomorrow

gen/api_defs.jl
```diff
@@ -205,6 +205,7 @@
 @bind h5s_get_simple_extent_dims(space_id::hid_t, dims::Ptr{hsize_t}, maxdims::Ptr{hsize_t})::Cint "Error getting the dimensions for a dataspace"
 @bind h5s_get_simple_extent_ndims(space_id::hid_t)::Cint "Error getting the number of dimensions for a dataspace"
 @bind h5s_get_simple_extent_type(space_id::hid_t)::Cint "Error getting the dataspace type"
+@bind h5s_get_select_npoints(space_id::hid_t)::hsize_t "Error getting the number of selected points"
```
jmert, line 208: Adding a new API binding should also add another line into `docs/src/api_bindings.md` that should be committed as well.

musm:
Great tests are passing now!

src/HDF5.jl
```diff
@@ -1465,53 +1460,61 @@ Base.axes(dset::Dataset) = map(Base.OneTo, size(dset))
...
-    end
-    if length(X) != prod(map(length, I))
-        error("number of elements in range and length of array must be equal")
+const IndexType = Union{AbstractRange{Int},Int,Colon}
+function Base.setindex!(dset::Dataset, X::Array{T}, I::IndexType...) where T
+    !isconcretetype(T) && error("type $T is not concrete")
+    U = get_jl_type(dset)
+
+    # perform conversions for numeric types
+    if (U <: Number) && (T <: Number) && U !== T
```
jmert, line 1469: `Base` doesn't have this restriction, so shouldn't we do the same — just convert if the types don't match?

For example:
```julia
julia> using Test

julia> a = GenericString["a1", "a2"]
2-element Vector{GenericString}:
 "a1"
 "a2"

julia> b = Vector{String}(undef, 2)
2-element Vector{String}:
 #undef
 #undef

julia> b[:] = a
       b
2-element Vector{String}:
 "a1"
 "a2"
```

kleinhenz: My reasoning here was that `T` and `U` might be technically different types but be structurally the same so that it is unnecessary to require a convert method. For example in the case of compound types `U` will always be a named tuple but `T` could be a user defined struct with the same layout. It seems desirable to support this case without overhead. 

Also, the string example from above already doesn't work with this or in the previous implementation because writing strings requires special handling in order to convert them to a format usable by hdf5. Unifying this in a similar way to #652 but on the write side would be nice but I'm not sure exactly the best way to do it yet.

musm (approved):
LGTM, jmert could you comment on the deprecation path for MAT/JLD ? Is this ok to merge without them?

jmert:
I just pushed a commit that adds the deprecations — I had previously already replaced the use of `hdf5_to_julia` in MAT (https://github.com/JuliaIO/MAT.jl/pull/145/commits/44e3512fbfa5e0f36bcd7c2119c7022a8f77cf5c) when updating for the `read()` changes earlier, and replacing uses of `hdf5_to_julia_eltype` in JLD was straight forward.

The change in the signature of `readmmap` is actually what tripped up the JLD test suite, and since that was actually publicly advertised, we definitely want the deprecations I added for that and `ismmappable`.

musm:
Great many thanks!  Any objections before merging?

jmert (approved):

PR #738, opened 2020-11-13
Fix #737, extend Base.setindex! in deprectation

### LINKED ISSUE #737: setindex! deprecation conflicts with Base.setindex!
The following does not work with HDF5 v0.14:

```julia
using HDF5
x = zeros(4)
setindex!(x, 1.2, 2)

# WARNING: both HDF5 and Base export "setindex!"; uses of it in module Main must be qualified
# ERROR: LoadError: UndefVarError: setindex! not defined
```

This is because of the deprecation

```julia
@deprecate setindex!(p::Properties, val, name::AbstractString) setindex!(p, val, Symbol(name))
```

which exports a `HDF5.setindex!` method, instead of `Base.setindex!`.

Unfortunately, it doesn't seem to be possible to write `@deprecate Base.setindex!(...) setindex!(...)`, which would be an easy solution. A simple alternative would be to manually define `Base.setindex!` with the old signature, and print a custom deprecation message.

jmert:


musm:
Many thanks, I'll release a patch with thix fix.

PR #739, opened 2020-11-16
Upgrade HDF5 to v1.12

musm:
We had to drop binaries for arm and i686 Linux  since they are no longer available. 

There is also some work to try and get cross-compilation working, but that will take some time. When cross-compilation works thought cmake, we should be able to ship binaries for all the platforms Julia supports but for not that is proving to be quite challenging.


.github/workflows/CI.yml
```diff
@@ -25,6 +25,9 @@ jobs:
             arch: x86
```
musm, line 25: test

musm:
@jmert  We can now look at implementing some of those v1.12 upgrades in the todo's scattered in the files.

PR #741, opened 2020-11-17
Upgrade API to v1.12

musm:


src/HDF5.jl
```diff
@@ -919,7 +919,7 @@ Base.haskey(dset::Union{Dataset,Datatype}, path::AbstractString) = h5a_exists(ch
 
 # Querying items in the file
 group_info(obj::Union{Group,File}) = h5g_get_info(checkvalid(obj))
-object_info(obj::Union{File,Object}) = h5o_get_info(checkvalid(obj))
+object_info(obj::Union{File,Object}) = h5o_get_info(checkvalid(obj), H5O_INFO_ALL)
```
musm, line 922: This function is prety low level as is to expose. Needs work.

mkitti:
Did we need to rebase this?

musm:
Probably, but the reason for not upgrading API is that it needs to be done in a backwards compatible way with older HDF5 API to support older library versions. I believe I found it somewhat challenging to determine the best way of going about this especially with the changes to the struct definition. I'd have to revisit.

PR #743, opened 2020-11-18
Restrict HDF5_jll to v1.10 series

jmert:
Note that this is set to merge into the v0.13 release branch.

With HDF5_jll serving v1.12 libraries, this has caused problems in the previous HDF5.jl release (which is still being used by JLD and MAT, for example) because the version bounds weren't sent restrictively enough to stick to the same libhdf5 "major" release.

This is an attempt to fix JuliaIO/JLD.jl#281


musm:
Thanks for the this

musm:
I should tag a release on this branch with this fix right.

jmert:
Yes, please.

PR #745, opened 2020-11-19
Fix #742, avoid instantiating large tuples

### LINKED ISSUE #742: Very slow reading of long strings in v0.14
I have some HDF5 files which contain very long strings (>300k characters). With `HDF5` v0.13.6, I can read these no problem:
```julia
import Random
import HDF5

file = HDF5.h5open("string.h5")
file["string"] = Random.randstring(300000)
close(file)
```
```julia
import HDF5
using BenchmarkTools
@btime HDF5.h5open("string.h5") do file
    read(file["string"]);
end
593.301 μs (12 allocations: 586.38 KiB)
```

Updating to v0.14.1, however, even reading this same dataset just once takes so long that I haven't been able to measure it. For shorter strings (10k characters), I see a ~3-fold slowdown (700 μs vs 220 μs), but nothing so egregious.

All of this is on Windows 10, Julia 1.5.1

Any ideas?

jmert:
This is an attempt at fixing #742 by replacing typed arrays with a generic bytes buffer for types which require special read handling.

The underlying issue in #742 appears to be that instantiating either of `FixedString` or `FixedArray` types which large values of `N` in the tuple causes Julia's type inference and/or codegen to take extremely long/fail. Note that instantiation to a "physical object" seems to be the problem, and type-only constructs (i.e. `Type{FixedString{30000,0}}`) seem to work OK, so it's convenient to leave the type declarations as-is since it makes things like computing the `sizeof` a `NamedTuple` with an arbitrarily-sized `FixedString` element work without any extra effort on our part.

The cost to keeping some of the type (size) information, though, is that we have to jump through hoops to avoid actually instantiating the large tuples. What this PR does to achieve that is to explicitly pass around an `Array{UInt8}` buffer to the internal `normalize_types` functions, which all had to be modified to work directly on the untyped data buffer instead.

I would very much appreciate more eyes on this code since it's pretty invasive. Furthermore, I haven't attempted any sort of profiling to see if this construction creates significant regression in performance (precompile/load time or actual runtime uses), so that should probably be done before merging.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1382,6 +1382,8 @@ Base.getindex(parent::Dataset, r::Reference) = _deref(parent, r) # defined separ
 
 # convert special types to native julia types
 function normalize_types(::Type{T}, buf::AbstractMatrix{UInt8}) where {T}
+    # First dimension spans bytes of a single element of type T --- (recursively) normalize
+    # each range of bytes to final type, returning vector of normalized data.
     return [_normalize_types(T, view(buf, :, ind)) for ind in axes(buf, 2)]
 end
 
@@ -1400,6 +1402,9 @@ end
 
 _normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {T} = _typed_load(T, buf)
 function _normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {K, T <: NamedTuple{K}}
+    # Compound data types do not necessarily have members of uniform size, so instead of
+    # dim-1 => bytes of single element and dim-2 => over elements, just loop over exact
+    # byte ranges within the provided buffer vector.
     nv = ntuple(length(K)) do ii
         elT = fieldtype(T, ii)
         off = fieldoffset(T, ii) % Int
@@ -1412,15 +1417,24 @@ function _normalize_types(::Type{V}, buf::AbstractVector{UInt8}) where {T, V <: 
     va = _typed_load(V, buf)
     pbuf = unsafe_wrap(Array, convert(Ptr{UInt8}, va.p), (sizeof(T), Int(va.len)))
     if do_normalize(T)
+        # If `T` a non-trivial type, recursively normalize the vlen buffer.
         return normalize_types(T, pbuf)
     else
+        # Otherwise if `T` is simple type, directly reinterpret the vlen buffer.
+        # (copy since libhdf5 will reclaim `pbuf = va.p` in `h5d_vlen_reclaim`)
         return copy(vec(reinterpret(T, pbuf)))
     end
 end
 function _normalize_types(::Type{F}, buf::AbstractVector{UInt8}) where {T, F <: FixedArray{T}}
     if do_normalize(T)
+        # If `T` a non-trivial type, recursively normalize the buffer after reshaping to
+        # matrix with dim-1 => bytes of single element and dim-2 => over elements.
         return reshape(normalize_types(T, reshape(buf, sizeof(T), :)), size(F)...)
     else
+        # Otherwise, if `T` is simple type, directly reinterpret the array and reshape to
+        # final dimensions. The copy ensures (a) the returned array is independent of
+        # [potentially much larger] read() buffer, and (b) that the returned data is an
+        # Array and not ReshapedArray of ReinterpretArray of SubArray of ...
         return copy(reshape(reinterpret(T, buf), size(F)...))
     end
 end
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1398,6 +1398,10 @@ function _typed_load(::Type{T}, buf::V) where {T, V <: Union{Vector{UInt8}, Base
     dest = Ref{T}()
     GC.@preserve dest buf Base._memcpy!(unsafe_convert(Ptr{Cvoid}, dest), pointer(buf), sizeof(T))
     return dest[]
+    # TODO: The above can maybe be replaced with
+    #   return GC.@preserve buf unsafe_load(convert(Ptr{t}, pointer(buf)))
+    # dependent on data elements being properly aligned for all datatypes, on all
+    # platforms.
 end
 
 _normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {T} = _typed_load(T, buf)
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1280,6 +1280,9 @@ function Base.read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
     end
 
     if do_normalize(T)
+        # The entire dataset is read into in a buffer matrix where the first dimension at
+        # any stage of normalization is the bytes for a single element of type `T`, and
+        # the second dimension of the matrix runs through all elements.
         buf = Matrix{UInt8}(undef, sizeof(T), prod(sz))
     else
         buf = Array{T}(undef, sz...)
@@ -1382,12 +1385,13 @@ function normalize_types(::Type{T}, buf::AbstractMatrix{UInt8}) where {T}
     return [_normalize_types(T, view(buf, :, ind)) for ind in axes(buf, 2)]
 end
 
-# high-level description which should always work
+# high-level description which should always work --- here, the buffer contains the bytes
+# for exactly 1 element of an object of type T, so reinterpret the `UInt8` vector as a
+# length-1 array of type `T` and extract the (only) element.
 function _typed_load(::Type{T}, buf::AbstractVector{UInt8}) where {T}
     return @inbounds reinterpret(T, buf)[1]
 end
-# fast-path for common concrete types with simple layout (which should be nearly all
-# cases)
+# fast-path for common concrete types with simple layout (which should be nearly all cases)
 function _typed_load(::Type{T}, buf::V) where {T, V <: Union{Vector{UInt8}, Base.FastContiguousSubArray{UInt8,1}}}
     dest = Ref{T}()
     GC.@preserve dest buf Base._memcpy!(unsafe_convert(Ptr{Cvoid}, dest), pointer(buf), sizeof(T))
@@ -1406,7 +1410,7 @@ function _normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {K, T <: 
 end
 function _normalize_types(::Type{V}, buf::AbstractVector{UInt8}) where {T, V <: VariableArray{T}}
     va = _typed_load(V, buf)
-    pbuf = unsafe_wrap(Array, convert(Ptr{UInt8}, va.p), (sizeof(T), Int(va.len)), own = false)
+    pbuf = unsafe_wrap(Array, convert(Ptr{UInt8}, va.p), (sizeof(T), Int(va.len)))
     if do_normalize(T)
         return normalize_types(T, pbuf)
     else
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1279,8 +1279,12 @@ function Base.read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
         end
     end
 
-    buf = Array{T}(undef, sz...)
-    memspace = dataspace(buf)
+    if do_normalize(T)
+        buf = Matrix{UInt8}(undef, sizeof(T), prod(sz))
+    else
+        buf = Array{T}(undef, sz...)
+    end
+    memspace = dataspace(sz)
 
     if obj isa Dataset
         h5d_read(obj, memtype, memspace, dspace, obj.xfer, buf)
@@ -1288,7 +1292,11 @@ function Base.read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
         h5a_read(obj, memtype, buf)
     end
 
-    out = do_normalize(T) ? normalize_types.(buf) : buf
+    if do_normalize(T)
+        out = reshape(normalize_types(T, buf), sz...)
+    else
+        out = buf
+    end
 
     xfer_id = obj isa Dataset ? obj.xfer.id : H5P_DEFAULT
     do_reclaim(T) && h5d_vlen_reclaim(memtype, memspace, xfer_id, buf)
@@ -1370,12 +1378,50 @@ Base.getindex(parent::Union{File,Group}, r::Reference) = _deref(parent, r)
 Base.getindex(parent::Dataset, r::Reference) = _deref(parent, r) # defined separately to resolve ambiguity
 
 # convert special types to native julia types
-normalize_types(x) = x
-normalize_types(x::NamedTuple{T}) where {T} = NamedTuple{T}(map(normalize_types, values(x)))
-normalize_types(x::Cstring) = unsafe_string(x)
-normalize_types(x::FixedString) = unpad(String(collect(x.data)), pad(x))
-normalize_types(x::FixedArray) = normalize_types.(reshape(collect(x.data), size(x)...))
-normalize_types(x::VariableArray) = normalize_types.(copy(unsafe_wrap(Array, convert(Ptr{eltype(x)}, x.p), x.len, own=false)))
+function normalize_types(::Type{T}, buf::AbstractMatrix{UInt8}) where {T}
+    return [_normalize_types(T, view(buf, :, ind)) for ind in axes(buf, 2)]
+end
+
+# high-level description which should always work
+function _typed_load(::Type{T}, buf::AbstractVector{UInt8}) where {T}
+    return @inbounds reinterpret(T, buf)[1]
+end
+# fast-path for common concrete types with simple layout (which should be nearly all
+# cases)
+function _typed_load(::Type{T}, buf::V) where {T, V <: Union{Vector{UInt8}, Base.FastContiguousSubArray{UInt8,1}}}
+    dest = Ref{T}()
+    GC.@preserve dest buf Base._memcpy!(unsafe_convert(Ptr{Cvoid}, dest), pointer(buf), sizeof(T))
+    return dest[]
+end
+
+_normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {T} = _typed_load(T, buf)
+function _normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {K, T <: NamedTuple{K}}
+    nv = ntuple(length(K)) do ii
+        elT = fieldtype(T, ii)
+        off = fieldoffset(T, ii) % Int
+        sub = view(buf, off .+ (1:sizeof(elT)))
+        return _normalize_types(elT, sub)
+    end
+    return NamedTuple{K}(nv)
+end
+function _normalize_types(::Type{V}, buf::AbstractVector{UInt8}) where {T, V <: VariableArray{T}}
+    va = _typed_load(V, buf)
+    pbuf = unsafe_wrap(Array, convert(Ptr{UInt8}, va.p), (sizeof(T), Int(va.len)), own = false)
+    if do_normalize(T)
+        return normalize_types(T, pbuf)
+    else
+        return copy(vec(reinterpret(T, pbuf)))
+    end
+end
+function _normalize_types(::Type{F}, buf::AbstractVector{UInt8}) where {T, F <: FixedArray{T}}
+    if do_normalize(T)
+        return reshape(normalize_types(T, reshape(buf, sizeof(T), :)), size(F)...)
+    else
+        return copy(reshape(reinterpret(T, buf), size(F)...))
+    end
+end
+_normalize_types(::Type{Cstring}, buf::AbstractVector{UInt8}) = unsafe_string(_typed_load(Ptr{UInt8}, buf))
+_normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {T <: FixedString} = unpad(String(buf), pad(T))
 
 do_normalize(::Type{T}) where {T} = false
 do_normalize(::Type{NamedTuple{T,U}}) where {U,T} = any(i -> do_normalize(fieldtype(U,i)), 1:fieldcount(U))
--- a/test/plain.jl
+++ b/test/plain.jl
@@ -1065,20 +1065,21 @@ end
     # issue #742, large fixed strings are readable
     mktemp() do path, io
         close(io)
+        num = Int64(9)
         ref = join('a':'z') ^ 1000
         fid = h5open(path, "w")
         # long string serialized as FixedString
         fid["longstring"] = ref
 
         # compound datatype containing a FixedString
-        compound_dtype = HDF5.Datatype(HDF5.h5t_create(HDF5.H5T_COMPOUND, sizeof(Int64) + sizeof(ref)))
-        HDF5.h5t_insert(compound_dtype, "n", 0, datatype(Int64))
-        HDF5.h5t_insert(compound_dtype, "a", sizeof(Int64), datatype(ref))
+        compound_dtype = HDF5.Datatype(HDF5.h5t_create(HDF5.H5T_COMPOUND, sizeof(num) + sizeof(ref)))
+        HDF5.h5t_insert(compound_dtype, "n", 0, datatype(num))
+        HDF5.h5t_insert(compound_dtype, "a", sizeof(num), datatype(ref))
         c = create_dataset(fid, "compoundlongstring", compound_dtype, dataspace(()))
-        # normally this is done with a `struct name; n::Int64; a::NTuple{N,Char}; end`, but
-        # we need to not actually instantiate the NTuple.
+        # normally this is done with a `struct name{N}; n::Int64; a::NTuple{N,Char}; end`,
+        # but we need to not actually instantiate the `NTuple`.
         buf = IOBuffer()
-        write(buf, Int64(9), ref)
+        write(buf, num, ref)
         @assert position(buf) == sizeof(compound_dtype)
         write_dataset(c, compound_dtype, take!(buf))
 
@@ -1093,7 +1094,7 @@ end
         T = HDF5.get_jl_type(c)
         @test T <: NamedTuple
         @test fieldnames(T) == (:n, :a)
-        @test read(c) == (n = 9, a = ref)
+        @test read(c) == (n = num, a = ref)
     end
 
     fix = HDF5.FixedArray{Float64,(2,2),4}((1, 2, 3, 4))
--- a/test/plain.jl
+++ b/test/plain.jl
@@ -1055,6 +1055,67 @@ dset = HDF5.create_external_dataset(hfile, "ext", fn_external, Int, (10,20))
 
 end
 
-# length for FixedString
-fix = HDF5.FixedString{4,0}((b"test"...,))
-@test length(fix) == 4
+@testset "FixedStrings and FixedArrays" begin
+    # properties for FixedString
+    fix = HDF5.FixedString{4,0}((b"test"...,))
+    @test length(typeof(fix)) == 4
+    @test length(fix) == 4
+    @test HDF5.pad(typeof(fix)) == 0
+    @test HDF5.pad(fix) == 0
+    # issue #742, large fixed strings are readable
+    mktemp() do path, io
+        close(io)
+        ref = join('a':'z') ^ 1000
+        fid = h5open(path, "w")
+        # long string serialized as FixedString
+        fid["longstring"] = ref
+
+        # compound datatype containing a FixedString
+        compound_dtype = HDF5.Datatype(HDF5.h5t_create(HDF5.H5T_COMPOUND, sizeof(Int64) + sizeof(ref)))
+        HDF5.h5t_insert(compound_dtype, "n", 0, datatype(Int64))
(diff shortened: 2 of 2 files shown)


kleinhenz:
wow, nice work. I'm impressed all the compound type stuff works. With this approach it should be possible to support opaque types using the same generic read machinery which would be nice to do.

There is some performance regression when reading compound types with small fixed sized array members.

```
using BenchmarkTools

using HDF5

struct foo
  a::NTuple{10, Float64}
end

function main()
  D = 10

  dtype = HDF5.Datatype(HDF5.h5t_create(HDF5.H5T_COMPOUND, sizeof(foo)))
  array_dtype = HDF5.Datatype(HDF5.h5t_array_create(datatype(Float64), 1, [D]))
  HDF5.h5t_insert(dtype, "a", fieldoffset(foo, 1), array_dtype)

  N = 5000
  v = [foo(tuple(rand(Float64, D)...)) for _ in 1:N]

  fn = "output.h5"
  h5open(fn, "w") do h5f
      dspace = dataspace(v)
      dset = create_dataset(h5f, "data", dtype, dspace)
      write_dataset(dset, dtype, v)
  end

  h5open(fn, "r") do h5f
    dset = h5f["data"]
    @btime read($dset)
  end

end

main()
```

before: `485.936 μs (10032 allocations: 1.95 MiB)`
after: `2.962 ms (40041 allocations: 3.09 MiB)`

I haven't dug deep enough to tell if this is avoidable or fundamental.

jmert:
Thanks for the test case, @kleinhenz. I've played around with the implementation a bit, and I've both simplified it (got rid most of the `unsafe_*` array handling that I'd introduced) and improved it's performance. (The biggest win is the change from the loop filling an `Array{Any}` in the `NamedTuple` normalization to a `ntuple` call.)

The benchmark times on my machine:
```
424.010 μs (10032 allocations: 1.95 MiB)  # master
3.164 ms (35041 allocations: 2.63 MiB)    # PR first commit
787.872 μs (5035 allocations: 1.18 MiB)   # PR ntuple commit
```
so surprisingly, there are now actually fewer allocations than there are on master, but it's still 1.8 times slower.

src/HDF5.jl
```diff
@@ -1272,12 +1281,33 @@ Base.getindex(parent::Union{File,Group}, r::Reference) = _deref(parent, r)
...
+end
+function _normalize_types(::Type{V}, buf::AbstractVector{UInt8}) where {T, V <: VariableArray{T}}
+    va = _typed_load(V, buf)
+    pbuf = unsafe_wrap(Array, convert(Ptr{UInt8}, va.p), (sizeof(T), Int(va.len)), own = false)
+    return normalize_types(T, pbuf)
+end
+function _normalize_types(::Type{F}, buf::AbstractVector{UInt8}) where {T, F <: FixedArray{T}}
+    return reshape(normalize_types(T, reshape(buf, sizeof(T), :)), size(F)...)
+end
+_normalize_types(::Type{Cstring}, buf::AbstractVector{UInt8}) = unsafe_string(reinterpret(Ptr{UInt8}, buf)[1])
```
musm, line 1309: Why does `pointer(buf)` not work in this case?

jmert: It probably does; I think just a left-over from as I was working my way down to a completely `Array{UInt8}`-based implementation.

musm: I'm a bit confused here. 
So on master if we wrote a matrix (say of size 2x2) of fixed strings, the buffer would be 2x2 matrix where each element would be a `Cstring` Ptr. 

Correct me if I'm wrong, in this PR, instead the buffer is a `sizeof(Cstring)` x 4 matrix. I'm lost from this point on how things are read in.

jmert: A fixed string matrix would have the data inline, yes, but that's read back through the code path `normalize_types` -> `_normalize_types(::Type{<:FixedString}, buf)` (for each element).

This method is called when the stored data is a variable-length string (which is what you get if you try to store a `Matrix` of `String`s):
```julia
julia> datatype(String["a" "b"; "c" "d"])
HDF5.Datatype: H5T_STRING {
      STRSIZE H5T_VARIABLE;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_UTF8;
      CTYPE H5T_C_S1;
   }
```

(It looks like we don't have an easy way to construct a fixed-string matrix at the moment:
```julia
julia> fix_strs = let f(x) = HDF5.FixedString{1,0}(tuple(x...))
           [f(b"a") f(b"b"); f(b"c") f(b"d")]
       end
2×2 Matrix{HDF5.FixedString{1, 0}}:
 FixedString{1, 0}((0x61,))  FixedString{1, 0}((0x62,))
 FixedString{1, 0}((0x63,))  FixedString{1, 0}((0x64,))

julia> datatype(fix_strs)
ERROR: MethodError: no method matching datatype(::Matrix{HDF5.FixedString{1, 0}})
Closest candidates are:
  datatype(::HDF5.Dataset) at /home/justin/.julia/dev/HDF5/src/HDF5.jl:988
  datatype(::HDF5.Attribute) at /home/justin/.julia/dev/HDF5/src/HDF5.jl:990
  datatype(::Union{Bool, Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8, HDF5.Reference}) at /home/justin/.julia/dev/HDF5/src/HDF5.jl:993
  ...
Stacktrace:
...
```
)

musm: @jmert Sorry, but I edited my comment almost the same time you commented.

musm: I think I follow the logic for the methods
```julia
_normalize_types(::Type{Cstring}, buf::AbstractVector{UInt8}) = unsafe_string(reinterpret(Ptr{UInt8}, buf)[1])
_normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {T <: FixedString} = unpad(String(buf), pad(T))
```
I need to go through the other methods, but I think I see how this is operating.

jmert: > Correct me if I'm wrong, in this PR, instead the buffer is a sizeof(Cstring) x 4 matrix. I'm lost from this point on how things are read in. In this example, is it the case that each column of the buffer is a pointer which in the original example would be the Cstring that would be stored directly? This is then reinterpreted read in and then resized?

Yes, that's principle I'm working with — the first dimension at any stage is bytes for a single data element, whatever the type.

These "leaf" methods (all of the `_normalize_types`) accept only an `AbstractVector{UInt8}` because they're supposed to contain only a single element to interpret.

`normalize_types` works with an `AbstractMatrix` buffer since it will produce an array of outputs, and some of the "leaf"s like for `FixedArray` and `VariableArray` elements re-dispatch through there after reshaping their buffer back to a matrix since they know how many elements they contain.

jmert:
Note to self: the following implementation of `_typed_load` is faster:
```diff
diff --git a/src/HDF5.jl b/src/HDF5.jl
index 94733f8..7e159a8 100644
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1287,6 +1287,11 @@ end
 function _typed_load(::Type{T}, buf::AbstractVector{UInt8}) where {T}
     return @inbounds reinterpret(T, buf)[1]
 end
+@inline function _typed_load(::Type{T}, buf::V) where {T, V <: Union{Vector{UInt8}, Base.FastContiguousSubArray{UInt8,1}}}
+    ref = Ref{T}()
+    GC.@preserve buf Base._memcpy!(unsafe_convert(Ptr{Cvoid}, ref), pointer(buf, 1), sizeof(T))
+    return ref[]
+end
 
 _normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {T} = _typed_load(T, buf)
 function _normalize_types(::Type{T}, buf::AbstractVector{UInt8}) where {K, T <: NamedTuple{K}}
```
```julia
julia> main(); # before
  687.034 μs (5035 allocations: 1.18 MiB)

julia> main(); # after
  421.945 μs (5035 allocations: 1.18 MiB)
```
but requires first monkey-patching `Base` to have a specialization of `pointer(::ReshapedArray, ::Integer)`:
```julia
Base.pointer(a::Base.ReshapedArray, i::Integer) = pointer(parent(a), i)
```

That patch should probably get filed upstream.

(Bonus native code comparison before and after:)
image

jmert:
Regarding my last comment, turns out the only thing stopping this from working was the use of unsigned integer range view indices — the definition of `unsafe_convert` to pointers on `SubArray`s is type-restricted to `Int` ranges, while `FastContiguousSubArray` allows for `UInt`s as well; making the types match lets us use the fast-path generically.

jmert:
I've been testing interactively with Julia master, so when I went to check older versions of Julia, I was still seeing a large performance regression. I've been able to partially mitigate that, but I doubt there's much more that can be done — I assume Julia master does so well because it can elide/optimize through the many `view`, `reinterpret`, and `reshape` operations while previous Julia versions cannot.

|         | v1.3.1                                   | v1.5.3                                   | v1.6.0-DEV.1545                          |
| ------- | ---------------------------------------- | ---------------------------------------- | ---------------------------------------- |
| master  | 431.976 μs (15034 allocations: 2.02 MiB) | 413.433 μs (10032 allocations: 1.95 MiB) | 419.412 μs (10032 allocations: 1.95 MiB) |
| 11cffb3 |   1.419 ms (95039 allocations: 4.62 MiB) |   1.188 ms (15037 allocations: 1.34 MiB) | 758.573 μs (5035 allocations: 1.18 MiB)  |
| 4e1fe08 |   1.010 ms (45039 allocations: 2.33 MiB) |   1.073 ms (15037 allocations: 1.34 MiB) | 622.548 μs (5035 allocations: 1.18 MiB)  |
| 21ccae6 | 854.999 μs (35039 allocations: 2.10 MiB) | 814.171 μs (15036 allocations: 1.34 MiB) | 419.919 μs (5034 allocations: 1.18 MiB)  |

musm:
Wow really nice, so with the latest commit it's just as fast as master and allocates far fewer. I still need to grok all the changes.

jmert:
Yeah, within factor of two or so for all released versions, and equal for upcoming Julia v1.6.

And I would argue that the apparent regression by a factor of two shouldn't be a big concern since the benchmarks are measuring the time to read when the file/disk cache is hot; under more typical circumstance, I expect the latency from loading from disk to completely dominate the total execution time.

src/HDF5.jl
```diff
@@ -1181,16 +1181,25 @@ function Base.read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
         end
     end
 
-    buf = Array{T}(undef, sz...)
-    memspace = dataspace(buf)
+    sz = Int.(sz)
+    if do_normalize(T)
+        buf = Matrix{UInt8}(undef, sizeof(T), prod(sz))
```
kleinhenz, line 1186: We do `prod` instead of just keeping all the dimensional information here?

jmert: I was motivated by trying to make sure the array comprehension didn't add too much overhead. I was hoping the `... ind in axes(buf, 2)` would optimize well across all versions of Julia, though I admit I haven't actually performance tested against `... ind in CartesianIndices(Base.tail(axes(buf)))`.

src/HDF5.jl
```diff
@@ -1370,12 +1378,50 @@ Base.getindex(parent::Union{File,Group}, r::Reference) = _deref(parent, r)
...
-normalize_types(x::FixedString) = unpad(String(collect(x.data)), pad(x))
-normalize_types(x::FixedArray) = normalize_types.(reshape(collect(x.data), size(x)...))
-normalize_types(x::VariableArray) = normalize_types.(copy(unsafe_wrap(Array, convert(Ptr{eltype(x)}, x.p), x.len, own=false)))
+function normalize_types(::Type{T}, buf::AbstractMatrix{UInt8}) where {T}
+    return [_normalize_types(T, view(buf, :, ind)) for ind in axes(buf, 2)]
+end
+
+# high-level description which should always work
+function _typed_load(::Type{T}, buf::AbstractVector{UInt8}) where {T}
+    return @inbounds reinterpret(T, buf)[1]
```
musm, line 1387: To confirm this will always be the first element, since `buf` will exactly always be  8 elements long and  T will always be an 8-byte pointer.  (at least on x64)? Maybe then just use `[]` ? 

Also, could we add comments above on what _typed_load is doing i.e. "reinterpret the vector buffer as a pointer locating the data. Note, by construction the buffer will exactly match the pointer size."

src/HDF5.jl
```diff
@@ -1370,12 +1378,50 @@ Base.getindex(parent::Union{File,Group}, r::Reference) = _deref(parent, r)
...
+end
+# fast-path for common concrete types with simple layout (which should be nearly all
+# cases)
+function _typed_load(::Type{T}, buf::V) where {T, V <: Union{Vector{UInt8}, Base.FastContiguousSubArray{UInt8,1}}}
+    ref = Ref{T}()
+    GC.@preserve buf Base._memcpy!(unsafe_convert(Ptr{Cvoid}, ref), pointer(buf, 1), sizeof(T))
```
musm, line 1393: Maybe just use `pointer(buf)` (it doesn't matter, but I don't see a reason to specifically single out the first element)

Is the `unsafe_convert` to a `Cvoid` needed to get the memory address of `ref`? 

Could you `sptr = pointer(buf)`   `dptr = unsafe_convert(Ptr{Cvoid}, ref)`
on the line above to break this up? 

I did think that `reinterpet` was fast on 1.6, but your benchmarks show that there's still some performance  on the table.

musm (commented):
Sorry, this got neglected for some time. I plan to finish review tomorrow or Monday.

test/plain.jl
```diff
@@ -1055,6 +1055,67 @@ dset = HDF5.create_external_dataset(hfile, "ext", fn_external, Int, (10,20))
...
+
+        # compound datatype containing a FixedString
+        compound_dtype = HDF5.Datatype(HDF5.h5t_create(HDF5.H5T_COMPOUND, sizeof(Int64) + sizeof(ref)))
+        HDF5.h5t_insert(compound_dtype, "n", 0, datatype(Int64))
+        HDF5.h5t_insert(compound_dtype, "a", sizeof(Int64), datatype(ref))
+        c = create_dataset(fid, "compoundlongstring", compound_dtype, dataspace(()))
+        # normally this is done with a `struct name; n::Int64; a::NTuple{N,Char}; end`, but
+        # we need to not actually instantiate the NTuple.
+        buf = IOBuffer()
+        write(buf, Int64(9), ref)
```
musm, line 1081: could you pull out the `9` value above so the variable is shared between `write` and `read`?

test/plain.jl
```diff
@@ -1055,6 +1055,67 @@ dset = HDF5.create_external_dataset(hfile, "ext", fn_external, Int, (10,20))
...
+        # long string serialized as FixedString
+        fid["longstring"] = ref
+
+        # compound datatype containing a FixedString
+        compound_dtype = HDF5.Datatype(HDF5.h5t_create(HDF5.H5T_COMPOUND, sizeof(Int64) + sizeof(ref)))
+        HDF5.h5t_insert(compound_dtype, "n", 0, datatype(Int64))
+        HDF5.h5t_insert(compound_dtype, "a", sizeof(Int64), datatype(ref))
+        c = create_dataset(fid, "compoundlongstring", compound_dtype, dataspace(()))
+        # normally this is done with a `struct name; n::Int64; a::NTuple{N,Char}; end`, but
+        # we need to not actually instantiate the NTuple.
```
musm, line 1079: What's the rationale for this. Is it to avoid instatiating very large NTuple, which would break  without this patch.

src/HDF5.jl
jmert: Yes, I can add a comment because it's not a pointer that is being looked for — "the buffer contains the bytes for exactly 1 element of an object of type `T`, so reinterpret `UInt8` vector as a length-1 array of type `T`."

src/HDF5.jl
jmert: > Maybe just use `pointer(buf)` (it doesn't matter, but I don't see a reason to specifically single out the first element)

I thought there was one of the abstract array types which didn't have a bare `pointer` method, but I can't seem to reproduce that anymore. It might have become unnecessary once I figured out the details in https://github.com/JuliaIO/HDF5.jl/pull/745#issuecomment-731631381.

> Is the `unsafe_convert` to a `Cvoid` needed to get the memory address of `ref`?

Yes, since `pointer(::RefValue)` isn't defined. It's what `ccall`s with `Ref`s lower to doing, so I think it's the canonical way to do it.

> Could you `sptr = pointer(buf)` `dptr = unsafe_convert(Ptr{Cvoid}, ref)` on the line above to break this up?

Sure.

test/plain.jl
jmert: Done.

test/plain.jl
jmert: Yes, it's to avoid instantiating large tuples, but at the Julia level, independent from trying to work around the issue here. Actually instantiating the example `name` struct would run afoul of e.g. https://github.com/JuliaLang/julia/issues/35619#issuecomment-731558911

src/HDF5.jl
jmert: >> Is the unsafe_convert to a Cvoid needed to get the memory address of ref?
>
> Yes, since pointer(::RefValue) isn't defined. It's what ccalls with Refs lower to doing, so I think it's the canonical way to do it.

Oh, it's also the fastest way to get the correct pointer type. The `memcpy` ccall is defined for `Ptr{UInt8}`, which most `Ref{T}` types won't "decay to" via automatic ccall conversion, but since `Ptr{Cvoid}` is a valid pointer type for any `Ref{T}` and automatically converts to `Ptr{UInt8}` in the ccall, this ends up being the shortest way to get what we need.

src/HDF5.jl
```diff
@@ -1370,12 +1378,56 @@ Base.getindex(parent::Union{File,Group}, r::Reference) = _deref(parent, r)
...
+        return normalize_types(T, pbuf)
+    else
+        return copy(vec(reinterpret(T, pbuf)))
+    end
+end
+function _normalize_types(::Type{F}, buf::AbstractVector{UInt8}) where {T, F <: FixedArray{T}}
+    if do_normalize(T)
+        return reshape(normalize_types(T, reshape(buf, sizeof(T), :)), size(F)...)
+    else
+        return copy(reshape(reinterpret(T, buf), size(F)...))
```
musm, line 1426: Is this `copy` defensive? Doesn't the reshape in this instance  copy the buffer anyways? Same question in the directly above method.

src/HDF5.jl
musm: Yeah, I'm guessing it's just too cumbersome to keep the dimension information. Is the thinking that the compilation time would be affected due to specialization on the dimension size in the `normalize_types` method? The Matrix keeps things simple in this stage, with the only negative that it requires a final reshape. 

I think it would be a good idea to add a comment above this line along the lines:
```
#  the data is stored in a buffer matrix where the first dimension at any stage are the bytes for a single data element
and the second-dimension of the matrix stores each element 
```

src/HDF5.jl
jmert: The `copy` is to (1) make the returned array independent of the original, parent buffer in `read()` since some of the wrapper types will just keep working with a view into the original buffer, and (2) it gets rid of possibly complex nested types:

For instance:
```julia
buf = reshape(collect(UInt8(1):UInt8(64)), 32, 2) # in read()
compound = view(buf, :, 1)  # in normalize_types()
pair = view(compound, 1:16) # in _normalize_types(::Type{NamedTuple})
el = reshape(reinterpret(UInt32, pair), (2, 2)) # in _normalize_types(::Type{FixedArray})
```
```julia
julia> typeof(el)
Base.ReshapedArray{UInt32, 2, Base.ReinterpretArray{UInt32, 1, UInt8, SubArray{UInt8, 1, Matrix{UInt8}, Tuple{UnitRange{Int64}, Int64}, true}, false}, Tuple{}}

julia> typeof(copy(el))
Matrix{UInt32} (alias for Array{UInt32, 2})
```
After potentially arbitrary levels of nesting, I think `copy`ing is the only way to guarantee you don't have arbitrarily complex types being returned.

src/HDF5.jl
jmert: > Is the thinking that the compilation time would be affected due to specialization on the dimension size in the normalize_types method? The Matrix keeps things simple in this stage, with the only negative that it requires a final reshape.

Yes, that was exactly my thinking. I have to admit I haven't actually extensively tested this assumption, though.

> I think it would be a good idea to add a comment above this line along the lines:

Sure, I'll get that added tomorrow.

src/HDF5.jl
```diff
@@ -1370,12 +1378,56 @@ Base.getindex(parent::Union{File,Group}, r::Reference) = _deref(parent, r)
...
+        elT = fieldtype(T, ii)
+        off = fieldoffset(T, ii) % Int
+        sub = view(buf, off .+ (1:sizeof(elT)))
+        return _normalize_types(elT, sub)
+    end
+    return NamedTuple{K}(nv)
+end
+function _normalize_types(::Type{V}, buf::AbstractVector{UInt8}) where {T, V <: VariableArray{T}}
+    va = _typed_load(V, buf)
+    pbuf = unsafe_wrap(Array, convert(Ptr{UInt8}, va.p), (sizeof(T), Int(va.len)), own = false)
```
musm, line 1415:
```suggestion
    pbuf = unsafe_wrap(Array, convert(Ptr{UInt8}, va.p), (sizeof(T), Int(va.len)))
```

musm:
Ok I'm impressed this all works so well. It takes some mental gymnastics to follow the chain with various types and if they recurse, but it looks good to me.

LGTM to merge.

musm:
I wonder why Windows this fails to optimize `_typed_load`  as on Linux
 
Windows:
image


Linux:
image

src/HDF5.jl
```diff
@@ -1370,12 +1378,56 @@ Base.getindex(parent::Union{File,Group}, r::Reference) = _deref(parent, r)
...
+end
+# fast-path for common concrete types with simple layout (which should be nearly all
+# cases)
+function _typed_load(::Type{T}, buf::V) where {T, V <: Union{Vector{UInt8}, Base.FastContiguousSubArray{UInt8,1}}}
+    dest = Ref{T}()
+    GC.@preserve buf begin
+        srcptr  = pointer(buf)
+        destptr = unsafe_convert(Ptr{Cvoid}, dest)
+        Base._memcpy!(destptr, srcptr, sizeof(T))
+    end
```
musm, line 1399:
```suggestion
    GC.@preserve buf Base._memcpy!(unsafe_convert(Ptr{Cvoid}, dest), pointer(buf), sizeof(T))
```
Sorry about this, but feel free to change this back

src/HDF5.jl
```diff
@@ -1370,12 +1378,56 @@ Base.getindex(parent::Union{File,Group}, r::Reference) = _deref(parent, r)
...
+end
+
+# high-level description which should always work --- here, the buffer contains the bytes
+# for exactly 1 element of an object of type T, so reinterpret the `UInt8` vector as a
+# length-1 array of type `T` and extract the (only) element.
+function _typed_load(::Type{T}, buf::AbstractVector{UInt8}) where {T}
+    return @inbounds reinterpret(T, buf)[1]
+end
+# fast-path for common concrete types with simple layout (which should be nearly all
+# cases)
```
musm, line 1392:
```suggestion
# fast-path for common concrete types with simple layout (which should be nearly all cases)
```

src/HDF5.jl
jmert: I think this needs to stay as `own = false` because the destination of the pointer is owned by libhdf5 and freed later by the call to `h5d_vlen_reclaim`. (That's the reason the `copy` is required in the second branch of the following `if` statement).

@kleinhenz, can you confirm?

jmert:
> I wonder why Windows this fails to optimize `_typed_load` as on Linux

My reading of the IR is that LLVM isn't inlining `memcpy` into simple machine code and is instead still calling a library function to do the work.

src/HDF5.jl
```diff
@@ -1370,12 +1381,65 @@ Base.getindex(parent::Union{File,Group}, r::Reference) = _deref(parent, r)
...
+# for exactly 1 element of an object of type T, so reinterpret the `UInt8` vector as a
+# length-1 array of type `T` and extract the (only) element.
+function _typed_load(::Type{T}, buf::AbstractVector{UInt8}) where {T}
+    return @inbounds reinterpret(T, buf)[1]
+end
+# fast-path for common concrete types with simple layout (which should be nearly all cases)
+function _typed_load(::Type{T}, buf::V) where {T, V <: Union{Vector{UInt8}, Base.FastContiguousSubArray{UInt8,1}}}
+    dest = Ref{T}()
+    GC.@preserve buf Base._memcpy!(unsafe_convert(Ptr{Cvoid}, dest), pointer(buf), sizeof(T))
+    return dest[]
```
jmert, line 1400:
```suggestion
    return GC.@preserve buf unsafe_load(convert(Ptr{T}, pointer(buf)))
```

@musm, does this optimize better one Windows? I see the same exact LLVM IR on Linux.

This is probably also simpler to read — I was motivated to use the `_memcpy!` version after digging through `ReinterpretArray` internals, but for this case where we're only loading a single element, `unsafe_load` is probably the better choice anyway.

jmert: Actually, on second thought, there might be a semantic difference — x86 is forgiving about unaligned loads (so loading 8-byte `Int` from an address where address mod 8 != 0), but that's not true on all architectures. IIUC, `memcpy` is defined to work correctly no matter the alignment, but doing so might take several instructions if loads must be aligned.

I _think_ this shows that: https://godbolt.org/z/snvrMM

src/HDF5.jl
musm: I was going off the documentation that `own=false` is the default for the `unsafe_wrap` method.

jmert: Ah, didn't realize that.

src/HDF5.jl
musm: This does optimize cleanly:

image

where `__untyped_load` is the modified one above.

jmert: Do you find something like
```ll
call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 1 %20, i8* %17, i64 8, i1 false)
```
in the unoptimized LLVM IR? I'm assuming yes, so Julia is giving LLVM the info it needs, but then LLVM isn't optimizing it away.

musm: Yes that's precisely what I find. This isn't anything blocking, but I was curious why it doesn't optimize.

musm:
@jmert thanks for doing this and adding the extra comments which will be helpful for future reference.

musm (approved):

src/HDF5.jl
musm: @jmert I ran the test suite on x86 on Windows with the modified method:
```julia
function _typed_load(::Type{T}, buf::V) where {T, V <: Union{Vector{UInt8}, Base.FastContiguousSubArray{UInt8,1}}}
    return GC.@preserve buf unsafe_load(convert(Ptr{T}, pointer(buf)))
end
```
and the suite passed.

I can't comment on which is technically correct.

musm: My understanding was that Julia protects us from this:

> Int from an address where address mod 8 != 0)

https://docs.julialang.org/en/v1/devdocs/object/

at the end of the document

> Note that all objects are allocated in multiples of 4 bytes and aligned to the platform pointer size. Memory is allocated from a pool for smaller objects, or directly with malloc() for large objects.

jmert: Once we've recursed into something like `_normalize_types(::Type{<:NamedTuple})` where we're taking arbitrary `view`s, Julia's allocation guarantee no longer applies to the alignment of any particular pointer we take.

musm: I think as long as the order is correct it shouldn't be an issue since Julia's member alignment is the same as C semantics.

jmert: OK — how about using the `unsafe_load` version, with the `_memcpy!` variant commented out and waiting just in case it has to be resurrected?

musm: I'm on board with that if you are?  If we get bug reports on the next release, we can quickly release a fix, but the test suit is passing for me.

> x86 is forgiving about unaligned loads (so loading 8-byte Int from an address where address mod 8 != 0), but that's not true on all architectures. IIUC, memcpy is defined to work correctly no matter the alignment, but doing so might take several instructions if loads must be aligned.

> I think this shows that: https://godbolt.org/z/snvrMM

I'm not 100% confident that this won't be an issue, but I'm inclinced to think it isn't.

musm: If we are very concerend, It might be worth opening a discourse question to confirm these assumptions and see what others say.

jmert:
I'm going to merge with the more conservative option; I've added a TODO commenting that a simpler command might be possible if we can be sure alignment will be correct in all situations, and that can happen in future work (if needed).

PR #747, opened 2020-11-22
WIP: Support reading opaque data recursively

jmert:
This builds on top of the changes in #745.

Most of the added code is just tests of reading and writing opaque objects, and I've simplified the special-cased `read(::AttributeOrDataset, ::Type{Opaque}` implementation just a bit.

The interesting change is to `get_mem_compatible_jl_type`  — a recursively-reached opaque type is reported as having a `FixedArray{UInt8}` layout. This allows for the type-normalization code to automatically handle reading in opaque data embedded in other data types. The downside is that the opaque tag information is lost. (Note that this doesn't affect a "top level" opaque attribute or dataset which will still load into a `HDF5.Opaque` object which contains both the data and the tag.)

For instance, the one new example use in the tests reads in the HDF5 datatype:
```
HDF5.Datatype: H5T_COMPOUND {
      H5T_STD_I64LE "v" : 0;
      H5T_OPAQUE {
         OPQ_SIZE 4;
         OPQ_TAG "opaque test";
      } "d" : 8;
   }
```
as the named tuple `(v = 1, d = rand(UInt8, 4))`.

One minor thing I noticed while writing the tests is that an opaque scalar object is returned as a 0-dimensional array of a vector of bytes (i.e. `fill(UInt8[...])`). The generic `read` function takes note of scalar objects and dereferences the container array upon return, but the opaque read doesn't. I left that as-is (and the tests check for that case now), but that'd be easy to change in the opaque-specific read function.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1329,28 +1329,24 @@ end
 
 # Read OPAQUE datasets and attributes
 function Base.read(obj::DatasetOrAttribute, ::Type{Opaque})
-    local buf
-    local len
-    local tag
-    sz = size(obj)
     obj_type = datatype(obj)
-    try
-        len = h5t_get_size(obj_type)
-        buf = Vector{UInt8}(undef,prod(sz)*len)
-        tag = h5t_get_tag(obj_type)
-        if obj isa Dataset
-            read_dataset(obj, obj_type, buf)
-        else
-            read_attribute(obj, obj_type, buf)
-        end
-    finally
-        close(obj_type)
+    sz  = size(obj)
+    buf = Matrix{UInt8}(undef, sizeof(obj_type), prod(sz))
+    if obj isa Dataset
+        read_dataset(obj, obj_type, buf, obj.xfer)
+    else
+        read_attribute(obj, obj_type, buf)
     end
-    data = Array{Array{UInt8}}(undef,sz)
-    for i = 1:prod(sz)
-        data[i] = buf[(i-1)*len+1:i*len]
+    tag = h5t_get_tag(obj_type)
+    close(obj_type)
+    if isempty(sz)
+        # scalar (only) result
+        data = vec(buf)
+    else
+        # array of opaque objects
+        data = reshape([buf[:,i] for i in 1:prod(sz)], sz...)
     end
-    Opaque(data, tag)
+    return Opaque(data, tag)
 end
 
 # Array constructor for datasets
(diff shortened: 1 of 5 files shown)


src/HDF5.jl
```diff
@@ -1181,16 +1181,25 @@ function Base.read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
         end
     end
 
-    buf = Array{T}(undef, sz...)
-    memspace = dataspace(buf)
+    sz = Int.(sz)
```
musm, line 1184: rebase and get rid of this?

jmert: Yeah, I can do that. I'll get this and #745 rebased.

src/HDF5.jl
```diff
@@ -1329,28 +1329,18 @@ end
...
+    buf = Matrix{UInt8}(undef, sizeof(obj_type), prod(sz))
+    if obj isa Dataset
+        read_dataset(obj, obj_type, buf, obj.xfer)
+    else
+        read_attribute(obj, obj_type, buf)
     end
-    Opaque(data, tag)
+    tag = h5t_get_tag(obj_type)
+    close(obj_type)
+    data = [buf[:,i] for (i, _) in enumerate(CartesianIndices(sz))]
```
musm, line 1342: `for i in prod(sz)` ?

jmert: `enumerate(CartesianIndices(sz))` has a shape while `prod(sz)` doesn't:

```julia
julia> [i for (i, _) in enumerate(CartesianIndices((2,2)))]
2×2 Matrix{Int64}:
 1  3
 2  4

julia> [i for i in 1:prod((2,2))]
4-element Vector{Int64}:
 1
 2
 3
 4
```

jmert: I could make it `reshape([buf[:,i] for i in 1:prod(sz)], sz...)`, though.

musm: Oh my bad,  I was looking at the other PR too long and automatically assumed this should be `prod(sz)` without spending much time looking at the details.

musm: > I could make it reshape([buf[:,i] for i in 1:prod(sz)], sz...), though.

Perhaps one argument for this is that it matches `reshape` we do for `read`

musm:
> The generic read function takes note of scalar objects and dereferences the container array upon return, but the opaque read doesn't. I left that as-is (and the tests check for that case now), but that'd be easy to change in the opaque-specific read function.

I think we should dereference them as well to be consistent with read.

musm (approved):

musm:
Eventually, sounds like we should be able to merge  `get_jl_type` into  `get_mem_compatible_jl_type` , with some restructing to preserve the tag and perhaps requiring a custom `_normalize_types` to extract of the data.

musm:
Feel free to merge for now, as we iterate.

jmert:
> I think we should dereference them as well to be consistent with read.

I just changed that, and when CI passes, I'll merge.

PR #748, opened 2020-11-23
Add custom printing for Dataspaces

jmert:
Exploring datasets interactively is a lot easier when you can quickly get feedback on the shapes of dataspaces, so it was annoying that all we had printing was the `hid_t` value.

The tests are a good summary of what this custom printing adds:
```julia
dspace_null = HDF5.Dataspace(HDF5.h5s_create(HDF5.H5S_NULL))
dspace_scal = HDF5.Dataspace(HDF5.h5s_create(HDF5.H5S_SCALAR))
dspace_norm = dataspace((100, 4))
dspace_maxd = dataspace((100, 4), max_dims = (256, 4))
dspace_slab = HDF5.hyperslab(dataspace((100, 4)), 1:20:100, 1:2:4)
dspace_irrg = HDF5.Dataspace(HDF5.h5s_combine_select(
        HDF5.h5s_copy(dspace_slab), HDF5.H5S_SELECT_OR,
        HDF5.hyperslab(dataspace((100, 4)), 2, 2)))

@test sprint(show, dspace_null) == "HDF5.Dataspace: H5S_NULL"
@test sprint(show, dspace_scal) == "HDF5.Dataspace: H5S_SCALAR"
@test sprint(show, dspace_norm) == "HDF5.Dataspace: (100, 4)"
@test sprint(show, dspace_maxd) == "HDF5.Dataspace: (100, 4) / (256, 4)"
@test sprint(show, dspace_slab) == "HDF5.Dataspace: (1:20:81, 1:2:3) / (1:100, 1:4)"
@test sprint(show, dspace_irrg) == "HDF5.Dataspace: (100, 4) [irregular selection]"
```

### DIFF
--- a/docs/src/api_bindings.md
+++ b/docs/src/api_bindings.md
@@ -189,14 +189,18 @@ h5r_get_region(loc_id::hid_t, ref_type::Cint, ref::Ptr{Cvoid})
 ## [`H5S`](https://portal.hdfgroup.org/display/HDF5/Dataspaces) — Dataspace Interface
 ```julia
 h5s_close(space_id::hid_t)
+h5s_combine_select(space1_id::hid_t, op::Cint, space2_id::hid_t)
 h5s_copy(space_id::hid_t)
 h5s_create(class::Cint)
 h5s_create_simple(rank::Cint, current_dims::Ptr{hsize_t}, maximum_dims::Ptr{hsize_t})
 h5s_get_regular_hyperslab(space_id::hid_t, start::Ptr{hsize_t}, stride::Ptr{hsize_t}, count::Ptr{hsize_t}, block::Ptr{hsize_t})
+h5s_get_select_hyper_nblocks(space_id::hid_t)
 h5s_get_select_npoints(space_id::hid_t)
+h5s_get_select_type(space_id::hid_t)
 h5s_get_simple_extent_dims(space_id::hid_t, dims::Ptr{hsize_t}, maxdims::Ptr{hsize_t})
 h5s_get_simple_extent_ndims(space_id::hid_t)
 h5s_get_simple_extent_type(space_id::hid_t)
+h5s_is_regular_hyperslab(space_id::hid_t)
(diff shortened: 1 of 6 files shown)


src/api_helpers.jl
```diff
@@ -237,6 +237,16 @@ end
 ### Dataspace Interface
 ###
 
+function h5s_get_regular_hyperslab(space_id)
```
musm, line 240: Working on https://github.com/JuliaIO/HDF5.jl/pull/749  perhaps we should do the conversions to `Int` in the `api_helpers` interface?  

It's quite messy to have to wrap or truncate constants to `Int` at the highest levels of the API, and makes interacting with the low-level functions more annoying.

musm: Looking again basically this is similar to `https://github.com/JuliaIO/HDF5.jl/pull/749` where if you want to practically use the output you have to reverse the dimensions and truncate the data to `Int`, which makes me think we should just do it in the intermediate level API instead so that the display function has to do less wrangling?

musm:
LGTM !

I feel like we might need to add some text before printing the size that describes what these numbers actually mean to help users?  Altough it's not clear to me what's the most appropriate text to show to strik the right balance between being too wordy and poluting the diplay and not providng enough information.

src/api_helpers.jl
musm: Perhaps the best option would be to add a higher level wrapper `get_simple_extent_dims` similar to `get_dims` that handles all the data conversion (`get_dims` probably should be rename to `get_simple_extent_dims` ) ?

musm:
That works with me. I think it's quite reasonable as is and the explanation makes sense. In any case, I couldn't think of anything better.

src/api_helpers.jl
jmert: How about making possibly adding a `get_regular_hyperslab` part of #749? That way this PR can just be a simple "add show" and doesn't mess with possibly more holistic API changes.

PR #749, opened 2020-11-23
Optimize get_dims and truncate return type to Int

musm:
This is about 4x faster, plus allocates fewer memory and fixes the annoying return type when used in e.g. https://github.com/JuliaIO/HDF5.jl/pull/745


src/api_helpers.jl
```diff
@@ -312,4 +312,3 @@ end
 ###
 ### Filter Interface
 ###
-
```
musm, line 315: my editor automatically does this, it's harmless here.

src/api_helpers.jl
```diff
@@ -242,7 +242,7 @@ function h5s_get_simple_extent_dims(space_id)
     dims = Vector{hsize_t}(undef, n)
     maxdims = Vector{hsize_t}(undef, n)
     h5s_get_simple_extent_dims(space_id, dims, maxdims)
-    return dims, maxdims
+    return n, dims, maxdims
```
musm, line 245: @jmert It's not a huge performance hit to calcluate `n` again in `get_dims`, but since we have this information I figured it's best to just use it.

musm: I recall we might have discussed returning `n` a while ago when you wrote these `api_helpers`. Happy to keep as is and calculate `N = length(h5_dims)`, if you prefer otherwise.

musm:
The only annoying aspect is now comparisons to `H5S_UNLIMITED` require type truncation, since we are comparing against an UInt.

src/HDF5.jl
```diff
@@ -1087,8 +1087,11 @@ dataspace(sz1::Int, sz2::Int, sz3::Int...; max_dims::Union{Dims,Tuple{}}=()) = _
 
 
 function get_dims(dspace::Dataspace)
-    dims = h5s_get_simple_extent_dims(dspace)
-    return tuple(reverse!(dims[1])...), tuple(reverse!(dims[2])...)
+    N, h5_dims, h5_maxdims = h5s_get_simple_extent_dims(dspace)
+    # reverse dimensions since hdf5 uses C-style order
+    dims = ntuple(i -> h5_dims[N-i+1] .% Int, N)
```
musm, line 1092: https://support.hdfgroup.org/HDF5/faq/limits.html#:~:text=There%20are%20no%20limitations%20on%20the%20file%20size.,datatypes%20or%20compound%20datatypes%20that%20are%20very%20large 

Makes me  uncomfortableabout the truncation

musm: I guess the face that `dataspace` only accepts `Dims{N}`, makes me more comfortable that this is fine. Albiet, there might me some edge case with esoteric HDF5 files with really large dimensions don't work with HDF5.jl. (probably something else to look into, because I'm assuming there would be more than this as the only issue).

src/HDF5.jl
```diff
@@ -1087,8 +1087,22 @@ dataspace(sz1::Int, sz2::Int, sz3::Int...; max_dims::Union{Dims,Tuple{}}=()) = _
 
 
 function get_dims(dspace::Dataspace)
-    dims = h5s_get_simple_extent_dims(dspace)
-    return tuple(reverse!(dims[1])...), tuple(reverse!(dims[2])...)
+    h5_dims, h5_maxdims = h5s_get_simple_extent_dims(dspace)
+    # reverse dimensions since hdf5 uses C-style order
+    N = length(h5_dims)
+    dims = ntuple(i -> Int(h5_dims[N-i+1]), N)
+    maxdims = ntuple(i -> h5_maxdims[N-i+1] % Int, N) # allows max_dims to be specified as -1 without triggering an overflow
```
jmert, line 1094:
```suggestion
    dims = ntuple(i -> @inbound Int(h5_dims[N-i+1]), N)
    maxdims = ntuple(i -> @inbounds h5_maxdims[N-i+1] % Int, N) # allows max_dims to be specified as -1 without triggering an overflow
```

I see this reducing the LLVM and native code line counts by about half.

jmert, line 1094:
```suggestion
    dims = ntuple(i -> @inbounds(Int(h5_dims[N-i+1])), N)
    maxdims = ntuple(i -> @inbounds(h5_maxdims[N-i+1]) % Int, N) # allows max_dims to be specified as -1 without triggering an overflow
```
Doing this, I get ~ 15% / 20% reduction in LLVM / machine code lengths.

src/HDF5.jl
```diff
@@ -1087,8 +1087,22 @@ dataspace(sz1::Int, sz2::Int, sz3::Int...; max_dims::Union{Dims,Tuple{}}=()) = _
...
+end
+
+function get_regular_hyperslab(dspace::Dataspace)
+    h5_start, h5_stride, h5_count, h5_block = h5s_get_regular_hyperslab(dspace)
+    N = length(h5_start)
+    start = ntuple(i -> Int(h5_start[N-i+1]), N)
+    stride = ntuple(i -> Int(h5_stride[N-i+1]), N)
+    count = ntuple(i -> Int(h5_count[N-i+1]), N)
+    block = ntuple(i -> Int(h5_block[N-i+1]), N)
+    return start, stride, count, block
```
jmert, line 1105:
```suggestion
    @inline rev(v) = ntuple(i -> @inbounds(Int(v[N-i+1])), N)
    return rev(h5_start), rev(h5_stride), rev(h5_count), rev(h5_block)
```
Similar here, with ~ 35% reduction in both LLVM / native.

jmert (approved):

src/api_helpers.jl
jmert: Yeah, my hesitation to returning it was/is that it deviates from the simple rule that the low-level API either follows the HDF5 docs (with the two auto-changes being `htri_t` -> throw on error or `Bool` and `herr_t` -> throw on error or `nothing`) or returns the documented out-args for the signature lacking those arguments.

I'm completely open to making the change here and elsewhere if the rule/transformation can be as clearly enumerated. (out args plus query-call return?)

src/HDF5.jl
jmert: Yeah, I haven't been worrying about it because it _seems_ like all of Julia's array constructors cast all sizes to `Int` before calling into the C runtime. I suspect there would be fundamental issues with working with arrays larger than `typemax(Int)` axes, so it's not likely this would be the single point of failure.

src/api_helpers.jl
musm: > Yeah, my hesitation to returning it was/is that it deviates from the simple rule that the low-level API either follows the HDF5 docs

Yeah, that's why I backtracked on these changes after thinking about it a little longer.  At times I felt like it might be nice to veer of this, but consistency is more important. 

I'm really liking the 3-levels of abstractions we currently have. (1) low-level, (2) convenience low-level wrappers and (3) high-level wrappers in the main source file.

I'm also thinking we should consolidate the naming of this function and other here to adhere to `h5s_get_simple_extent_dims` -> `get_simple_extent_dims`. I.e. at the highest level, remove the `h5(*)_` prefix.

PR #750, opened 2020-11-24
Switch from defining `Base.convert` to `Base.cconvert` on HDF5 objects

jmert:
By defining `Base.convert(::Type{hid_t}, x)` for all of the high-level
object types, implicit conversion to an integer is permitted.
This has the unfortunate consquence --- as found in JuliaIO/MAT.jl#145
and JuliaIO/MAT.jl#151 --- that incorrectly "wrapping" a wrapped object
was allowed, but doing so allows the garbage collector to reap the
object too early.

The main reason for allowing conversion from an object to integer
(namely `HDF5.hid_t`) is to support passing the objects to `ccall`s
without requiring reaching into the `obj.id` raw integer ID.

We can easily still allow this while disabling implicit conversion by
just replacing all definitions of `Base.convert` with `Base.cconvert`.
Doing this would have caught the mistake in MAT.jl at the Julia type
level rather than having to debug a runtime error.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -189,7 +189,7 @@ mutable struct File <: H5DataStore
         f
     end
 end
-Base.convert(::Type{hid_t}, f::File) = f.id
+Base.cconvert(::Type{hid_t}, f::File) = f.id
 
 mutable struct Group <: H5DataStore
     id::hid_t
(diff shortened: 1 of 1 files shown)


jmert:
Ah, the MAT.jl integration test now shows the mistake this is guarding against nicely :-) (until a new MAT.jl is tagged with the bug fix)

musm:
Totally agreed, this is the point of `cconvert`, surprised we didin't  notice this earlier.

jmert:
I didn't actually read that CI error closely enough — it says `Dataset` and not `Datatype`. Running MAT.jl tests locally with this PR, there's (at least) another case to be fixed yet.

musm:
Nice MAT.jl tests are finally green.

Issue #752, opened 2020-11-30
Performance regression in h5read


laborg:
My system:
```
julia> versioninfo()
Julia Version 1.5.3
Commit 788b2c77c1 (2020-11-09 13:37 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
```

On 0.13.7:
```
@btime h5read("3bc6206c-bad8-43ba-9642-b2e06c4c7cd4.h5","acc", (1,1))
  117.233 μs (53 allocations: 2.16 KiB)
-0.0027361426f0
```

on 14.1
```
@btime h5read("3bc6206c-bad8-43ba-9642-b2e06c4c7cd4.h5", "acc", (1,1))
  1.799 ms (31 allocations: 1.22 KiB)
-0.0027361426f0
```

Any way to work around this, besides downgrading?

musm:
Dupe of https://github.com/JuliaIO/HDF5.jl/issues/742.  We have outstanding PRs that address these performance issues, should be fixed in the next week or so.

musm:
Best bet is to wait if you can.

jmert:
Is this a duplicate? The output suggests this is just reading a single Float32 value, while PR #747 is all about avoiding the (ab)use of large tuples in fixed-length arrays/compound data types. I _think_ this case should be independent of that.

@laborg — the one thing I'd be interested in seeing further is whether you can show the same regression on a bare `read` rather than `h5read` which also must open and close the file on each invocation. That's to say, if there's been a regression in `h5open`/`close` time (maybe from switching to libhdf5 v1.12), the obvious suggestion is to open and hold a handle to the open file rather than opening and closing often.

musm:
I'll leave this open until, we have feedback from OP.

laborg:
An MWE, taken from the HDF5 docs, shown on 0.13.7 and master:

```
# create data set
f = tempname()
h5write(f, "mygroup2/A", collect(reshape(1:120, 15, 8)))
```

### 0.13.7
```
(jl_KT5sZW) pkg> st
Status `/tmp/jl_KT5sZW/Project.toml`
  [f67ccb44] HDF5 v0.13.7

[...]

PR #753, opened 2020-12-01
Avoid creating default properties unnecessarily

jmert:
When there are no keyword arguments, just use the global
`DEFAULT_PROPERTIES` rather than creating new `HDF5.Properties` with
default values.

This improves the regression identified in issue #752.
Specifically, the benchmark used in https://github.com/JuliaIO/HDF5.jl/issues/752#issuecomment-736676421 is improved from
```julia
julia> @‌btime fid["mygroup/A"];
  13.872 μs (3 allocations: 96 bytes)
```
on master to
```julia
julia> @‌btime fid["mygroup/A"];
  6.778 μs (1 allocation: 32 bytes)
```
with this commit.


musm:
3 hour queu to run CI.... huh

PR #754, opened 2020-12-03
Close objects in `h5*` immediate functions

jmert:
The try-catch blocks make sure to close the file, but we might as well close the objects themselves deterministically rather than waiting for the GC and/or libhdf5's `H5F_CLOSE_STRONG` to reap them. This should also be a net performance improvement in `h5writeattr` even ignoring the subtler `close` issue just due to opening the parent object only once.

This was discovered as a way to improve `h5read` performance in https://github.com/JuliaIO/HDF5.jl/issues/752#issuecomment-737705125 and https://github.com/JuliaIO/HDF5.jl/issues/752#issuecomment-737809048.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -536,6 +536,7 @@ function h5read(filename, name::AbstractString; pv...)
     try
         obj = getindex(fid, name; pv...)
         dat = read(obj)
+        close(obj)
     finally
         close(fid)
     end
@@ -548,6 +549,7 @@ function h5read(filename, name_type_pair::Pair{<:AbstractString,DataType}; pv...
     try
         obj = getindex(fid, name_type_pair[1]; pv...)
         dat = read(obj, name_type_pair[2])
+        close(obj)
     finally
         close(fid)
     end
@@ -560,6 +562,7 @@ function h5read(filename, name::AbstractString, indices::Tuple{Vararg{Union{Abst
     try
         dset = getindex(fid, name; pv...)
         dat = dset[indices...]
+        close(dset)
     finally
         close(fid)
     end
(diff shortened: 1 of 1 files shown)


jmert:
Failures on Julia nightly are unrelated to this PR and are an upstream problem.

musm:
Makes sense.

musm:
Are we missing a close on `https://github.com/JuliaIO/HDF5.jl/blob/master/src/HDF5.jl#L1452` dtype as well?

src/HDF5.jl
```diff
@@ -536,6 +536,7 @@ function h5read(filename, name::AbstractString; pv...)
     try
         obj = getindex(fid, name; pv...)
         dat = read(obj)
+        close(obj)
```
musm, line 539: should these actually be in the `finally` block?

jmert: Technically, they might need a set of nested try-catch blocks: you can't close `obj` if `getindex` failed, but then you do want to close it if `read` fails.

jmert: Though allowing `obj` to escape the try-catch block if `read` fails is basically what the status quo was — GC or libhdf5's close semantics should eventually take care of the object.

musm: I'm not 100% confident we are following the correct close semantics in all our usages of the `try`--`finally` blocks, many of them are legacy, but I believe are mostly correct. Ideally, we could minimize their usage to only essential places, since `try` `finally` blocks are slow calls.

jmert: Yeah, I haven't thought hard about the `try`-`finally` use yet, but IIUC they're most essential in cases like this were a file handle is being opened and closed. I agree in trying to see if more uses can be removed elsewhere, though.

musm: > Though allowing obj to escape the try-catch block if read fails is basically what the status quo was — GC or libhdf5's close semantics should eventually take care of the object.

If `read` fails,  `close(obj)` won't execute, and then we fall back to closing via libhdf5's or GC's close semantics, which were demonstrated to cause some regressions. This is an edge case and so the performance isn't critical.

jmert: Sorry, yeah, I guess I only implied it, but that's what I meant.

PR #755, opened 2020-12-03
Limit HDF5_jll compatible versions

jmert:
HDF5_jll follows the versioning scheme of libhdf5 which uses the second set of digits as their "major" versions. Set the compatibility to agree with this use of versioning so that HDF5.jl doesn't try to pull in a newer library (which may be incompatible) without being explicitly allowed to in the future.

This was the issue on the HDF5.jl v0.13.6 release which required #743 and a subsequent patch release as v0.13.7 to fix. We should probably be more proactive about avoiding a similar issue.


musm:
Good idea

PR #758, opened 2020-12-04
Add `size`, `ndims`, `length`, etc methods to `Dataspace`

jmert:
This adds size-related methods to `Dataspace` and has the corresponding `Dataset` and `Attribute` methods call through these instead. There is also a new variant of the helper `h5s_get_simple_extent_dims` method which returns only the size and not max dims as a way to remove an allocation of a vector (and tuple-reversal) where it won't be needed. This is a small part of my quest to deal with #752 — times aren't changed much (within the noise, I think), but allocation count goes down:
image
(I took the opportunity to also add `copy`, `==`, and `hash` just to round out some common interactions.)

The bulk of this PR ends up being the tests. As far as I can tell, testing aspects of `Dataspace` is completely through implicit use in other operations, so this adds a new file which explicitly tests the interface. This is actually a model I'd like to see repeated for other parts of the API (cleaning up the very large `test/plain.jl` in the process).


src/HDF5.jl
```diff
@@ -1094,6 +1112,28 @@ dataspace(sz::Dims{N}; max_dims::Union{Dims{N},Tuple{}}=()) where {N} = _dataspa
...
+    h5_dims = h5s_get_simple_extent_dims(dspace, nothing)
+    return @inbounds Int(h5_dims[N - d + 1])
+end
+function Base.length(dspace::Dataspace)
+    checkvalid(dspace)
+    h5_dims = h5s_get_simple_extent_dims(dspace, nothing)
+    return Int(prod(h5_dims))
+end
+
 function get_dims(dspace::Dataspace)
```
musm, line 1137: We should figure out a way to get rid of this and add it as a method of `size` or at least rename it

jmert: Something like `size(obj; max = true)` or `size(obj, :max)` (for any spelling of `max`) to return both the size and max dims?

src/api_helpers.jl
```diff
@@ -254,6 +254,12 @@ function h5s_get_simple_extent_dims(space_id)
     h5s_get_simple_extent_dims(space_id, dims, maxdims)
     return dims, maxdims
 end
+function h5s_get_simple_extent_dims(space_id, ::Nothing)
```
musm, line 257: I don't think there's a good way to avoid this right?

Doing something like:
```julia
function h5s_get_simple_extent_dims(space_id; maxdims = false)
           n = h5s_get_simple_extent_ndims(space_id)
           dims = Vector{hsize_t}(undef, n)
           if maxdims
               mdims = Vector{hsize_t}(undef, n)
               h5s_get_simple_extent_dims(space_id, dims, mdims)
               return dims, mdims
           else
               h5s_get_simple_extent_dims(space_id, dims, C_NULL)
               return dims
           end
end
```
would make the return type unstable.

jmert: Yeah, that's why I just went with a separate method. I went with `nothing` argument since `C_NULL = Ptr{Nothing}(0)`, so _(vigorous hand waving)_ they're kind of similar-ish.

src/HDF5.jl
musm: Yeah that would be ideal. Wouldn't we run into the same issue with a type-unstable output  if we used `max=true`?

jmert: I think so, yes.

musm: Yeah I don't think there's a good way to get rid of this by piggybacking on existing methods.  Unless we dispatch on `Val` types which is ugly and seems unnecessary. 


Why don't we rename it to `get_simple_extent_dims`  that way it adhere's  to the general api structure we are following with respect to naming.

musm: `get_dims` is also missing the `close` method on the dataspace.

jmert: > Why don't we rename it to `get_simple_extent_dims` that way it adhere's to the general api structure we are following with respect to naming.

I'm OK with that; just unfortunate we didn't get there before tagging 0.14, but :shrug: I guess we could add a silent deprecation for a while.

> `get_dims` is also missing the close method on the `dataspace`.

I can edit that, too; also should be extended to `Union{Dataset,Attribute}` since everything else is defined over that union already.

jmert:
Ah, in adding more tests, I found
```julia
julia> ds_null = HDF5.Dataspace(HDF5.h5s_create(HDF5.H5S_NULL))
HDF5.Dataspace: H5S_NULL

julia> length(ds_null)
1
```
but that looks wrong — I would expect length `== 0` since it can't contain anything.

jmert:
I realized I could get rid of one extra `checkvalid` from most methods by merging the implementations of `::Dataspace` and `::Union{Dataset,Attribute}` methods. I checked that the compiler does appear to specialize away the `isa` branches appropriately.

src/HDF5.jl
jmert: I realized, I was thinking of both `get_dims` methods — one on `Dataspace` and the other on `Dataset` — at the same time, but they don't have to be named the same. `get_dims(::Dataset)` is the method that's actually advertised in the docs. Were you thinking that only the `Dataspace` method should be renamed and we leave the `Dataset` method as-is?

musm: > I'm OK with that; just unfortunate we didn't get there before tagging 0.14, but shrug I guess we could add a silent deprecation for a while.


Yeah, I think it's fine though. I don't think we're 100% there yet with the external api yet, and I think it was a good idea to have an intermediate release(s) to get feedback and bug reports. Plus in the next minor release we can get rid of the depreciations added in this cycle.

musm: It's documented, but not exported and kind of 'internal exported api'. I think either (1) both should get renamed or (2) at least the method on  `Dataspace` and getting rid of the `Dataset` method `get_dims` method. I suspect most  users actually care about the `Base.size`  method, and determining `max_dims` is more for special use cases.

musm (approved):

jmert:
I think the Windows x86 test has failed every time I've pushed a new commit, so that needs to be checked out before merging...

jmert:
OK, I think that was part of it, but more problems have been revealed...

musm:
Strange so this is failing on x86 due to  calls like 

```
HDF5.h5s_create_simple(1, [0], [0])
```

Which error in a non-deterministic fashion.

```
HDF5-DIAG: Error detected in HDF5 (1.12.0) thread 0:
  #000: D:/mingwbuild/mingw-w64-hdf5/src/hdf5-1.12.0/src/H5S.c line 1459 in H5Screate_simple(): maxdims is smaller than dims
    major: Invalid arguments to routine
    minor: Bad value
```

musm:
I think we need to correctly truncate the  `Ints` to `h5size_t` for this to work.

musm:
I guess we need to force the following in the generate or another high-level wrapper method
```julia
    var"#status#" = ccall((:H5Screate_simple, libhdf5), hid_t, (Cint, Ptr{hsize_t}, Ptr{hsize_t}), rank, current_dims .% UInt64, maximum_dims .% UInt64)

```

jmert:
Ah, I misunderstood that `Base.cconvert` was going to do that for us:
```julia
julia> Base.cconvert(Ptr{UInt64}, Int32[1])
1-element Vector{Int32}:
 1
```
I'd expected to see a `Vector{Int64}` result. So yeah, that looks like an issue to be solved...

jmert:
> I guess we need to force the following (`current_dims .% UInt64`) in the binding generator...

I'm going to leave working on that to a future PR.

PR #759, opened 2020-12-06
Add binding for H5Tenum_insert

david-macmahon:
I needed this function to create an H5T_ENUM Datatype that would "conform to the `h5py` implementation of the numpy boolean type" as used in the UVH5 file format described [here](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/raw/main/docs/references/uvh5_memo.pdf).

Typical usage for this case would be:

```
# Create H5T_ENUM Datatype
boolenumtype = create_datatype(HDF5.H5T_ENUM, sizeof(Bool)); # Semicolon to avoid showing in REPL

# Insert enum names/values
HDF5.h5t_enum_insert(boolenumtype.id, "FALSE", false)
HDF5.h5t_enum_insert(boolenumtype.id, "TRUE", true)

# Create dataset using enum Datatype
flags = create_dataset(uvh5, "Data/flags", boolenumtype, flagsdataspace)
```

### DIFF
--- a/gen/api_defs.jl
+++ b/gen/api_defs.jl
@@ -225,6 +225,7 @@
 @bind h5t_commit(loc_id::hid_t, name::Ptr{UInt8}, dtype_id::hid_t, lcpl_id::hid_t, tcpl_id::hid_t, tapl_id::hid_t)::herr_t "Error committing type"
 @bind h5t_copy(dtype_id::hid_t)::hid_t "Error copying datatype"
 @bind h5t_create(class_id::Cint, sz::Csize_t)::hid_t error("Error creating datatype of id ", class_id)
+@bind h5t_enum_insert(dtype_id::hid_t, name::Cstring, value::Ref{Bool})::herr_t error("Error adding ", name, " to enum datatype")
 @bind h5t_equal(dtype_id1::hid_t, dtype_id2::hid_t)::hid_t "Error checking datatype equality"
 @bind h5t_get_array_dims(dtype_id::hid_t, dims::Ptr{hsize_t})::Cint "Error getting dimensions of array"
 @bind h5t_get_array_ndims(dtype_id::hid_t)::Cint "Error getting ndims of array"
--- a/gen/api_defs.jl
+++ b/gen/api_defs.jl
@@ -225,7 +225,7 @@
 @bind h5t_commit(loc_id::hid_t, name::Ptr{UInt8}, dtype_id::hid_t, lcpl_id::hid_t, tcpl_id::hid_t, tapl_id::hid_t)::herr_t "Error committing type"
 @bind h5t_copy(dtype_id::hid_t)::hid_t "Error copying datatype"
 @bind h5t_create(class_id::Cint, sz::Csize_t)::hid_t error("Error creating datatype of id ", class_id)
-@bind h5t_enum_insert(dtype_id::hid_t, name::Cstring, value::Ref{Bool})::herr_t error("Error adding ", name, " to enum datatype")
+@bind h5t_enum_insert(dtype_id::hid_t, name::Cstring, value::Ptr{Cvoid})::herr_t error("Error adding ", name, " to enum datatype")
 @bind h5t_equal(dtype_id1::hid_t, dtype_id2::hid_t)::hid_t "Error checking datatype equality"
 @bind h5t_get_array_dims(dtype_id::hid_t, dims::Ptr{hsize_t})::Cint "Error getting dimensions of array"
 @bind h5t_get_array_ndims(dtype_id::hid_t)::Cint "Error getting ndims of array"
--- a/docs/src/api_bindings.md
+++ b/docs/src/api_bindings.md
@@ -214,7 +214,7 @@ h5t_commit(loc_id::hid_t, name::Ptr{UInt8}, dtype_id::hid_t, lcpl_id::hid_t, tcp
 h5t_committed(dtype_id::hid_t)
 h5t_copy(dtype_id::hid_t)
 h5t_create(class_id::Cint, sz::Csize_t)
-h5t_enum_insert(dtype_id::hid_t, name::Cstring, value::Ref{Bool})
+h5t_enum_insert(dtype_id::hid_t, name::Cstring, value::Ptr{Cvoid})
 h5t_equal(dtype_id1::hid_t, dtype_id2::hid_t)
 h5t_get_array_dims(dtype_id::hid_t, dims::Ptr{hsize_t})
 h5t_get_array_ndims(dtype_id::hid_t)
--- a/docs/src/api_bindings.md
+++ b/docs/src/api_bindings.md
@@ -214,6 +214,7 @@ h5t_commit(loc_id::hid_t, name::Ptr{UInt8}, dtype_id::hid_t, lcpl_id::hid_t, tcp
 h5t_committed(dtype_id::hid_t)
 h5t_copy(dtype_id::hid_t)
 h5t_create(class_id::Cint, sz::Csize_t)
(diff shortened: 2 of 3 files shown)


gen/api_defs.jl
```diff
@@ -225,6 +225,7 @@
 @bind h5t_commit(loc_id::hid_t, name::Ptr{UInt8}, dtype_id::hid_t, lcpl_id::hid_t, tcpl_id::hid_t, tapl_id::hid_t)::herr_t "Error committing type"
 @bind h5t_copy(dtype_id::hid_t)::hid_t "Error copying datatype"
 @bind h5t_create(class_id::Cint, sz::Csize_t)::hid_t error("Error creating datatype of id ", class_id)
+@bind h5t_enum_insert(dtype_id::hid_t, name::Cstring, value::Ref{Bool})::herr_t error("Error adding ", name, " to enum datatype")
```
jmert, line 228:
```suggestion
@bind h5t_enum_insert(dtype_id::hid_t, name::Cstring, value::Ptr{Cvoid})::herr_t error("Error adding ", name, " to enum datatype")
```

jmert (changes requested):
The low-level API types need to match the C API's.

gen/api_defs.jl
jmert: With this change, your setup then becomes:
```julia
boolenumtype = create_datatype(HDF5.H5T_ENUM, sizeof(Bool))
HDF5.h5t_enum_insert(boolenumtype, "FALSE", Ref(false))
HDF5.h5t_enum_insert(boolenumtype, "TRUE", Ref(true))
```

musm: Ideally, we could call this using something like the following (but not the following).

That way we wrap the input in a `Ref` automatically. Unfortunately, due to the untyped method  in the generator, I had to add a prefix underscore to avoid the recursive call. 

Not sure if there's a way to do something like this but without messing with underscore with the generator. 


```diff

--- a/gen/api_defs.jl
+++ b/gen/api_defs.jl
@@ -225,7 +225,7 @@
 @bind h5t_commit(loc_id::hid_t, name::Ptr{UInt8}, dtype_id::hid_t, lcpl_id::hid_t, tcpl_id::hid_t, tapl_id::hid_t)::herr_t "Error committing type"
 @bind h5t_copy(dtype_id::hid_t)::hid_t "Error copying datatype"
 @bind h5t_create(class_id::Cint, sz::Csize_t)::hid_t error("Error creating datatype of id ", class_id)
-@bind h5t_enum_insert(dtype_id::hid_t, name::Cstring, value::Ref{Bool})::herr_t error("Error adding ", name, " to enum datatype")
+@bind _h5t_enum_insert(dtype_id::hid_t, name::Cstring, value::Ref{Any})::herr_t error("Error adding ", name, " to enum datatype")
 @bind h5t_equal(dtype_id1::hid_t, dtype_id2::hid_t)::hid_t "Error checking datatype equality"
 @bind h5t_get_array_dims(dtype_id::hid_t, dims::Ptr{hsize_t})::Cint "Error getting dimensions of array"
 @bind h5t_get_array_ndims(dtype_id::hid_t)::Cint "Error getting ndims of array"
diff --git a/src/api.jl b/src/api.jl
index bba6d9e..1622c4b 100644
--- a/src/api.jl
+++ b/src/api.jl
@@ -884,8 +884,8 @@ function h5t_create(class_id, sz)
     return var"#status#"
 end
 
-function h5t_enum_insert(dtype_id, name, value)
-    var"#status#" = ccall((:H5Tenum_insert, libhdf5), herr_t, (hid_t, Cstring, Ref{Bool}), dtype_id, name, value)
+function _h5t_enum_insert(dtype_id, name, value)
+    var"#status#" = ccall((:H5Tenum_insert, libhdf5), herr_t, (hid_t, Cstring, Ref{Any}), dtype_id, name, value)
     var"#status#" < 0 && error("Error adding ", name, " to enum datatype")
     return nothing
 end
@@ -1140,4 +1140,3 @@ function h5z_register(filter_class)
     var"#status#" < 0 && error("Unable to register new filter")
     return nothing
 end
-
diff --git a/src/api_helpers.jl b/src/api_helpers.jl
index 0bb5b35..e78e980 100644
--- a/src/api_helpers.jl
+++ b/src/api_helpers.jl
@@ -266,6 +266,9 @@ end
 ### Datatype Interface
 ###
 
+h5t_enum_insert(dtype_id, name, value) =
+    _h5t_enum_insert(dtype_id, name, Ref(value))
+
 function h5t_get_array_dims(type_id)
     nd = h5t_get_array_ndims(type_id)
     dims = Vector{hsize_t}(undef, nd)

```

jmert: > Not sure if there's a way to do something like this but without messing with underscore with the generator.

The generator could probably be taught that `Ref{Any}` is a special case and could add an expression like
```julia
carg = arg isa Ref{<:Any} ? arg : Base.cconvert(Ref{eltype(arg)}, arg)
...
ccall((cfunc, :libhdf5), hid_t, (..., carg), ..., Ptr{Cvoid})
```
which I think should allow arguments of any `Ref` type to pass through unchanged, pass through arrays (which are already convertible to `Ptr{Cvoid}` without extra wrapping), and capture other scalar types and wrap them.

Not sure if there are other corner cases to worry about, though, so that would probably require some careful testing.

musm: Why do we need `Base.cconvert(Ref{eltype(arg)}, arg)` in the case `arg` is not a `Ref`?

musm: oh nevermind, I see what that's doing. Yes that would work I think.

jmert: It's the more general than `Ref(arg)` in that it deals with arrays directly:
```julia
julia> arg = [1, 2]
2-element Vector{Int64}:
 1
 2

julia> Ref(arg)
Base.RefValue{Vector{Int64}}([1, 2])

julia> Base.cconvert(Ref{eltype(arg)}, arg)
Base.RefArray{Int64, Vector{Int64}, Nothing}([1, 2], 1, nothing)
```

musm: I see, I'm unfamiliar with `RefArray`'s, so this looks like it will also be able to handle scalar's in addition to arrays and correctly wrap them, and also   <:Nothing? 

Looks like `cconvert` directly calls `convert` when `T` isn't a `Ptr` so might be better to just use `convert `here. 

I'm on board with this.

jmert: > Looks like cconvert directly calls convert when T isn't a Ptr so might be better to just use convert here.

I think we have to keep `cconvert` because that's the overload point for converting for later passing to `ccall` — consider #750 where we just changed everything to have `cconvert` methods but no `convert` methods.

jmert: > I'm on board with this.

I'll work on the generator update that makes this possible.

musm: > I think we have to keep cconvert because that's the overload point for converting for later passing to ccall — consider #750 where we just changed everything to have cconvert methods but no convert methods.

Oh right, good point.

david-macmahon: Glad my "type sloppiness" started a useful discussion.  Wish I could say that was my plan from the start, but I'm not that clever 😄

david-macmahon:
Good point about matching the C API types.  I like the "purity" of that, but I also like the idea of a slightly higher level more convenient interface as well.  I've matched the types for now, but look forward to seeing what happens with the updated generator idea.  Thanks for all the quick responses!

jmert:
@musm I'd vote for merging this as-is right now and then adding the special case behavior in a follow-up. The explicit form of wrapping in `Ref`s should be backwards compatible with a smarter generator-made wrapper.

jmert (approved):

david-macmahon:
Thanks!

PR #762, opened 2020-12-07
RFC: Use fewer objects in h5open and read

jmert:
This PR closes most of the performance gap identified in reads from #752. The changes are:

1. Reuse the `Datatype` identified in each of the untyped `read()` functions when handing off to the generic typed-read. This is accomplished by renaming the generic `read()` to `_read()` and giving it both the HDF5 `Datatype` and the Julia `Type` as arguments.
2. Only create a new in-memory `Dataspace` if a hyperslab is being selected; otherwise, reuse the file `Dataspace`.
3. Only create the file access property list once in `h5open`, and only create a create-file property list if non-default options are being set (and close it if one was created).

image

Part of the performance gap appears to be due to libhdf5 v1.12 being a bit slower than v1.10, but if the library version is forced back to v1.10, each of the operations are on-par or faster than with HDF5.jl v0.13.7.

The RFC comes from the renaming of `read()` — I'm not sure just hiding the generic read as `_read()` is really the best option since it does make specializing on particular types require overloading the internal `_read` instead of public `read` if special data handling is required (as demonstrated by the `_read(..., ::Type{Opaque})` case). Ultimately, though, it does seems some sort of similar structuring will be desired to retain the performance; might just need to iterate on the exact restructuring a bit.

(Note that the first commit is from #747 since both that PR and the changes here modify the opaque read method.)

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1292,7 +1292,7 @@ function generic_read(obj::DatasetOrAttribute, filetype::Datatype, ::Type{T}, I.
     else
         buf = Array{T}(undef, sz...)
     end
-    memspace = dataspace(sz)
+    memspace = isempty(I) ? dspace : dataspace(sz)
 
     if obj isa Dataset
         h5d_read(obj, memtype, memspace, dspace, obj.xfer, buf)
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -408,11 +408,9 @@ Pass `swmr=true` to enable (Single Writer Multiple Reader) SWMR write access for
 "r+", or SWMR read access for "r".
 """
 function h5open(filename::AbstractString, mode::AbstractString = "r"; swmr::Bool = false, pv...)
-    apl = create_property(H5P_FILE_ACCESS; pv...) # file access property list
     # With garbage collection, the other modes don't make sense
-    # (Set this first, so that the user-passed properties can overwrite this.)
-    apl[:fclose_degree] = H5F_CLOSE_STRONG
-    cpl = create_property(H5P_FILE_CREATE; pv...) # file create property list
+    apl = create_property(H5P_FILE_ACCESS; pv..., fclose_degree = H5F_CLOSE_STRONG) # file access property list
+    cpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(H5P_FILE_CREATE; pv...) # file create property list
     rd, wr, cr, tr, ff =
         mode == "r"  ? (true,  false, false, false, false) :
         mode == "r+" ? (true,  true,  false, false, true ) :
@@ -425,13 +423,6 @@ function h5open(filename::AbstractString, mode::AbstractString = "r"; swmr::Bool
         error("HDF5 does not support appending without writing")
     end
 
-    close_apl = false
-    if apl == DEFAULT_PROPERTIES
-        apl = create_property(H5P_FILE_ACCESS)
-        close_apl = true
-        # With garbage collection, the other modes don't make sense
-        apl[:fclose_degree] = H5F_CLOSE_STRONG
-    end
     if cr && (tr || !isfile(filename))
         flag = swmr ? H5F_ACC_TRUNC|H5F_ACC_SWMR_WRITE : H5F_ACC_TRUNC
         fid = h5f_create(filename, flag, cpl, apl)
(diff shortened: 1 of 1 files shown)


src/HDF5.jl
```diff
@@ -443,11 +443,10 @@ Pass `swmr=true` to enable (Single Writer Multiple Reader) SWMR write access for
 "r+", or SWMR read access for "r".
 """
 function h5open(filename::AbstractString, mode::AbstractString = "r"; swmr::Bool = false, pv...)
-    apl = create_property(H5P_FILE_ACCESS; pv...) # file access property list
     # With garbage collection, the other modes don't make sense
     # (Set this first, so that the user-passed properties can overwrite this.)
-    apl[:fclose_degree] = H5F_CLOSE_STRONG
-    cpl = create_property(H5P_FILE_CREATE; pv...) # file create property list
+    apl = create_property(H5P_FILE_ACCESS; fclose_degree = H5F_CLOSE_STRONG, pv...) # file access property list
+    cpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(H5P_FILE_CREATE; pv...) # file create property list
```
jmert, line 449: There's also the question of setting `apl` properties versus what the comment says. The previous code sets `:fclose_degree` after constructing the property, so it's _not_ doing what the comment says, while here I switched it to match the comment. For comparison, the `_h5open` function above will always overwrite the `:fclose_degree` property.

I assume I should actually remove the comment and change back to always overriding `:fclose_degree`?

musm: Assuming the comment on GC is true, it sounds like we should always overwrite the  user-passed property for `fclose_degree`.

musm: Looks like I forgot to delete `_h5open` at some point.

src/HDF5.jl
```diff
@@ -460,13 +459,6 @@ function h5open(filename::AbstractString, mode::AbstractString = "r"; swmr::Bool
         error("HDF5 does not support appending without writing")
     end
 
-    close_apl = false
-    if apl == DEFAULT_PROPERTIES
```
musm, line 464: Yikes, by construction `apl !=  DEFAULT_PROPERTIES`, since `    apl[:fclose_degree] = H5F_CLOSE_STRONG	` on the line above.

src/HDF5.jl
```diff
@@ -481,12 +473,9 @@ function h5open(filename::AbstractString, mode::AbstractString = "r"; swmr::Bool
...
         fid = h5f_open(filename, flag, apl)
     end
-    if close_apl
-        # Close properties manually to avoid errors when the file is
-        # closed before the properties are gc'ed
-        close(apl)
-    end
-    File(fid, filename)
+    close(apl)
+    cpl != DEFAULT_PROPERTIES && close(cpl)
```
musm, line 477: We were missing this before? I'm guessing this is defensive and the hdf5 library was handling closing this when the file handle was closed previously?

musm:
LGTM

src/HDF5.jl
jmert: Yeah, presumably GC and/or libhdf5 was handling it, but as #754 showed, it's probably advantageous to do this deterministically and early.

musm:
> The RFC comes from the renaming of read() — I'm not sure just hiding the generic read as _read() is really the best option since it does make specializing on particular types require overloading the internal _read instead of public read if special data handling is required (as demonstrated by the _read(..., ::Type{Opaque}) case). Ultimately, though, it does seems some sort of similar structuring will be desired to retain the performance; might just need to iterate on the exact restructuring a bit.

Yeah I don't see a way to avoid that, but I also don't see this as a huge issue. How else would it be possible to allow such re-structing without the current signature were the Datatype is passed in as the second argument?

jmert:
I guess the question if we're OK with the idea (and sitting on it for another day, no other alternative has come to me) would be should we allocate a better, more descriptive name than `_read`? `generic_read` might be better. It'd make describing in a future "dev docs" how to do an Opaque-like overload look more supported.

musm:
> generic_read might be better. It'd make describing in a future "dev docs" how to do an Opaque-like overload look more supported. 

That sounds good to me `generic_read` or `h5_read` (I kind of don't like this as much)

I think as long as it's documented, having the second argument require the `datatype` pass though, should be fine IMO.

jmert:
CI all green!

musm:
I don't know why, but seeing all green checkmarks  gives me  a dopamine hit 🎉

src/HDF5.jl
musm: On second thought looking at https://github.com/JuliaIO/HDF5.jl/pull/617/files it looks like for virtual datasets we need a way to override  fclose_degree

PR #767, opened 2020-12-09
Use sizeof for Datatype

musm:
Not sure what to make of the `Csize_t`. Ideally we could truncate all of the return types to `Csize_t % Int` but I wonder if there are cases where this would fail i.e. when the unsigned size is larger than Int.


jmert (approved):
I think converting to `Int` rather than truncating is fine for now. It's the slightly more conservative option that will throw clear errors if the size overflows, and it can always be changed to truncation in the future if through experience there'd be no impact anyway (for any practically usable situation).

PR #768, opened 2020-12-09
Automatically convert C Int types to Julia's Int type for convenience

musm:
Admittedly this had less of an impact on simplifying the code base as I originally hoped, but it does overall enhance usability with Julia since `Int` is the canonical integer type, so I think it's a good change.

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -372,7 +372,7 @@ end
 function h5i_dec_ref(obj_id)
     var"#status#" = ccall((:H5Idec_ref, libhdf5), Cint, (hid_t,), obj_id)
     var"#status#" < 0 && error("Error decementing reference")
-    return var"#status#"
+    return Int(var"#status#")
 end
 
 function h5i_get_file_id(obj_id)
@@ -390,19 +390,19 @@ end
 function h5i_get_ref(obj_id)
     var"#status#" = ccall((:H5Iget_ref, libhdf5), Cint, (hid_t,), obj_id)
     var"#status#" < 0 && error("Error getting reference count")
-    return var"#status#"
+    return Int(var"#status#")
 end
 
 function h5i_get_type(obj_id)
     var"#status#" = ccall((:H5Iget_type, libhdf5), Cint, (hid_t,), obj_id)
     var"#status#" < 0 && error("Error getting type")
-    return var"#status#"
+    return Int(var"#status#")
 end
 
 function h5i_inc_ref(obj_id)
     var"#status#" = ccall((:H5Iinc_ref, libhdf5), Cint, (hid_t,), obj_id)
     var"#status#" < 0 && error("Error incrementing identifier refcount")
-    return var"#status#"
+    return Int(var"#status#")
 end
 
 function h5i_is_valid(obj_id)
@@ -522,7 +522,7 @@ end
 function h5p_get_chunk(plist_id, n_dims, dims)
     var"#status#" = ccall((:H5Pget_chunk, libhdf5), Cint, (hid_t, Cint, Ptr{hsize_t}), plist_id, n_dims, dims)
     var"#status#" < 0 && error("Error getting chunk size")
-    return var"#status#"
+    return Int(var"#status#")
 end
 
 function h5p_get_create_intermediate_group(lcpl_id, crt_intermed_group)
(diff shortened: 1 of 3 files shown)


src/api.jl
```diff
@@ -964,25 +964,25 @@ end
...
 function h5t_get_sign(dtype_id)
     var"#status#" = ccall((:H5Tget_sign, libhdf5), Cint, (hid_t,), dtype_id)
     var"#status#" < 0 && error("Error getting sign")
-    return var"#status#"
+    return Int(var"#status#")
 end
 
 function h5t_get_size(dtype_id)
     var"#status#" = ccall((:H5Tget_size, libhdf5), Csize_t, (hid_t,), dtype_id)
     var"#status#" < 0 && error("Error getting size")
```
musm, line 978: THe error check here should be <= 0

src/api.jl
```diff
@@ -935,18 +935,18 @@ end
...
     var"#status#" = ccall((:H5Tget_member_index, libhdf5), Cint, (hid_t, Ptr{UInt8}), dtype_id, membername)
     var"#status#" < 0 && error("Error getting index of compound datatype member \"", membername, "\"")
-    return var"#status#"
+    return Int(var"#status#")
 end
 
 function h5t_get_member_offset(dtype_id, index)
     var"#status#" = ccall((:H5Tget_member_offset, libhdf5), Csize_t, (hid_t, Cuint), dtype_id, index)
-    return var"#status#"
+    return Int(var"#status#")
```
musm, line 949: This looks like it's missing it's error check

src/api.jl
```diff
@@ -899,31 +899,31 @@ end
...
 function h5t_get_cset(dtype_id)
     var"#status#" = ccall((:H5Tget_cset, libhdf5), Cint, (hid_t,), dtype_id)
     var"#status#" < 0 && error("Error getting character set encoding")
-    return var"#status#"
+    return Int(var"#status#")
 end
 
 function h5t_get_ebias(dtype_id)
     var"#status#" = ccall((:H5Tget_ebias, libhdf5), Csize_t, (hid_t,), dtype_id)
     var"#status#" < 0 && error("Error getting datatype floating point exponent bias")
```
musm, line 925: Same here
check should be <= 0 
https://portal.hdfgroup.org/display/HDF5/H5T_GET_EBIAS

src/api.jl
musm: Nvm, this one looks like it does not error

src/api.jl
musm: Sorry neither of these should have errors.

jmert:
I'm definitely on board with automatically converting `Cint` since C's `int` and Julia's `Int` are the default integer types in each language.

I'm less certain about automatically converting `Csize_t` within the low-level API even though nearly all of the mid- and high-level API will want to convert. (With reservations at the "uncertain"/"questioning" level, far from an actual "objection".)

musm:
> I'm less certain about automatically converting Csize_t within the low-level API even though nearly all of the mid- and high-level API will want to convert. (With reservations at the "uncertain"/"questioning" level, far from an actual "objection".)

To me this conversion is even more safe since it's going from UInt32 -> Int. Looking at the only  three methods that return a Csize_t it seems  harmless to do the conversion.

jmert:
Is `Csize_t` == `UInt32` on Windows? On Linux, it's `UInt64`.

musm:
Whops, my bad, it's not (I was running x86, probably from testing JLD)

musm:
This boils down to the same concerns we had with https://github.com/JuliaIO/HDF5.jl/blob/master/src/api_helpers.jl#L257  where we apply the `Int` conversion https://github.com/JuliaIO/HDF5.jl/blob/master/src/HDF5.jl#L1044. For that specific case, it is fairly a 'safe' conversion, since  Julia's array constructors cast all sizes to Int.

So I share your concern's with the autoconversion in the Csize_t case as well. 

We are going to have to make a similar decision wrt  vector **inputs* that need conversion to `hsize_t`.

musm:
For now I've limited the conversion to Cint -> Int. We can explore the consequences of Csize_t conversions (I overall think that's safe, but we can let that simmer for a bit longer)

jmert (approved):

PR #775, opened 2020-12-10
Check if object exists before create_(dataset|group)

musm:
This gets rid of the nasty HDF5 library error print out.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -704,7 +704,9 @@ end
 function create_group(parent::Union{File,Group}, path::AbstractString,
                   lcpl::Properties=_link_properties(path),
                   gcpl::Properties=DEFAULT_PROPERTIES)
-    Group(h5g_create(checkvalid(parent), path, lcpl, gcpl, H5P_DEFAULT), file(parent))
+    checkvalid(parent)
+    haskey(parent, path) && error("cannot create group: object \"", path, "\" already exists at ", name(parent))
+    Group(h5g_create(parent, path, lcpl, gcpl, H5P_DEFAULT), file(parent))
 end
 
 function create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::Datatype, dspace::Dataspace,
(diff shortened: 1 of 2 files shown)


musm:
Apparently `haskey` can't handle attribute types though `h5l_exists`

src/HDF5.jl
```diff
@@ -718,6 +720,8 @@ function create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::
     dcpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(H5P_DATASET_CREATE; pv...)
     dxpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(H5P_DATASET_XFER; pv...)
     dapl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(H5P_DATASET_ACCESS; pv...)
+    checkvalid(parent)
+    haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
     Dataset(h5d_create(checkvalid(parent), path, dtype, dspace, _link_properties(path), dcpl, dapl), file(parent), dxpl)
```
jmert, line 725:
```suggestion
    checkvalid(parent)
    haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
    Dataset(h5d_create(parent, path, dtype, dspace, _link_properties(path), dcpl, dapl), file(parent), dxpl)
```

Though I would also suggest getting rid of the explicit `checkvalid(parent)` in both functions and to edit `haskey`'s implementation to do the `checkvalid` (which is right now inconsistently applied — checked for attributes, datasets, and datatypes but not for files and groups).

That could be left to a followup  PR if you prefer, though.

jmert (approved):

musm:
I'll save that to another PR. Something looks a bit fishy with our handling of attributes for `HDF5.Fil`. For example, I'm not sure it makes sense to allow writing attributes to `File` yet we permit it   (the docs suggest it's possible, but I don't see why). This threw a small wrench in the works when testing using `haskey`.

jmert:
> For example, I'm not sure it makes sense to allow writing attributes to `File` yet we permit it (the docs suggest it's possible, but I don't see why).

I think throughout much of libhdf5, the file handle is treated almost equivalently as the handle to the "/" group. So in that sense, writing attributes is fine since they're just attached to the root group.

PR #776, opened 2020-12-10
Delete straggling create_dataset method

musm:
Not sure if we really need a deprecation for this, but looks like it accidentally snuck in v0.14.

EDIT: yes we do, looks like JLD is using this. (Finally good use of the integration testing)

Altough the failure there is mem map failure not anything related to the deprecation AFAIK.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -447,4 +447,5 @@ end
 ### Changed in PR#776
 @deprecate create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::Datatype, dspace::Dataspace,
 lcpl::Properties, dcpl::Properties,
-dapl::Properties, dxpl::Properties) create_dataset(parent, path, dtype, dspace; lcpl=lcpl, dcpl=dcpl, dapl=dapl, dxpl=dxpl) false
+dapl::Properties, dxpl::Properties) HDF5.Dataset(
+        HDF5.h5d_create(parent, path, dtype, dspace, lcpl, dcpl, dapl), HDF5.file(parent), dxpl)) false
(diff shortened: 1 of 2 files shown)


src/deprecated.jl
```diff
@@ -438,3 +438,13 @@ end
...
+
+
+###
+### v0.15 deprecations
+###
+
+### Changed in PR#776
+@deprecate create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::Datatype, dspace::Dataspace,
+lcpl::Properties, dcpl::Properties,
+dapl::Properties, dxpl::Properties) create_dataset(parent, path, dtype, dspace; lcpl=lcpl, dcpl=dcpl, dapl=dapl, dxpl=dxpl) false
```
musm, line 450: this is wrong..

jmert, line 450: I think the literal translation that can occur automatically is
```suggestion
@deprecate create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::Datatype, dspace::Dataspace,
lcpl::Properties, dcpl::Properties,
dapl::Properties, dxpl::Properties) HDF5.Dataset(
        HDF5.h5d_create(parent, path, dtype, dspace, lcpl, dcpl, dapl), file(parent), dxpl)) false
```
though the deprecation warning would probably be more useful if it tried to steer users to using keyword properties on a `create_dataset` call instead.

calebwin:
@musm @jmert Wait, this is deprecated? what should I use instead? just `file[dataset_name] = some_array`?

musm:
Yep. Also if you run julia with depwarn set to true you will get the depreciation warnings printed, which will making upgrading trivial.

PR #777, opened 2020-12-10
Wrap a subset of the libhdf5 table interface methods

musm:
Methods from https://github.com/JuliaIO/HDF5.jl/pull/511


src/api_helpers.jl
```diff
@@ -325,6 +325,23 @@ end
...
+function h5tb_get_table_info(loc_id, table_name)
+    nfields = Ref{hsize_t}()
+    nrecords = Ref{hsize_t}()
+    h5tb_get_table_info(loc_id, table_name, nfields, nrecords)
+    return nfields[], nrecords[]
+end
+
+function h5tb_get_field_info(loc_id, table_name)
+    nfields, = h5tb_get_table_info(loc_id, table_name)
+    field_names = Vector{UInt8}[fill(0x00,255) for i in 1:2]
```
musm, line 337: I don't know if there is any way to determine the length of these. 
In https://confluence.hdfgroup.org/display/HDF5/H5TB_GET_FIELD_INFO they just allocate 255 sized buffer.

src/api_helpers.jl
```diff
@@ -325,6 +325,30 @@ end
...
+function h5tb_get_field_info(loc_id, table_name)
+    nfields, = h5tb_get_table_info(loc_id, table_name)
+    strlen = 64 # string buffer size
+    field_sizes = Vector{Csize_t}(undef,nfields)
+    field_offsets = Vector{Csize_t}(undef,nfields)
+    type_size = Ref{Csize_t}()
+    while true
+        field_names = [Vector{UInt8}(undef,strlen) for _ in 1:nfields]
+        h5tb_get_field_info(loc_id, table_name, field_names, field_sizes, field_offsets, type_size)
+        if any(!in(0x00, buf) for buf in field_names)
```
musm, line 344: I'm hoping it's safe to assume HDF5 is nullterminating the string. since that's my assumption.

src/api_helpers.jl
```diff
@@ -325,6 +325,30 @@ end
...
+    return nfields[], nrecords[]
+end
+
+function h5tb_get_field_info(loc_id, table_name)
+    nfields, = h5tb_get_table_info(loc_id, table_name)
+    strlen = 64 # string buffer size
+    field_sizes = Vector{Csize_t}(undef,nfields)
+    field_offsets = Vector{Csize_t}(undef,nfields)
+    type_size = Ref{Csize_t}()
+    while true
```
musm, line 341: To work around the issue of determing the string buffer size, here I just initially arbitrary set it to 64 then we keep doubling until it until the string fits (detected through presence of a null byte)

musm:
This is now ready for review.

src/api_helpers.jl
jmert: Unfortunately, I don't think this is valid/safe: https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/hl/src/H5TB.c?at=refs%2Fheads%2Fhdf5_1_10#3092 libhdf5 just does a `strcpy`, so for a long name, it is overwriting the buffer.

src/api_helpers.jl
jmert: I think that is correct.

src/api_helpers.jl
musm: Right, which is why if it doesn't fit, we resize the buffer and then check if there's a `0x00` in the buffer which would indicate it was copied correctly.  If it didin't fit it would overflow the buffer and no null byte would be detected, in which case we increase the buffer size.

src/api_helpers.jl
musm: Yea I know think so to, thanks to your reference below  (`strcpy`). Thanks for checking.

src/api_helpers.jl
jmert: But overflowing means you're overwriting other bytes in memory — this is a classic buffer overflow attack vector.

jmert: It appears this is just a wrapper over datatype querying functions, so unfortunately the safe thing to do might be to just pass `C_NULL` pointer for the names (which the function implementation shows should just skip over filling), and then to do the loop of `h5t_get_member_name` over the table dataset's datatype ourselves.

musm: Hmm, good point. I don't know what we can do other than perhaps loop through and call `H5Tget_member_name` to manually determine the required buffer size? At which point we're just reimplementing this function?

Heck even their example is also susceptible  to overflow:
https://confluence.hdfgroup.org/display/HDF5/H5TB_GET_FIELD_INFO

jmert: Yeah, I'm kind of disappointed with both the documentation and implementation of this function :-\

musm: Ok I updated this PR to avoid the buffer attack, by computing the column names via `h5t_get_member_name `

musm:
Sam's objections I plan to merge soon

src/api_helpers.jl
```diff
@@ -325,6 +325,29 @@ end
...
+    field_offsets = Vector{Csize_t}(undef, nfields)
+    type_size = Ref{Csize_t}()
+    # pass C_NULL to field_names argument since libhdf5 does not provide a way to determine if the
+    # allocated buffer is the correct length, which is thus susceptible to a buffer overflow if
+    # an incorrect buffer length is passed. Instead, we manually compute the column names using the
+    # same calls that h5tb_get_field_info internally uses.
+    h5tb_get_field_info(loc_id, table_name, C_NULL, field_sizes, field_offsets, type_size)
+    did = h5d_open(loc_id, table_name, H5P_DEFAULT)
+    tid = h5d_get_type(did)
+    field_names = [h5t_get_member_name(tid, i-1) for i in 1:nfields]
```
jmert, line 347:
```suggestion
    did = h5d_open(loc_id, table_name, H5P_DEFAULT)
    tid = h5d_get_type(did)
    h5d_close(did)
    field_names = [h5t_get_member_name(tid, i-1) for i in 1:nfields]
    h5t_close(tid)
```

PR #778, opened 2020-12-11
Fix error message for KeyError

musm:


jmert:
I think my intention was throwing the tuple of values; apparently I missed a set of parentheses.
``julia
julia> KeyError((1, 2, 3))
KeyError((1, 2, 3))
```
That'd avoid the string interpolation machinery that otherwise has to get inlined into the function as well.

jmert (approved):

PR #780, opened 2020-12-14
Fix ishdf5 error handling and improve general error silencing

musm:
The docs state that h5f_is_hdf5 does not error throw, but that is not true.
We first suppress the hdf5 libraries stack trace printing, and then fix
the return to ensure false condition is thrown when the file is not
accessible in the hdf5 format.

We have also renamed the hiding_errors method and ensured that the final
stacktrace reset is guaranteed in case the underlying function call
throws an error.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -409,8 +409,8 @@ Pass `swmr=true` to enable (Single Writer Multiple Reader) SWMR write access for
 """
 function h5open(filename::AbstractString, mode::AbstractString = "r"; swmr::Bool = false, pv...)
     # With garbage collection, the other modes don't make sense
-    apl = create_property(H5P_FILE_ACCESS; pv..., fclose_degree = H5F_CLOSE_STRONG) # file access property list
-    cpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(H5P_FILE_CREATE; pv...) # file create property list
+    fapl = create_property(H5P_FILE_ACCESS; pv..., fclose_degree = H5F_CLOSE_STRONG) # file access property list
+    fcpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(H5P_FILE_CREATE; pv...) # file create property list
     rd, wr, cr, tr, ff =
         mode == "r"  ? (true,  false, false, false, false) :
         mode == "r+" ? (true,  true,  false, false, true ) :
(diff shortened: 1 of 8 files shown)


src/HDF5.jl
```diff
@@ -1968,14 +1975,15 @@ function create_external(source::Union{File,Group}, source_relpath, target_filen
...
-    old_func, old_client_data = h5e_get_auto(error_stack)
-    h5e_set_auto(error_stack, C_NULL, C_NULL)
-    res = f()
-    h5e_set_auto(error_stack, old_func, old_client_data)
-    return res
+function silence_errors(f::Function)
+    estack = H5E_DEFAULT
+    func, client_data = h5e_get_auto(estack)
+    h5e_set_auto(estack, C_NULL, C_NULL)
+    try
```
musm, line 1982: the `try`...`finally` statement is required in the case  where `f()` throws an `error` message, otherwise the error stack wouldn't get reset back.

src/HDF5.jl
```diff
@@ -635,9 +633,18 @@ end
 """
     ishdf5(name::AbstractString)
 
-Returns `true` if `name` is a path to a valid hdf5 file, `false` otherwise.
+Returns `true` if the file specified by `name` is in the HDF5 format, and `false` otherwise.
 """
-ishdf5(name::AbstractString) = h5f_is_hdf5(name)
+function ishdf5(name::AbstractString)
+    # v1.12 use more robust h5f_isaccesible
```
jmert, line 639:
```suggestion
function ishdf5(name::AbstractString)
    isfile(name) || return false
    # v1.12 use more robust h5f_isaccesible
```
I think the main case where `h5f_is_hdf5` prints errors rather than just returning false is when the file isn't accessible to do the check (such as it not existing). We could do any number of fancier checks on permissions to try to avoid having to do the error handler swapping, but I expect those eventually become cheap compared to a large number of filesystem accesses. Just checking if the file exists should be cheap though because (a) if it doesn't, we avoid the error handler swap, and (b) if it does, libhdf5 is just going to do the same lookup, and that should be cached by the OS so should be fast.

jmert (approved):

src/HDF5.jl
musm: > I think the main case where h5f_is_hdf5 prints errors rather than just returning false is when the file isn't accessible to do the check (such as it not existing). 

yes, I think that's right, that was the main case I was working around. I think its reasonable to assume that that's the main case where an `error` is mistakenly thrown by libhdf5.


To be clear, are you proposing that I change this function to only do the  `isfile` check ? 
```
function ishdf5(name::AbstractString)
    isfile(name) || return false
    # v1.12 use more robust h5f_is_accesible
    h5f_is_hdf5(name)
end

```

jmert: No, I think the rest probably is necessary and unavoidable since there are a myriad number of other ways the file could exist but not be validly readable (`touch unreadable && chmod 000 unreadable` seems to do it...), but I thought that fast-pathing past a nonexistent file might be worth the one extra check (since the OS filesystem caches should make that basically free in the case where the file does get opened later).

Issue #278, opened 2015-12-09
FileIO integration


SimonDanisch:
I see that the file format is registered in FileIO in https://github.com/JuliaIO/FileIO.jl/blob/master/src/registry.jl#L106 
But it seems like there is no load function defined for it in `HDF5` itself (JuliaGL/GLVisualize.jl#44).
Is there a reason for that? 

Best,
Simon

timholy:
No reason other than it hasn't been added.

I think the JLD one is also importing `load` and `save`, and with the redesign it shouldn't be.

SimonDanisch:
Good to know! Well then, I might create a PR if If I find the time.

ggggggggg:
What should these do? `load` seems straightforward, just return an HDF5File. `save` is less, say it is passed an array, what should it call the dataset that it writes to? What happens if you call `save` again, write to a new dataset? Is `save(fname,data)` supposed to be able to be followed by `load(fname)==data` being true?

jonas-schulze:
Are there any updates on this?

I was trying out [DrWatson.jl](https://juliadynamics.github.io/DrWatson.jl/dev/) and saving any `.hdf5` file fails with
```
Error encountered while saving "/Users/jonas/Research/FooProject/data/sim/prefix_N=50.hdf5".
Fatal error:
ERROR: LoadError: UndefVarError: save not defined
```

musm:
Doesn't this work? (I haven't tried it) Looking at https://github.com/JuliaIO/FileIO.jl/blob/master/src/registry.jl#L303 it seems like it's in the registry?

PR #782, opened 2020-12-15
Add FileIO interface

musm:
Still need to figure out what to do with the `save` methods. 

Most of these align with  the JLD2 fileio definitions. 

close https://github.com/JuliaIO/HDF5.jl/issues/278

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -2129,6 +2129,7 @@ function __init__()
     UTF8_ATTRIBUTE_PROPERTIES[]  = create_property(H5P_ATTRIBUTE_CREATE; char_encoding = H5T_CSET_UTF8)
 
     @require MPI="da04e1cc-30fd-572f-bb4f-1f8673147195" @eval include("mpio.jl")
+    @require FileIO="5789e2e9-d7fb-5bc7-8068-2c6fae9b9549" @eval include("fileio.jl")
 
     return nothing
 end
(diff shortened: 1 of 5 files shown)


src/HDF5.jl
```diff
@@ -2034,6 +2034,7 @@ function __init__()
     UTF8_ATTRIBUTE_PROPERTIES[]  = create_property(H5P_ATTRIBUTE_CREATE; char_encoding = H5T_CSET_UTF8)
 
     @require MPI="da04e1cc-30fd-572f-bb4f-1f8673147195" @eval include("mpio.jl")
+    @require FileIO="5789e2e9-d7fb-5bc7-8068-2c6fae9b9549" include("fileio.jl")
```
jmert, line 2037:
```suggestion
    @require FileIO="5789e2e9-d7fb-5bc7-8068-2c6fae9b9549" @eval include("fileio.jl")
```
for same reason as in https://github.com/JuliaIO/HDF5.jl/pull/783#issuecomment-745100252

jonas-schulze:
When do you plan to have the next release? I would really like to use this (without `] add ...#master`). 🙂

musm:
Hopefully soon. This is a new feature and thus requires us to bump the minor version.  We have some really good upcoming changes to HDF5.jl and we have another  PR that I'm hoping to finish reviewing soon, which will overhaul a lot of things for the better  for the v0.16 release.  I can say it shouldn't be much longer. I'd rather have a really good v0.16 release instead of making a minor release with just this feature addition followed by another v0.17 release. Sorry about that, but I think you will also greatly enjoy the new upcoming changes.

PR #783, opened 2020-12-15
Remove at-eval from mpio file load

musm:
@jipolanco  I don't think we need the `@eval` here


jipolanco:
It's not needed, but according to the [Requires documentation](https://github.com/JuliaPackaging/Requires.jl), the `@eval` here enables precompilation of the included code:

> You can even use
> 
> ```julia
> function __init__()
>     @require Gadfly="c91e804a-d5a3-530f-b6f0-dfbca275c004" @eval using MyGluePkg
> end
> ```
> 
> if you wish to exploit precompilation for the new code.

musm:
Ah didin't know that. That sounds like something we want to keep.

Issue #786, opened 2020-12-23
Displaying H5 File Structure Gets Stuck


SebastianAment:
I have an h5 file that can be read as expected if I suppress the REPL output.  However, if I try to display the file structure for the top two HDF5.Group objects in the REPL, it gets stuck and I need to manually interrupt the process to continue. I've attached the stack trace after interrupting after a couple of seconds.

```
Error showing value of type HDF5.Group:
ERROR: InterruptException:
Stacktrace:
 [1] Array at ./boot.jl:406 [inlined]
 [2] Array at ./boot.jl:415 [inlined]
 [3] similar at ./abstractarray.jl:675 [inlined]
 [4] similar at ./abstractarray.jl:674 [inlined]
 [5] _array_for at ./array.jl:678 [inlined]
 [6] collect(::Base.Generator{UnitRange{UInt64},HDF5.var"#30#31"{HDF5.Attributes}}) at ./array.jl:691
 [7] keys(::HDF5.Attributes) at /Users/sebastianament/.julia/packages/HDF5/Vy3Ac/src/HDF5.jl:997
 [8] _tree_children at /Users/sebastianament/.julia/packages/HDF5/Vy3Ac/src/show.jl:179 [inlined]
 [9] _show_tree(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::HDF5.Dataset, ::String; attributes::Bool) at /Users/sebastianament/.julia/packages/HDF5/Vy3Ac/src/show.jl:201
 [10] _show_tree(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::HDF5.Group, ::String; attributes::Bool) at /Users/sebastianament/.julia/packages/HDF5/Vy3Ac/src/show.jl:212 (repeats 3 times)
 [11] _show_tree at /Users/sebastianament/.julia/packages/HDF5/Vy3Ac/src/show.jl:198 [inlined] (repeats 2 times)
 [12] show_tree(::IOContext{REPL.Terminals.TTYTerminal}, ::HDF5.Group; kws::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/sebastianament/.julia/packages/HDF5/Vy3Ac/src/show.jl:220
 [13] show_tree at /Users/sebastianament/.julia/packages/HDF5/Vy3Ac/src/show.jl:219 [inlined]
[...]

PR #787, opened 2020-12-23
Grab keys via callback iteration rather than many absolute calls

jmert:
It appears there's significant overhead in directly obtaining the name
of every sub-element using a loop of `h5?_get_name_by_idx` calls.
For instance, a v7.3-formatted Matlab save file I have access to has
deeply nested and large struct which are serialized into a huge number
of datasets within a `#refs#` group.

```julia
julia> Int(length(fid["#refs#"]))
36236
```

Before this change, enumerating the keys takes almost a minute:
```julia
julia> @time keys(fid["#refs#"]);
 54.193914 seconds (72.47 k allocations: 3.594 MiB)
```
but with this change, that's reduced to a fraction of a second:
```julia
julia> @time keys(fid["#refs#"]);
  0.013353 seconds (36.24 k allocations: 1.383 MiB)
```

This is relevant in the context of #786, though it's not enough on its own.


musm:
LGTM. 

Trying to figure out the v1.12 renaming issue is going to be tricky.

gen/bind_generator.jl
```diff
@@ -6,12 +6,14 @@ const bind_exceptions = Dict{Symbol,Symbol}()
...
 # Version-numbered identifiers
 push!(bind_exceptions, :h5a_create => :H5Acreate2)
+push!(bind_exceptions, :h5a_iterate => :H5Aiterate2)
 push!(bind_exceptions, :h5d_create => :H5Dcreate2)
 push!(bind_exceptions, :h5d_open => :H5Dopen2)
 push!(bind_exceptions, :h5e_get_auto => :H5Eget_auto2)
 push!(bind_exceptions, :h5e_set_auto => :H5Eset_auto2)
 push!(bind_exceptions, :h5g_create => :H5Gcreate2)
 push!(bind_exceptions, :h5g_open => :H5Gopen2)
+push!(bind_exceptions, :h5l_iterate => :H5Literate1)
```
musm, line 16: If we get rid of this, will this work on v1.12 ? The documentation make it sound like it should still work, but it might depend on how the underlying libraries were built

https://confluence.hdfgroup.org/display/HDF5/H5L_ITERATE
> The function H5L_ITERATE was renamed to H5L_ITERATE1 and deprecated in this release.The macro H5L_ITERATE and the function H5L_ITERATE2 were introduced in this release.
--

src/api_helpers.jl
```diff
@@ -37,6 +37,13 @@ function h5a_get_name_by_idx(loc_id, obj_name, idx_type, order, idx, lapl_id)
     return String(buf)
 end
 
+function h5a_iterate(f, obj_id, idx_type, order, idx = 0)
+    idxref = Ref{hsize_t}(idx)
+    fptr = @cfunction($f, herr_t, (hid_t, Ptr{Cchar}, Ptr{H5A_info_t}, Ptr{Cvoid}))
```
musm, line 42: This needs a runtime closure?

gen/bind_generator.jl
jmert: The problem is they implement the rename as C header `#define`s, so the symbol doesn't really exist in the library.
```julia
julia> dlsym(HDF5_jll.libhdf5_handle, :H5Literate)
ERROR: could not load symbol "H5Literate":
/home/justin/.julia/artifacts/997813d46a8a06e6e9871a2a01483f91ce954eca/lib/libhdf5.so: undefined symbol: H5Literate
Stacktrace:
 [1] #dlsym#1
   @ ./libdl.jl:56 [inlined]
 [2] dlsym(hnd::Ptr{Nothing}, s::Symbol)
   @ Base.Libc.Libdl ./libdl.jl:54
 [3] top-level scope
   @ REPL[27]:1

julia> dlsym(HDF5_jll.libhdf5_handle, :H5Literate1)
Ptr{Nothing} @0x00007ff367429980
```

src/api_helpers.jl
jmert: I'm not sure — I did my best to interpret the documentation, and I thought from the [`@cfunction`](https://docs.julialang.org/en/v1/base/c/#Base.@cfunction) docs saying
```
And that these arguments will be evaluated in global scope during compile-time (not deferred until runtime).
```
that handling the closures would be needed. I can try without.

musm: I see it sounds correct then

gen/bind_generator.jl
musm: Darn it...

src/api_helpers.jl
jmert: I just confirmed that removing the `$` results in an `f not defined` error get thrown during precompilation.

gen/bind_generator.jl
jmert: So I've done some more generator macro hacking to try to work around this.

Second commit makes the generator a bit smarter in dealing with versioned symbols so that there are only a few exceptions in the `bind_exceptions` Dict.

The third commit is what actually does the version-specific hacking. I'm not convinced yet that there isn't a smarter way to go about this, but this is what came to mind and I could make work.

docs/src/api_bindings.md
```diff
@@ -122,6 +123,8 @@ h5l_delete(obj_id::hid_t, pathname::Ptr{UInt8}, lapl_id::hid_t)
 h5l_exists(loc_id::hid_t, pathname::Ptr{UInt8}, lapl_id::hid_t)
 h5l_get_info(link_loc_id::hid_t, link_name::Ptr{UInt8}, link_buf::Ptr{H5L_info_t}, lapl_id::hid_t)
 h5l_get_name_by_idx(loc_id::hid_t, group_name::Ptr{UInt8}, index_field::Cint, order::Cint, n::hsize_t, name::Ptr{UInt8}, size::Csize_t, lapl_id::hid_t)
+h5l_iterate(group_id::hid_t, idx_type::Cint, order::Cint, idx::Ptr{hsize_t}, op::Ptr{Cvoid}, op_data::Ptr{Cvoid})
+h5l_iterate(group_id::hid_t, idx_type::Cint, order::Cint, idx::Ptr{hsize_t}, op::Ptr{Cvoid}, op_data::Ptr{Cvoid})
```
jmert, line 127: Oops. I'll have to figure out how to fix that.

gen/api_defs.jl
```diff
@@ -129,14 +131,18 @@
 @bind h5l_exists(loc_id::hid_t, pathname::Ptr{UInt8}, lapl_id::hid_t)::htri_t error("Cannot determine whether ", pathname, " exists")
 @bind h5l_get_info(link_loc_id::hid_t, link_name::Ptr{UInt8}, link_buf::Ptr{H5L_info_t}, lapl_id::hid_t)::herr_t error("Error getting info for link ", link_name)
 @bind h5l_get_name_by_idx(loc_id::hid_t, group_name::Ptr{UInt8}, index_field::Cint, order::Cint, n::hsize_t, name::Ptr{UInt8}, size::Csize_t, lapl_id::hid_t)::Cssize_t "Error getting object name"
+# libhdf5 v1.10 provides the name H5Literate
+# libhdf5 v1.10 provides the same under H5Literate1, and a newer interface on H5Literate2
```
jmert, line 135:
```suggestion
# libhdf5 v1.12 provides the same under H5Literate1, and a newer interface on H5Literate2
```

gen/bind_generator.jl
```diff
@@ -114,6 +98,11 @@ macro bind(sig::Expr, err::Union{String,Expr,Nothing} = nothing)
     else
         # turn e.g. h5f_close into H5Fclose
         cfuncname = Symbol(uppercase(prefix), rest)
+        # Remove the version number if present
+        if occursin(r"\d$", String(jlfuncname))
```
musm, line 144: what if we change this to
```
occursin(r"\d(?<!hdf5)$", String(jlfuncname))  #(negative lookbehind to exclude hdf5 at the end)
```
or 
```
occursin(r"\d$", String(jlfuncname)) && !occursin(r"hdf5$", String(jlfuncname))
```

so we can interely get rid of the exceptions list?

gen/bind_generator.jl
```diff
@@ -80,7 +78,35 @@ It is assumed that the HDF library names are given in global constants named `li
 and `libhdf5_hl`. The former is used for all `ccall`s, except if the C library name begins
 with "H5DO" or "H5TB" then the latter library is used.
 """
-macro bind(sig::Expr, err::Union{String,Expr,Nothing} = nothing)
+macro bind(sig::Expr, err::Union{String,Expr,Nothing} = nothing,
+           vers::Union{Expr,Nothing} = nothing)
+    expr = _bind(__module__, __source__, sig, err)
```
musm, line 83: What's the rationale for  passing  __module, __source__ in this context?

musm: Nvm, I just read the comment regarding `__source__`  lower in the file.

jmert: The `__source__` is used just before the end of the function. I included `__module__` just because it makes the function call basically equivalent to the original macro definition.

musm: Albeit we strip the line number anyways, but they are probably helpful for debugging?  (no objections, just nothing this)

jmert: Yeah, in the grand scheme, we still strip away that line node when writing out `api.jl`, but in principle the macro maintains some source information if you're to use it interactively.

musm: Although it looks like we end up spitting the same line number information twice.  

E.g.
```
#= api_defs.jl:22 =#

function h5_close()
    #= api_defs.jl:22 =#
    var"#status#" = ccall((:H5close, libhdf5), herr_t, ())
    var"#status#" < 0 && error("Error closing the HDF5 resources")
    return nothing
end
```

I'm not sure where that first line-number information is coming from.

jmert: I think the first one is added during parsing of the `api_defs.jl`-contents String:
```julia
julia> defs = read("gen/api_defs.jl", String);

julia> exprs = Meta.parseall(defs, filename = "api_defs.jl"); # called within Base.include_string

julia> exprs.head = :block; # from :toplevel to :block to get nicer printing

julia> exprs
...
    #= api_defs.jl:307 =#
    #= api_defs.jl:307 =# @bind h5tb_read_table(loc_id::hid_t, table_name::Ptr{UInt8}, dst_size::Csize_t, dst_offset::Ptr{Csize_t}, dst_sizes::Ptr{Csize_t}, dst_buf::Ptr{Cvoid})::herr_t "Error reading table"
    #= api_defs.jl:308 =#
    #= api_defs.jl:308 =# @bind h5tb_write_records(loc_id::hid_t, table_name::Ptr{UInt8}, start::hsize_t, nrecords::hsize_t, type_size::Csize_t, field_offsets::Ptr{UInt8}, field_sizes::Ptr{UInt8}, data::Ptr{Cvoid})::herr_t "Error writing record to table"
    #= api_defs.jl:314 =#
    #= api_defs.jl:314 =# @bind h5z_register(filter_class::Ref{H5Z_class_t})::herr_t "Unable to register new filter"
end
```

musm:
```diff
diff --git a/gen/gen_wrappers.jl b/gen/gen_wrappers.jl
index 8a53685..89ce993 100644
--- a/gen/gen_wrappers.jl
+++ b/gen/gen_wrappers.jl
@@ -52,9 +52,11 @@ for (mod, desc, urltail) in (
     ("H5TB", "Table Interface", "Tables"),
     )
     global apidocs
-    funcs = join(sort!(bound_api[mod]), "\n")
+    funcs = sort!(unique(bound_api[mod])) # unique removes duplicate methods (occurs when we have multiply defined methods due to hdf5 compatibility)
+    funcs = join(funcs, "\n```\n\n```julia\n")
     apidocs *= """
         ## [`$mod`](https://portal.hdfgroup.org/display/HDF5/$urltail) — $desc
+
         ```julia
         $funcs
         ```
```

I think it is   better to give each method its own code block, which mimics how method docstrings typically printed out.

musm:
Also these changes should make it easy to implement an efficient iteration protocol over keys.

gen/bind_generator.jl
```diff
@@ -4,29 +4,26 @@ using Base.Meta: isexpr, quot
...
 # Distinguishes 32-bit vs 64-bit handle arguments
 push!(bind_exceptions, :h5p_get_fapl_mpio32 => :H5Pget_fapl_mpio)
 push!(bind_exceptions, :h5p_get_fapl_mpio64 => :H5Pget_fapl_mpio)
 push!(bind_exceptions, :h5p_set_fapl_mpio32 => :H5Pset_fapl_mpio)
 push!(bind_exceptions, :h5p_set_fapl_mpio64 => :H5Pset_fapl_mpio)
 
+# An expression which is injected at the beginning of the API defitions to aid in doing
+# (pre)compile-time conditional compilation based on the libhdf5 version.
+_libversion_build_expr = quote
+    _libversion_build = let
```
musm, line 16: Maybe rename this to something that denotes it refers to the libhdf5 library, e.g.
```
_libhdf5_build_ver
```

musm:
Alternative to  calling `unique` in https://github.com/JuliaIO/HDF5.jl/pull/787#issuecomment-750923410 
We might just want to add the filtering in the `@bind` macro generator phase


```diff
diff --git a/gen/bind_generator.jl b/gen/bind_generator.jl
index 0404b30..c27134e 100644
--- a/gen/bind_generator.jl
+++ b/gen/bind_generator.jl
@@ -147,7 +147,7 @@ function _bind(__module__, __source__, sig::Expr, err::Union{String,Expr,Nothing

     # Store the function prototype in HDF5-module specific lists:
     funclist = get!(bound_api, uppercase(prefix), Vector{String}(undef, 0))
-    push!(funclist, string(funcsig))
+    !(string(funcsig) in funclist) && push!(funclist, string(funcsig))  # duplicate method may occur when we have multiply defined methods due to hdf5 compatibility bounds

     # Determine the underlying C library to call
     lib = startswith(string(cfuncname), r"H5(DO|DS|LT|TB)") ? :libhdf5_hl : :libhdf5
```

musm:
LGTM I'd just squash  and rearange the commits that deal with the binding process

musm:
> The third commit is what actually does the version-specific hacking. I'm not convinced yet that there isn't a smarter way to go about this, but this is what came to mind and I could make work.

Yeah, I'm sure we could make it more elegant or use a different format to parse the compatibility bounds, but it's simple enough and I couldn't think of anything better.

jmert:
> LGTM I'd just squash and rearange the commits that deal with the binding process

I've rebased and squashed to just three commits — first changes the key iteration implementation, second gets rid of most of the binding exceptions by having the generator automatically handle version-number suffixes, and the third adds the ability to specify that a particular binding is libhdf5-version-dependent.

Is that the level of squashing you were thinking of?

musm:
Thanks! This is indeed a much needed change,  and I'm glad we now have a way to address different library verisons.

PR #788, opened 2020-12-24
Add some print-limiting heuristics to default show

jmert:
This is a partial fix for #786, with the more efficient key iteration in #787 also required to avoid a very long pause on my MAT v7.3 file test case (though it does not actually finish printing, which is a significant improvement in itself).

I'm open to suggestions for whether these heuristics are sufficient and what the thresholds should be set to.

(A future TODO is to replace the `_tree_children` list building and loops with direct iteration via `h5a_iterate` and `h5l_iterate`, which should be more efficient yet.)

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -126,6 +126,20 @@ Configurable option to control whether emoji icons (`true`) or a plain-text anno
 """
 const SHOW_TREE_ICONS = Ref{Bool}(true)
 
+"""
+    SHOW_TREE_MAX_DEPTH = Ref{Int}(5)
+
+Maximum recursive depth to descend during printing.
+"""
+const SHOW_TREE_MAX_DEPTH = Ref{Int}(5)
+
+"""
+    SHOW_TREE_MAX_CHILDREN = Ref{Int}(50)
+
+Maximum number of children to show at each node.
+"""
+const SHOW_TREE_MAX_CHILDREN = Ref{Int}(50)
+
 function Base.show(io::IO, ::MIME"text/plain", obj::Union{File,Group,Dataset,Attributes,Attribute})
     if SHOW_TREE[]
         show_tree(io, obj)
@@ -155,6 +169,13 @@ _tree_icon(obj::Attributes) = _tree_icon(obj.parent)
 
 _tree_head(io::IO, obj) = print(io, _tree_icon(obj), " ", obj)
 _tree_head(io::IO, obj::Datatype) = print(io, _tree_icon(obj), " HDF5.Datatype: ", name(obj))
+
+_tree_count(parent::Union{File,Group}, attributes::Bool) =
+    length(parent) + (attributes ? length(HDF5.attributes(parent)) : 0)
+_tree_count(parent::Dataset, attributes::Bool) =
+    attributes ? length(HDF5.attributes(parent)) : 0
+_tree_count(parent::Attributes, _::Bool) = length(parent)
+_tree_count(parent::Union{Attribute,Datatype}, _::Bool) = 0
 
 function _tree_children(parent::Union{File, Group}, attributes::Bool)
     names = keys(parent)
@@ -189,13 +210,26 @@ function _tree_children(parent::Union{Attribute, Datatype}, attributes::Bool)
 end
 
 function _show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Attribute}, indent::String="";
-                    attributes::Bool = true)
+                    attributes::Bool = true, depth::Int = 1)
     isempty(indent) && _tree_head(io, obj)
-    !isvalid(obj) && return
+    isvalid(obj) || return
+
+    childstr(io, n, more=" ") = print(io, "\n", indent, "└─ (", n, more,
+            n == 1 ? "child" : "children", ")")
+
+    nchildren = _tree_count(obj, attributes)
+    if nchildren > 0 && depth > SHOW_TREE_MAX_DEPTH[]
+        childstr(io, nchildren)
+        return
+    end
 
     names, children = _tree_children(obj, attributes)
-    nchildren = length(children)
     for ii in 1:nchildren
+        if ii > max(2, SHOW_TREE_MAX_CHILDREN[] ÷ depth)
+            childstr(io, nchildren - ii + 1, " more ")
+            break
+        end
+
         name = names[ii]
         child  = children[ii]
 
@@ -204,7 +238,7 @@ function _show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,At
         print(io, "\n", indent, islast ? "└─ " : "├─ ", icon, " ", name)
 
         nextindent = indent * (islast ? "   " : "│  ")
-        _show_tree(io, child, nextindent; attributes = attributes)
+        _show_tree(io, child, nextindent; attributes = attributes, depth = depth + 1)
     end
     return nothing
 end
--- a/src/show.jl
+++ b/src/show.jl
@@ -112,13 +112,6 @@ function Base.show(io::IO, dspace::Dataspace)
 end
 
 """
-    SHOW_TREE = Ref{Bool}(true)
-
-Configurable option to control whether the default `show` for HDF5 objects is printed
-using `show_tree` or not.
-"""
-const SHOW_TREE = Ref{Bool}(true)
-"""
     SHOW_TREE_ICONS = Ref{Bool}(true)
 
 Configurable option to control whether emoji icons (`true`) or a plain-text annotation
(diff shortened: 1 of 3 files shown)


src/show.jl
```diff
@@ -204,7 +238,9 @@ function _show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,At
         print(io, "\n", indent, islast ? "└─ " : "├─ ", icon, " ", name)
 
         nextindent = indent * (islast ? "   " : "│  ")
-        _show_tree(io, child, nextindent; attributes = attributes)
+        # To limit amount of printing, only propagate `attributes` to Groups
+        attributes′ = attributes && child isa Group
```
musm, line 242: I'm not sure how  I feel abou this exception.

src/show.jl
```diff
@@ -189,13 +210,26 @@ function _tree_children(parent::Union{Attribute, Datatype}, attributes::Bool)
 end
 
 function _show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Attribute}, indent::String="";
-                    attributes::Bool = true)
+                    attributes::Bool = true, depth::Int = 1)
     isempty(indent) && _tree_head(io, obj)
     !isvalid(obj) && return
```
musm, line 215: unrelated, but this reads better 
```suggestion
    isvalid(obj) || return
```
I think usually with `&&` or `||`, the  first  check should evaluate to `true`  for  action.

src/show.jl
jmert: Sure — I threw that in because recursing through my example very large v7.3 MAT file ends up printing a bunch of `└─ (1 child)` lines once it reaches the depth limit that just ends up being really ugly due to every element having at least a `MATLAB_class` attribute, but I have no problem dropping this change. It can always be debated later how to more extensively tweak the heuristics.

jmert:
I just added another commit which makes use of `h5a_iterate` and `h5l_iterate` to further speed up the normal printing. The commit message for that commit in particular includes benchmarks on my large test file, where I saw a factor of ~100× speed up:

Before
```julia
julia> IOBuffer(sprint(HDF5.show_tree, hfile)) |> countlines
396

julia> @btime HDF5.show_tree(buf, hfile) setup = (buf = IOBuffer());
  824.235 ms (76435 allocations: 3.59 MiB)
```

After:
```julia
julia> IOBuffer(sprint(HDF5.show_tree, hfile)) |> countlines
396

julia> @btime HDF5.show_tree(buf, hfile) setup = (buf = IOBuffer());
  7.313 ms (3035 allocations: 158.80 KiB)
```

src/show.jl
```diff
@@ -134,77 +148,84 @@ function Base.show(io::IO, ::MIME"text/plain", obj::Union{File,Group,Dataset,Att
...
+        print(io, "\n", indent, ELBOW * "(", n, more, n == 1 ? "child" : "children", ")")
     end
-    return (names, objs)
-end
-function _tree_children(parent::Dataset, attributes::Bool)
-    names = String[]
-    objs = Union{Object, Attribute}[parent[n] for n in names]
+    @inline function depth_check()
+        counter += 1
+        if counter > max(2, SHOW_TREE_MAX_CHILDREN[] ÷ depth)
```
musm, line 188: Is your intention that this print  at depth 1 this prints 50, then 25, 12, 6, 3, 2   children at their respective depths?

src/show.jl
```diff
@@ -134,77 +148,84 @@ function Base.show(io::IO, ::MIME"text/plain", obj::Union{File,Group,Dataset,Att
...
+_tree_count(parent::Union{File,Group}, attributes::Bool) =
+    length(parent) + (attributes ? length(HDF5.attributes(parent)) : 0)
+_tree_count(parent::Dataset, attributes::Bool) =
+    attributes ? length(HDF5.attributes(parent)) : 0
+_tree_count(parent::Attributes, _::Bool) = length(parent)
+_tree_count(parent::Union{Attribute,Datatype}, _::Bool) = 0
+
+function _show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Attribute}, indent::String="";
+                    attributes::Bool = true, depth::Int = 1)
+    isempty(indent) && _tree_head(io, obj)
```
musm, line 172:
```suggestion
    limit = get(io, :limit, false)::Bool
    isempty(indent) && _tree_head(io, obj)
```

src/show.jl
musm, line 188:
```suggestion
        if limit && counter > max(2, SHOW_TREE_MAX_CHILDREN[] ÷ depth)
```

src/show.jl
```diff
@@ -134,77 +148,84 @@ function Base.show(io::IO, ::MIME"text/plain", obj::Union{File,Group,Dataset,Att
...
+    @inline function depth_check()
+        counter += 1
+        if counter > max(2, SHOW_TREE_MAX_CHILDREN[] ÷ depth)
+            childstr(io, nchildren - counter + 1, " more ")
+            return true
+        end
+        return false
+    end
+
+    if nchildren > 0 && depth > SHOW_TREE_MAX_DEPTH[]
```
musm, line 195:
```suggestion
    if limit && nchildren > 0 && depth > SHOW_TREE_MAX_DEPTH[]
```

musm:
I'm still going through some alternative limiting logic. 
If you're fine with that, this may take longer to finish/review. Or we can do follow up PRs.

src/show.jl
jmert: Yeah, something like that. I thought that it would balance wanting to see a reasonable number of direct children on the printed object while still avoiding very long outputs if your 50 direct children each have 50 children of their own (repeated down to the max-depth levels).

jmert:
> I'm still going through some alternative limiting logic.
> If you're fine with that, this may take longer to finish/review. Or we can do follow up PRs.

Yeah, that should be fine — I assume the `get(io, :limit, false)` your starting to add suggestions for is the same print-limiting that Base uses for Array printing, and setting the key is a standard way of disabling?

musm:
That's right and by using the established way of controlling printing it would  allow us to remove  a  `Ref` that controls the printing of the tree here.  For example, suppose you wanted to print the whole tree to a file, with the `limit` key that allows you to control that in the io Context.

src/show.jl
```diff
@@ -126,85 +119,107 @@ Configurable option to control whether emoji icons (`true`) or a plain-text anno
...
+        return nothing
+    end
+
     if attributes
-        attrn = keys(HDF5.attributes(parent))
-        attro = Union{Object, Attribute}[HDF5.attributes(parent)[n] for n in attrn]
-        names = append!(attrn, names)
-        objs  = append!(attro, objs)
+        obj′ = obj isa Attributes ? obj.parent : obj
+        h5a_iterate(obj′, H5_INDEX_NAME, H5_ITER_INC) do _, cname, _, _
```
musm, line 196: Is it worth it to set up this custom call-back instead of directly iterating over the attributes, which would be more readable and simpler to maintain?

src/show.jl
```diff
@@ -126,85 +119,107 @@ Configurable option to control whether emoji icons (`true`) or a plain-text anno
...
+_tree_count(parent::Union{File,Group}, attributes::Bool) =
+    length(parent) + (attributes ? length(HDF5.attributes(parent)) : 0)
+_tree_count(parent::Dataset, attributes::Bool) =
+    attributes ? length(HDF5.attributes(parent)) : 0
+_tree_count(parent::Attributes, _::Bool) = length(parent)
+_tree_count(parent::Union{Attribute,Datatype}, _::Bool) = 0
+
+function _show_tree(io::IO, obj::Union{File,Group,Dataset,Datatype,Attributes,Attribute}, indent::String="";
+                    attributes::Bool = true, depth::Int = 1)
+    isempty(indent) && _tree_head(io, obj)
```
musm, line 165: I feel like we should instead use `displaysize(io)` to determine a suitable truncation point instead of the heuristics.

src/show.jl
jmert: Do you mean a custom `h5a_iterate(::Attributes, ...)`? I'd thought about that, but didn't since it could also make sense to instead define `Base.cconvert(::Type{hid_t}, a::Attributes) = cconvert(hid_t, a.parent)`, but I hadn't thought/dug through the implications of enabling that yet.

src/show.jl
jmert: I agree in spirit, though it'll probably take another revamp of this PR to do that. I think more contextual information will have to be sent through the recursions — let me mull that over.

src/show.jl
musm: I guess we don't have a way to directly iterate over Attributes as in https://github.com/JuliaIO/HDF5.jl/blob/master/src/HDF5.jl#L956

src/show.jl
musm: Yeah the recursive tree printing makes this  more complicated.  Feel free to ignore this comment for now.

musm:
LGTM, feel free to merge

src/show.jl
jmert: Oh, the other thing about the iteration is that at least for attributes, we don't ever actually have to open the objects — just having the name is sufficient. The [hypothetical] `Base.iterate` method returns an object, so that's extra allocations/libhdf5 overhead we can avoid by using `h5a_iterate` instead.

src/show.jl
jmert: We can open an issue to make sure this goal isn't forgotten.

PR #790, opened 2020-12-28
Remove v0.14 deprecations

musm:


jmert:
The MAT.jl integration failure is real — it extends the deprecated `HDF5.exists` method so that MAT and JLD both make method specializations on the same function.

We can either (a) just provide an empty `function exists end` just for MAT and JLD to extend, or (b) move JLD and MAT to new major versions that switch to the `keys`/`haskey` interface that HDF5 has moved to.

musm:
I think (b) makes more sense, PR's created to address those issues.

musm:
Ok  I decided to add the empty `function exists end` method, so we can at least have `MAT,JLD` work  with HDF5 v0.15 in backward compatible way, which I think is important since v0.15 will have important fixes and this will smooth over the deprecation process when we bump everyhing again.   After 0.15 is released, we can make PRs in MAT,JLD to  bump the minor version and remove  `exists`. 

So I think this should be ok:

1. Release HDF5 v0.15
2. Add compatability of HDF5 v0.15 to JLD,MAT  in their `Project.toml` and bump their patch version.
2. Remove the deprecations for `exists` in MAT,JLD and bump their minor version (these versions will be compatible with HDF5 v0.14, v0.15, v0.16)
3. Remove these deprecation in HDF5 v0.16

jmert (approved):
I think that plan makes sense.

PR #792, opened 2020-12-31
Have the binding generator build automatic doc strings for all functions

jmert:
This package currently has pretty low coverage of doc strings on functions and types — this PR makes the easy-ish first step of using the binding generator to automatically build doc strings for all of the low-level API functions.

One of the features I'm particular I'm excited about is that the upstream doc URL can be printed for each function, so it's much faster to look up the detailed behavior than having to pull out a web search each time manually.

The version-conditioned expressions posed the most problems for me and took a while to sort out — both pretty-printing the doc string into the output `src/api.jl` was difficult, and then it took me a while to also realize that they need `@doc ` prefixes (otherwise they're apparently just dropped during precompilation and never actually get registered as docs).


jmert:
Replaced by #795

PR #794, opened 2021-01-01
[testing]

musm:
test if PR previews work with documenter.


musm:
Ok this is working, I just don't think they work with forks

jmert:
Would you like me to re-push #792 to JuliaIO/HDF5.jl rather than my fork to get the doc previews?

musm:
That works with me.

I'm confident the documentation looks good (I'll likely check it out locally to test out anyways), but it is nice to have previews working.

PR #795, opened 2021-01-01
Have the binding generator build automatic doc strings for all functions

jmert:
*Replaces #792, to have doc previews available on this PR.*

This package currently has pretty low coverage of doc strings on functions and types — this PR makes the easy-ish first step of using the binding generator to automatically build doc strings for all of the low-level API functions.

One of the features I'm particular I'm excited about is that the upstream doc URL can be printed for each function, so it's much faster to look up the detailed behavior than having to pull out a web search each time manually.

The version-conditioned expressions posed the most problems for me and took a while to sort out — both pretty-printing the doc string into the output src/api.jl was difficult, and then it took me a while to also realize that they need @doc prefixes (otherwise they're apparently just dropped during precompilation and never actually get registered as docs).


jmert:
https://juliaio.github.io/HDF5.jl/previews/PR795/api_bindings/ :+1: Seems to have worked.

musm:
Can we add an Index to each section?

It'll just require adding: 
````
```@index
```
````

**Comment by jmert** [comment]
Testing locally, that seems to add the entire list of documented functions, not just the functions within the section. I assume that's not what you want?

**Comment by musm** [comment]
Yikes, let me take a look.

**Comment by jmert** [comment]
I think this does what you want:
```diff
diff --git a/gen/gen_wrappers.jl b/gen/gen_wrappers.jl
index 2db2714..980fda4 100644
--- a/gen/gen_wrappers.jl
+++ b/gen/gen_wrappers.jl
@@ -79,9 +79,12 @@ for (mod, desc, urltail) in (
     ("H5TB", "Table Interface", "Tables"),
     )
     global apidocs
-    funcs = join(sort!(bound_api[mod]), "\n")
+    funclist = sort!(bound_api[mod])
+    index = join(["- [`$f`](@ref HDF5.$f)" for f in funclist], "\n")
+    funcs = join(funclist, "\n")
     apidocs *= """
         ## [`$mod`](https://portal.hdfgroup.org/display/HDF5/$urltail) — $desc
+        $index
         ```@docs
         $funcs
         ```
```

jmert:
Oh, just noticed that this now adds a *huge* list to the bottom of the main docs page: https://juliaio.github.io/HDF5.jl/previews/PR795/#API-Reference

jmert:
This fixes most of it:
````diff
diff --git a/docs/make.jl b/docs/make.jl
index e6cdcf4..da960a1 100644
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -2,6 +2,11 @@ using Documenter
 using HDF5
 using MPI  # needed to generate docs for parallel HDF5 API
 
+# Used in index.md to filter the autodocs list
+wherefrom(m::Method) = String(m.file)
+not_from_api(m::Method) = !endswith(wherefrom(m), "src/api.jl")
+not_from_api(f::Function) = all(not_from_api, methods(f))
+not_from_api(o) = true
 
 makedocs(;
     modules=[HDF5],
diff --git a/docs/src/index.md b/docs/src/index.md
index 62bce8a..0b3ba5f 100644
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -546,8 +546,10 @@ when reading or writing.
 Below we include a limited number of API references. Note not all of these are public interfaces, thus proceed with caution.
 
 ```@index
+Pages = ["index.md"]
 ```
 
 ```@autodocs
 Modules = [HDF5]
+Filter = not_from_api
 ```
````
The manually-defined low-level API functions (such as `h5p_get_class_name`) still get added because they're defined in `api_helpers.jl` — they could also be excluded by defining `not_from_api(::typeof(HDF5.h5p_get_class_name)) = false`, I suppose.

**Comment by musm** [comment]
I don't have an issue with how the docstring is built up in this PR,  but wouldn't this be more direct/simpler? 

```diff
-    docstr = "    $docfunc"
+    docstr = "\"\"\"\n    $docfunc"

     # Determine the underlying C library to call
     lib = startswith(string(cfuncname), r"H5(DO|DS|LT|TB)") ? :libhdf5_hl : :libhdf5
@@ -198,7 +198,7 @@ function _bind(__module__, __source__, sig::Expr, err::Union{String,Expr,Nothing
     end

     docstr *= "\n\nSee `libhdf5` documentation for [`$cfuncname`]" *
-              "(https://portal.hdfgroup.org/display/HDF5/$(uppercase(string(funcsig.args[1])))).\n"
+              "(https://portal.hdfgroup.org/display/HDF5/$(uppercase(string(funcsig.args[1])))).\n\"\"\"\n"

     # Then assemble the pieces. Doing it through explicit Expr() objects
     # avoids inserting the line number nodes for the macro --- the call site
diff --git a/gen/gen_wrappers.jl b/gen/gen_wrappers.jl
index 2db2714..ee9c965 100644
--- a/gen/gen_wrappers.jl
+++ b/gen/gen_wrappers.jl
@@ -6,7 +6,7 @@ defs = read(joinpath(@__DIR__, "api_defs.jl"), String)
 exprs = Base.include_string(@__MODULE__, "@macroexpand1 begin\n" * defs * "\nend", "api_defs.jl")
 # Insert the conditional version helper expression
 prepend!(exprs.args, _libhdf5_build_ver_expr.args)
-exprs = Base.remove_linenums!(exprs)
+Base.remove_linenums!(exprs)

 # Definitions which are not automatically generated, but should still be documented as
 # part of the raw low-level API:
@@ -24,19 +24,18 @@ open(joinpath(@__DIR__, "..", "src", "api.jl"), "w") do fid
 # To add new bindings, define the binding in `gen/api_defs.jl`, re-run
 # `gen/gen_wrappers.jl`, and commit the updated `src/api.jl`.
 """)
-    function triplequote(s::String, indent = "", prefix = "")
-        ret = indent * prefix * "\"\"\"\n"
+    function indentstring(s::String, indent = "")
+        ret = ""
         for l in eachline(IOBuffer(s))
             ret *= isempty(l) ? "\n" : indent * l * "\n"
         end
-        ret *= indent * "\"\"\"\n"
         return ret
     end
     for funcblock in exprs.args
         if isexpr(funcblock, :block, 2) && funcblock.args[1] isa String
             # Pretty print the doc string — default printing produces the string in just
             # one line with `\n` literals, so manually switch to triple quoted string.
-            print(fid, triplequote(funcblock.args[1]))
+            print(fid, funcblock.args[1])
             print(fid, funcblock.args[2], "\n\n")
         elseif isexpr(funcblock, :macrocall, 3) && funcblock.args[1] === Symbol("@static") &&
                isexpr(funcblock.args[3], :if, 2) &&
@@ -47,7 +46,7 @@ open(joinpath(@__DIR__, "..", "src", "api.jl"), "w") do fid
             docstr = funcblock.args[3].args[2].args[1]
             funcblock.args[3].args[2].args[1] = "SENTINEL_DOC"
             buf = sprint(print, funcblock)
-            buf = replace(buf, r"^\s+\"SENTINEL_DOC\"\n"m => triplequote(docstr, " "^8, "@doc "))
+            buf = replace(buf, r"^\s+\"SENTINEL_DOC\"\n"m => indentstring("@doc " * docstr, " "^8))
             print(fid, buf, "\n\n")
         else
```

jmert:
I thought the way I'd defined it would allow interactive `@bind` use to have documentation (while your suggestions would result in the docstring containing a triple-quoted string itself), but it appears I didn't achieve that goal and the docs are always dropped when used iteractively.

It seems doc string interpretation is special to parsing, and the AST representation isn't doing what I thought. It looks like to get what I wanted, everything needs a `@doc` macro at the AST level, and then the output printer can undo that back to juxtaposed strings in the unconditional case.

docs/src/index.md
```diff
@@ -546,8 +546,10 @@ when reading or writing.
 Below we include a limited number of API references. Note not all of these are public interfaces, thus proceed with caution.
 
 ```@index
+Pages = ["index.md"]
```
musm, line 549: Unintentional ?

docs/make.jl
```diff
@@ -2,6 +2,15 @@ using Documenter
 using HDF5
 using MPI  # needed to generate docs for parallel HDF5 API
 
+# Used in index.md to filter the autodocs list
+wherefrom(m::Method) = String(m.file)
+not_from_api(m::Method) = !endswith(wherefrom(m), "src/api.jl")
+not_from_api(f::Function) = all(not_from_api, methods(f))
+not_from_api(o) = true
+# Manually-defined low-level API (in source file src/api_helpers.jl)
+not_from_api(::typeof(HDF5.h5p_get_class_name)) = false
```
musm, line 11: naming is a bit confusing, perhaps `not_low_level_api`?

docs/src/index.md
jmert: Intentional — I think it's required to ignore all the documented functions in `api_bindings.md` (but maybe I'm misreading Documenter.jl's docs).

musm:
That's fine, I guess it might be easier to work on the string in the binding generator if need be in the future.

Aside from 2 commnets LGTM.

docs/src/index.md
musm: No you're right here.

docs/make.jl
```diff
@@ -2,6 +2,15 @@ using Documenter
 using HDF5
 using MPI  # needed to generate docs for parallel HDF5 API
 
+# Used in index.md to filter the autodocs list
+wherefrom(m::Method) = String(m.file)
+not_from_api(m::Method) = !endswith(wherefrom(m), "src/api.jl")
```
musm, line 7: mabye not worth to add the `wherefrom` method 
```suggestion
not_from_api(m::Method) = !endswith(wherefrom(String(m.file)), "src/api.jl")
```

jmert:
Just a thought for the future — the macro and expression parsing is getting complex enough that it's getting pretty hard to read; relying on `MacroTools` instead as a generator dependency could make everything clearer.

musm:
> Just a thought for the future — the macro and expression parsing is getting complex enough that it's getting pretty hard to read; relying on MacroTools instead as a generator dependency could make everything clearer.

Agreed, the only way I now understand these is by manually going to each line to verify what's happening.
MacroTools would definitely help simplify  the matching.   I'm not sure what else we could use to help simplify pretty-printing the raw-expressions to a text file. 

LGTM on the latest changes.

PR #797, opened 2021-01-05
Improve error handling

musm:


musm:
I think with this change this covers most of the nasty HDF5 error messages being printed in the Julia REPL.

jmert (approved):
LGTM

PR #798, opened 2021-01-05
Deprecate get_dims and set_dims!

musm:
`Base.size` is what people will likely want instead of these methods. 

I figured it would be a good idea to either remove these methods or rename them. Internally, we don't use them after upgrading `Base.size`

@jmert  thoughts on this change (also open to other suggestions you might have  in leu of these changes)

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1107,21 +1107,6 @@ function isnull(obj::Union{Dataspace,Dataset,Attribute})
     return ret
 end
 
-"""
-    HDF5.get_simple_extent_dims(obj::Union{HDF5.Dataspace, HDF5.Dataset,HDF5.Attribute}) -> dims, maxdims
-
-Get the array dimensions from a dataset or attribute and return a tuple of dims and maxdims.
-"""
-function get_simple_extent_dims(obj::Union{Dataspace,Dataset,Attribute})
-    dspace = obj isa Dataspace ? checkvalid(obj) : dataspace(obj)
-    h5_dims, h5_maxdims = h5s_get_simple_extent_dims(dspace)
-    # reverse dimensions since hdf5 uses C-style order
-    N = length(h5_dims)
-    dims = ntuple(i -> @inbounds(Int(h5_dims[N-i+1])), N)
-    maxdims = ntuple(i -> @inbounds(h5_maxdims[N-i+1]) % Int, N) # allows max_dims to be specified as -1 without triggering an overflow
-    obj isa Dataspace || close(dspace)
-    return dims, maxdims
-end
 
 function get_regular_hyperslab(dspace::Dataspace)
     start, stride, count, block = h5s_get_regular_hyperslab(dspace)
@@ -1130,14 +1115,6 @@ function get_regular_hyperslab(dspace::Dataspace)
     return rev(start), rev(stride), rev(count), rev(block)
 end
 
-
-"""
-    HDF5.set_extent!(dset::HDF5.Dataset, new_dims::Dims)
-
-Change the current dimensions of a dataset to `new_dims`, limited by
-`max_dims = get_simple_extent_dims(dset)[2]`. Reduction is possible and leads to loss of truncated data.
-"""
-set_extent!(dset::Dataset, new_dims::Dims) = h5d_set_extent(checkvalid(dset), hsize_t[reverse(new_dims)...])
 
 """
     start_swmr_write(h5::HDF5.File)
@@ -1746,6 +1723,9 @@ function Base.setindex!(chunk_storage::ChunkStorage, v::Tuple{<:Integer,Vector{U
 end
 
 # end of high-level interface
+
+
+include("api_midlevel.jl")
 
 
 ### HDF5 utilities ###
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1087,7 +1087,7 @@ end
 Base.isempty(dspace::Union{Dataspace,Dataset,Attribute}) = length(dspace) == 0
 
 """
-    isnull(dspace::Union{Dataspace, Dataset, Attribute})
+    isnull(dspace::Union{HDF5.Dataspace, HDF5.Dataset, HDF5.Attribute})
 
 Determines whether the given object has no size (consistent with the `H5S_NULL` dataspace).
 
@@ -1107,12 +1107,19 @@ function isnull(obj::Union{Dataspace,Dataset,Attribute})
     return ret
 end
 
-function get_dims(dspace::Dataspace)
+"""
+    HDF5.get_simple_extent_dims(obj::Union{HDF5.Dataspace, HDF5.Dataset,HDF5.Attribute}) -> dims, maxdims
+
+Get the array dimensions from a dataset or attribute and return a tuple of dims and maxdims.
+"""
+function get_simple_extent_dims(obj::Union{Dataspace,Dataset,Attribute})
+    dspace = obj isa Dataspace ? checkvalid(obj) : dataspace(obj)
     h5_dims, h5_maxdims = h5s_get_simple_extent_dims(dspace)
     # reverse dimensions since hdf5 uses C-style order
     N = length(h5_dims)
     dims = ntuple(i -> @inbounds(Int(h5_dims[N-i+1])), N)
     maxdims = ntuple(i -> @inbounds(h5_maxdims[N-i+1]) % Int, N) # allows max_dims to be specified as -1 without triggering an overflow
+    obj isa Dataspace || close(dspace)
     return dims, maxdims
 end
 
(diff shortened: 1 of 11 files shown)


src/deprecated.jl
```diff
@@ -11,3 +11,6 @@ function exists end
 ### Changed in PR#776
 @deprecate create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::Datatype, dspace::Dataspace,
     lcpl::Properties, dcpl::Properties, dapl::Properties, dxpl::Properties) HDF5.Dataset(HDF5.h5d_create(parent, path, dtype, dspace, lcpl, dcpl, dapl), HDF5.file(parent), dxpl) false
+
+@deprecate get_dims(dspace::Union{Dataspace,Dataset,Attribute}) get_simple_extent_dims(dspace) false
```
musm, line 15:
```suggestion
### Changed in PR#798
@deprecate get_dims(dspace::Union{Dataspace,Dataset,Attribute}) get_simple_extent_dims(dspace) false
```

src/HDF5.jl
```diff
@@ -1123,25 +1130,14 @@ function get_regular_hyperslab(dspace::Dataspace)
...
 """
-    set_dims!(dset::HDF5.Dataset, new_dims::Dims)
+    HDF5.set_extent!(dset::HDF5.Dataset, new_dims::Dims)
 
 Change the current dimensions of a dataset to `new_dims`, limited by
-`max_dims = get_dims(dset)[2]`. Reduction is possible and leads to loss of truncated data.
+`max_dims = get_simple_extent_dims(dset)[2]`. Reduction is possible and leads to loss of truncated data.
 """
-set_dims!(dset::Dataset, new_dims::Dims) = h5d_set_extent(checkvalid(dset), hsize_t[reverse(new_dims)...])
+set_extent!(dset::Dataset, new_dims::Dims) = h5d_set_extent(checkvalid(dset), hsize_t[reverse(new_dims)...])
```
musm, line 1140: Also remove the `!` since we don't really adhere to that notation throughout?

musm:
we might want to move these to api_helpers ?

jmert:
I agree with the general idea, but a few thoughts:
1. The asymmetry in `get_simple_extent_dims` versus `set_extent` seems off to me. I know the [`h5d_set_extent`](https://portal.hdfgroup.org/display/HDF5/H5D_SET_EXTENT) interface is what's being used, but there's also the [`h5s_set_extent_simple`](https://portal.hdfgroup.org/display/HDF5/H5S_SET_EXTENT_SIMPLE) which is closer (though still annoyingly swaps the order of "simple" and "extent"...). Maybe we should just provide both setters (first acting on dataspace and second acting on actual datasets)? As mid-level routines, we could choose to normalize naming a bit, too — provide `get_simple_extent_dims` and `set_simple_extent_dims` for the dataspace operations?
2. I'd say rather than moving to `api_helpers.jl` which right now is pretty focused on just low-level wrappings of the low-level api, a new `midlevel.jl` (name TBD) would be nice — I'd still like to see `src/HDF5.jl` broken up more to more clearly separate the high-level interface from the mid-level interface, so maybe this will just be the first set of functions to make the move. (?)

musm:
@jmert Updated the PR as per comment. Is that what you were more thinking of?

src/api_midlevel.jl
```diff
@@ -0,0 +1,47 @@
...
+Get the array dimensions from a dataspace, dataset, or attribute and return a tuple of `dims` and `maxdims`.
+"""
+function get_extent_dims(obj::Union{Dataspace,Dataset,Attribute})
+    dspace = obj isa Dataspace ? checkvalid(obj) : dataspace(obj)
+    h5_dims, h5_maxdims = h5s_get_simple_extent_dims(dspace)
+    # reverse dimensions since hdf5 uses C-style order
+    N = length(h5_dims)
+    dims = ntuple(i -> @inbounds(Int(h5_dims[N-i+1])), N)
+    maxdims = ntuple(i -> @inbounds(h5_maxdims[N-i+1]) % Int, N) # allows max_dims to be specified as -1 without triggering an overflow
+    obj isa Dataspace || close(dspace)
```
musm, line 45: I don't actually understand this line. From L39 `obj` will always be a `Dataspace`, so this check is always true ` close(dspace)` is never called.

jmert: `obj` can be a `Dataset` or `Attribute`, in which case the dataspace `dspace` is newly opened. It's only the case when `obj` is a literal `Dataspace` that we don't want to close it (since the caller "owns" it).

musm: yikes I thought this line was  `dspace isa Dataspace`....

src/api_midlevel.jl
```diff
@@ -0,0 +1,46 @@
...
+    checkvalid(dset)
+    h5d_set_extent(dset, hsize_t[reverse(size)...])
+end
+
+"""
+    HDF5.set_extent_dims(dset::HDF5.Dataspace, new_dims::Dims, max_dims::Union{Dims,Nothing} = nothing)
+
+Change the current dimensions of a dataspace `dset` to `new_dims`, limited by
+`max_dims = get_simple_extent_dims(dset)[2]`. Reduction is possible and leads to loss of truncated data.
+"""
```
jmert, line 21:
```suggestion
"""
    HDF5.set_extent_dims(dset::HDF5.Dataspace, new_dims::Dims, max_dims::Union{Dims,Nothing} = nothing)

Change the dimensions of a dataspace `dspace` to `new_dims`, optionally with the maximum possible
dimensions `max_dims` different from the active size `new_dims`. If not given, `max_dims` is set equal
to `new_dims`.
"""
```

Two sets of changes — 1) simple rename `dset` -> `dspace` in this doc string. 2) I think this doc string needs to explain `max_dims` (especially what the default `nothing` means), so I've suggested a sample rewrite (but feel free to reword yet again).

src/api_midlevel.jl
```diff
@@ -0,0 +1,46 @@
...
+function set_extent_dims(dspace::Dataspace, size::Dims, max_dims::Union{Dims,Nothing} = nothing)
+    checkvalid(dspace)
+    rank = length(size)
+    current_size = hsize_t[reverse(size)...]
+    maximum_size = isnothing(max_dims) ? C_NULL : hsize_t[reverse(max_dims)...]
```
jmert, line 26:
```suggestion
    maximum_size = isnothing(max_dims) ? C_NULL : [reverse(max_dims .% hsize_t)...]
```
to support using `-1` to set unlimited max size.

(Unfortunately, using `max_dims::Dims` makes this incompatible with unlimited sizes given as `(HDF5.H5S_UNLIMITED,)` ... not sure what to do about that, but it's not a regression from what `set_dims!` was doing already.)

test/extend_test.jl
```diff
@@ -1,23 +1,40 @@
...
+
+dspace_maxd = dataspace((100, 4), max_dims = (256, 5))
+@test HDF5.get_extent_dims(dspace_maxd)[1] == (100, 4)
+@test HDF5.get_extent_dims(dspace_maxd)[2] == (256, 5)
+HDF5.set_extent_dims(dspace_maxd, (8, 2))
+@test HDF5.get_extent_dims(dspace_maxd)[1] == (8, 2)
+HDF5.set_extent_dims(dspace_maxd, (3, 1), (4,2))
+@test HDF5.get_extent_dims(dspace_maxd)[1] == (3, 1)
+@test HDF5.get_extent_dims(dspace_maxd)[2] == (4, 2)
+
```
jmert, line 21:
```suggestion
HDF5.set_extent_dims(dspace_maxd, (3, 1), (-1, -1)) # unlimited max size
@test HDF5.get_extent_dims(dspace_maxd)[1] == (3, 1)
@test HDF5.get_extent_dims(dspace_masd)[2] == (-1, -1)
```

test/extend_test.jl
```diff
@@ -1,23 +1,40 @@
 using HDF5
 using Test
 
-@testset "extendtest" begin
+@testset "extend" begin
+
+
+## mid-level routines
```
jmert, line 7: These tests I think would make more sense in `test/dataspace.jl`. The `get_extent_dims` method is already tested pretty exhaustively there, so I think it just needs a new block of tests to run and check `set_extent_dims`.

jmert (changes requested):

jmert:
> @jmert Updated the PR as per comment. Is that what you were more thinking of?

Yeah, this is the direction I was thinking of. Just some minor comments, but I agree with the approach as a whole.

src/api_midlevel.jl
musm: Yeah I think this will take some more careful thinking on how to address  setting `H5S_UNLIMITED` more broadly.

musm: I think the simplest thing is to just document the correct way to set an unlimited dimension is to set it to `-1`  and not use the internal HDF5.H5S_UNLIMITED constant?  A little more instrusive would be to set `HDF5.H5S_UNLIMITED=-1`, but I think that would require enforcing truncation to which ever inter api call excepts `HDF5.H5S_UNLIMITED`  since `cconvert` to `hsize_t` would throw an error.

jmert: I agree that it's mostly a documentation issue — low-level API will want `HDF5.H5S_UNLIMITED` (and should remain as `-1 % hsize_t` to agree with the C header literal value), and everything in mid-level and high-level API will instead use `-1` and do the data type conversions. I think so far we just don't have the documentation to make that clear.

PR #799, opened 2021-01-08
Update CI.yml

musm:


musm:
I guess their migration starting Nov 30 is stalled (ref https://github.com/actions/virtual-environments/issues/1816)

musm:
Closing since the roll out will take longer than anticipated.

PR #804, opened 2021-01-14
Document gotchas when interoperation with python

jw3126:
#785


docs/src/h5py.md
```diff
@@ -0,0 +1,105 @@
...
+end
+```
+Similarly `reversedims` can be used before saving arrays intended for use from python.
+If copying of data is undesirable, other options are:
+* using Fortran memory layout on the python side
+* using C-memory layout on the Julia side (e.g. a lazy variant of `reversedims`)
+
+The whole example as a Julia executable script:
+```julia
+using PyCall
```
jw3126, line 61: We could add `PyCall` as a dependency and turn this into a doctest. However it might be fragile since we need to install h5py.

musm: Is possible to use `Conda.jl` to install the hdf5  library in order to turn these to doctests. I'm not opposed to adding a PyCall test dependency.  In general, it does seem to be a good idea to ensure inter-op.

docs/src/h5py.md
```diff
@@ -0,0 +1,105 @@
...
+    @assert arr2d == [1 2 3; 4 5 6]
+    @assert arr3d == reshape(arr2d, (1,2,3))
+end
+```
+Similarly `reversedims` can be used before saving arrays intended for use from python.
+If copying of data is undesirable, other options are:
+* using Fortran memory layout on the python side
+* using C-memory layout on the Julia side (e.g. replace `permutedims` by `PermutedDimsArray` above)
+
+The whole example as a Julia executable script:
```
musm, line 59: do we really need to repeat this?

docs/src/h5py.md
```diff
@@ -0,0 +1,105 @@
...
+    @assert size(arr3d) == (3,2,1)
+end
+```
+To fix this, we can simply reverse the dimensions again:
+
+```julia
+using HDF5
+function reversedims(arr)
+    dims = ntuple(identity, Val(ndims(arr)))
+    return permutedims(arr, reverse(dims))
```
musm, line 41:
```julia
    return permutedims(arr, reverse(1:ndims(arr)))
```

docs/src/h5py.md
jw3126: I also thought about this and don't like this repetition. The alternatives seem worse to me though.
The issue is, that above we use \```python``` code, that e.g. `Literate` can not handle. We would need to wrap it using PyCall which is extra boiler plate and confusing.

docs/src/h5py.md
jw3126: We could, but this introduces further headaches. If you run tests locally using e.g. system python things may fail.

docs/src/h5py.md
jw3126: I think I found a way to fix this.

docs/src/h5py.md
jw3126: I added this anyway.

jw3126:
@musm https://github.com/JuliaIO/HDF5.jl/pull/804/checks?check_run_id=1708414939#step:4:115 any ideas?

jw3126:
BTW is it intended, that CI passes, despite the doctest fail?

musm:
> BTW is it intended, that CI passes, despite the doctest fail?

It shouldn't, not sure why it passed yet.

musm:
I'm trying to decide if we should add `Literate` as a test dependency.  What was the specific reason before the last commit that spawned you to switch over to use Literate?

jw3126:
> I'm trying to decide if we should add `Literate` as a test dependency. What was the specific reason before the last commit that spawned you to switch over to use Literate?

The only reason is, that I find it easier to develop a `.jl` document, which is executable Julia code, instead of a `.md` document that I can only run doctests on. I removed the Literate dependency.

jw3126:
@musm https://github.com/JuliaIO/HDF5.jl/runs/1762155054?check_suite_focus=true#step:4:112 ideas? I cannot reproduce this error locally.

PR #810, opened 2021-01-19
Move silence errors to midlevel and remove straggling definition of rehash

musm:


src/api_midlevel.jl
```diff
@@ -45,3 +45,19 @@ function get_extent_dims(obj::Union{Dataspace,Dataset,Attribute})
     obj isa Dataspace || close(dspace)
     return dims, maxdims
 end
+
+"""
+    silence_errors(f::Function)
+
+Silence errors when calling the function `f` from internal hdf5 library methods that should not throw an error message.
```
jmert, line 52:
```suggestion
During execution of the function `f`, disable printing of internal HDF5 library error messages.
```
? The way it's currently written, I kind of read that as when `f` is called by libhdf5 methods (as in a callback).

jmert (approved):

PR #823, opened 2021-03-20
Add module scoping to unexported functions in docs

kimlaberinto:
I noticed that some code examples in the docs didn't have module scoping where it needed it.

Referencing the HISTORY.md and #699 , I looked for the un-exported functions in the docs (using the following command snippet).
```
cd docs/
grep -RIn -Ei "(file\(|filename\(|name\(|get_chunk\(|get_datasets\(|iscontiguous\(|ishdf5\(|ismmappable\(|root\(|readmmap\(|set_dims\!\(|get_access_properties\(|get_create_properties\(|create_external_dataset\()" . -C 3
```
(the grep line is to look for each function in the list with a parenthesis e.g `ismmappable(` in the docs folder and display them to console with 3 lines of context.)

I then manually added `HDF5` scoping to them I thought appropriate.

I haven't checked if the code works for every single one but I checked a few simple ones like `HDF5.immappable` and it works. I wanted to touch base with you all (the maintainers) to see if I should double check all the code snippets, or if you think this should be good.

Let me know if there might be more unexported functions that I missed in the docs or if there's anything I should change. Otherwise I think it should be ready to merge/review.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -238,8 +238,8 @@ You can use the `ismmappable` function to test whether this is possible; for exa
 
 ```julia
 dset = g["x"]
-if ismmappable(dset)
-    dset = readmmap(dset)
+if HDF5.ismmappable(dset)
+    dset = HDF5.readmmap(dset)
 end
 val = dset[15]
 ```
@@ -257,8 +257,8 @@ The following fails:
 
 ```julia
 vec_dset = create_dataset(g, "v", datatype(Float64), dataspace(10_000,1))
-ismmappable(vec_dset)    # == true
-vec = readmmap(vec_dset) # throws ErrorException("Error mmapping array")
+HDF5.ismmappable(vec_dset)    # == true
+vec = HDF5.readmmap(vec_dset) # throws ErrorException("Error mmapping array")
 ```
 
 because although the dataset description has been added, the space within the HDF5 file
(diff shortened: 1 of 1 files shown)


kimlaberinto:
Now I also checked the functions referenced in `src/HDF5.jl` but nothing came up. So those functions should also be good to go.
```
grep -RIn -EI "(iscompact\(|ischunked\(|refresh\(|start_swmr_write\(|create_external\()" docs/ -C 3 
```

So all of these functions listed in `src/HDF5.jl` I now checked in the docs and should be good. (copied from `src/HDF5.jl`)
```
### The following require module scoping ###

# file, filename, name,
# get_chunk, get_datasets,
# get_access_properties, get_create_properties,
# root, readmmap, set_dims!,
# iscontiguous, iscompact, ischunked,
# ishdf5, ismmappable,
# refresh
# start_swmr_write
# create_external, create_external_dataset
```

jmert (approved):
Thank you for the contribution! Docs are something I (we) haven't gotten around to improving as much as I'd hoped, so thank you.

Just a quick review looks good to me, and it's not an error to "unnecessarily" scope a call, so I think this is a safe change. (Not saying this is unnecessary, just that not having explicitly, checked it's OK to be wrong.)

kimlaberinto:
Thank you! And you're very welcome! Thank you for the Julia package

Feel free to merge!

jmert:
@musm It appears the docs preview deploy is errorring but CI didn't fail. Not a barrier to merging this PR, but something we should probably look into.

musm:
I don't think doc deploys work on forks

musm:
Whops sorry @jmert  that wasn't what you are refering to. Indeed, let me investigate..

PR #833, opened 2021-05-23
Split out src/HDF5.jl into multiple files

simonbyrne:
The file was getting huge (>2000 lines), and was difficult to find anything. I've tried to organise it along the lines of how the libhdf5 library itself is arranged. No functionality has changed, though I did need to split out a few Union methods into individual ones depending on where the type was defined.


simonbyrne:
I will admit to having some ambitions to refactor the code: I'm not sure how much time I can commit, but this seemed like a necessary first step.

musm:
Many thanks for the PR and effort here.  
Honestly, I think some headers and separation in the main source file would be easier/better than splitting this for the sake of splitting. In general, I'm not a huge fan of having many files with short definitions, since I often find it easier to scroll through a single file to identify and find patterns, as long as that single file is well structured.  For example, the separate file for chunked doesn't seem as effective as opposed to a header indicating this is for the chunked interface in the original source file.  Perhaps I could be convinced if there is a greater plan in play or after letting this sink for some time.  

Going to CC @jmert on this as well since it might impact some of his plans.

simonbyrne:
> For example, the separate file for chunked doesn't seem as effective as opposed to a header indicating this is for the chunked interface in the original source file. 

Chunked is something I wasn't sure about, as it doesn't map nicely to the underlying HDF5 api. I have made a start on a further refactor, will open a PR in a day or two.

jmert:
Sorry about the delay in responding.

I think my opinion falls somewhere between musm and symonbyrne's — I think breaking down `HDF5.jl` a bit more is warranted, but I don't see that quite this many individual files are necessary. The way I'd originally imagined the reorganization was to continue toward more clearly separating the high-level and mid-level APIs into separate source files (with maybe the exception of the type-conversion handling which is complicated enough on its own that a separate file dedicated to just that is probably OK too) — i.e. all of the `File`, `Properties`, etc types would be defined in one source file along with the `getindex`/`setindex!`/`read`/`write`/etc generic function definitions, and then the "middle level" methods like `read_dataset`/`create_group`/etc would share a file.

simonbyrne:
I'll close this for now since it has diverged too much from master. I've made a smaller change in #851

PR #834, opened 2021-05-26
Refactor properties

simonbyrne:
This builds on #833 to refactor the Properties interface. The main changes are:
- the `Properties` object is now parametrised on a `PropertyClass`: this is used to emulate the inheritance of HDF5 property lists (e.g. so the properties in `ObjectCreateProperties` are inherited by `GroupCreateProperties`)
  - this allows dispatch to be more specific on which properties object is used.
  - it also makes it easier to add new properties (there are still quite a few that are not exposed)
- I switched property access from `getindex`/`setindex!` to `getproperty`/`setproperty!`: I felt that this was more consistent with using keywords as constructors (rather than `=>`). It also allows tab-completion.
- Passing keyword options to multiple property lists will result in an error if a keyword is not found in any of the lists. This found at least one error: in `create_external_dataset`, `dcpl` was being passed as a keyword parameter, and was silently dropped as there was no key of that name:
https://github.com/JuliaIO/HDF5.jl/commit/ee1b512652ceb7c893af239a5f20183dd9646228#diff-8f07b43441469a1a53b2bfd828733e207d9fc11c0bccf786d08a1cfe7e1c7843L6-R7

Any thoughts/feedback? One open question is whether to rename `Properties` to `PropertyList` to match HDF5?


simonbyrne:
I've added a few more changes:
- all low-level api functions are now in a separate directory `src/api/` and submodule (`API)`.
- option setters/getters now take and return symbol names (e.g. `alloc_time=:early` instead of `alloc_time=HDF5.H5D_ALLOC_TIME_EARLY`, or Booleans where appropriate (`create_intermediate_group=true`).
- created a `FilterPipeline` object which wraps a `Properties` object, and acts like a vector. Filters themselves are objects and can be `push!`ed onto the pipeline.
- drivers are also objects: currently only two are exposed `POSIX` (the default) and `MPIO` (for MPI).

musm:
Sorry for the late response, I'm just getting to this. This is looking tricky to review since it combines https://github.com/JuliaIO/HDF5.jl/pull/833 

Should I first be looking at that previous PR? This PR also looks to have further split things out and reorganized the API. I think it's going to be challenging to review this all in a single PR.  Some of these changes look much welcome.

simonbyrne:
Yeah, I'm not sure the best way forward. #833 doesn't actually change any code, so is probably a good place to start. There have been a few PRs merged since this was forked, but none of them look too difficult to incorporate. If you're in favor of this PR, I can merge in the changes (some such as #840 actually overlap, but mostly they shouldn't be too hard to combine).

I have another change which is almost done (creating a `HDF5Error` that hooks into the HDF5 error handling machinery), was planning on adding it to this, but I can wait if you want to review this first.

simonbyrne:
I think I can break off the `api` subdirectory stuff into a separate PR if that would help.

musm:
Sounds good,  I'll start with #833. The main thing there is that both jmert and I think that overly separating out into tiny files doesn't bring much benefit and that we prefer file separation based on the API 'level'. I can check out that PR and create a new PR with changes a top that PR with proposed modifications. The API subdirectory changes would be better in a seperate PR to make the review easier. I think after #833, we can look at the API submodule change, then this PR and proposed future changes. I think it'll be easier that way since the deprecation process will be easier.

simonbyrne:
Certainly. I'm also happy to rearrange it if you can suggest what can be combined. My only request is that we keep the type conversion logic somewhat separate, as that is by far the most complicated code.

PR #839, opened 2021-06-13
Support Base.getindex(::Attribute, I...)

### LINKED ISSUE #838: Base.getindex not implemented for Attribute type
The only way I can see to get the value of an `Attribute` is through `read(attr)`. I'd like to be able to use the Dataset-like way of accessing the value of an `Attribute` via the `Base.getindex` function (i.e. by using `attr[]` instead of `read(attr)`).

I see that `src/HDF5.jl` defines `Base.getindex(dset::Dataset, I...)`, but the neighboring functions operate on a `DatasetOrAttribute` Union type.  I can't find any comments in the code or in commit log messages that explain why this function is limited to just the `Dataset` type rather than the broader `DatasetOrAttribute` type.

The `generic_read` function used by `getindex` already operates on `DatasetOrAttribute`, which suggest that it's just an oversight, but is there some reason for restricting `getindex` to just `Dataset` types?

david-macmahon:
Widen the type of the first parameter from...

    Base.getindex(dset::Dataset, I...)

...to...

    Base.getindex(obj::DatasetOrAttribute, I...)

`DatasetOrAttribute` is already defined and supported elsewhere.  Most
importantly, `generic_read` already accepts that Union type.

The allows one to access attribute values using syntax like:

    h5file["dataset"]["attribute"][]

rather than having to call `read` directly like:

    read(h5file["dataset"]["attribute"])

Fixes #838

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1182,10 +1182,10 @@ function Base.read(obj::DatasetOrAttribute)
     return val
 end
 
-function Base.getindex(dset::Dataset, I...)
-    dtype = datatype(dset)
+function Base.getindex(obj::DatasetOrAttribute, I...)
+    dtype = datatype(obj)
     T = get_jl_type(dtype)
-    val = generic_read(dset, dtype, T, I...)
+    val = generic_read(obj, dtype, T, I...)
     close(dtype)
     return val
 end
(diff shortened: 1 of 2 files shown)


musm:
LGTM, any chance you could add  a test to prevent regressions?

musm (approved):

musm:
will merge in 24 hours sans objections

PR #840, opened 2021-06-13
Add `filter` keyword argument to create_dataset

david-macmahon:
This allows `create_dataset` to create a dataset with a user-defined
filter by setting the `filter` keyword to a Tuple of integers that
describe and configure the filter.  For example, to use a filter
specified by `filter_id=32013`, `flags=0`, `cd_nelmts=3`, and
`cd_values=(2,0,8)`:

    ds, dt = create_dataset(h5, "data", data,
                            chunk=(256,1,256),
                            filter=(32013,0,2,0,8))

Note that `cd_nelmts` is not included explicitly.  It is inferred from
the length of the Tuple.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1927,6 +1927,8 @@ end
 
 # Set a single filter
 function set_filter(p::Properties, filter_id, flags, cd_values...)
+    # Passing cd_values as Cuint[] allocates less than passing as
+    # Ref{NTuple{N,Cuint}} (and it is compatible with Julia 1.3)
     h5p_set_filter(p::Properties, filter_id, flags, length(cd_values), Cuint[cd_values...])
 end
 
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1927,13 +1927,13 @@ end
 
 # Set a single filter
 function set_filter(p::Properties, filter_id, flags, cd_values...)
-    h5p_set_filter(p::Properties, filter_id, flags, length(cd_values), Ref(UInt32.(cd_values)))
+    h5p_set_filter(p::Properties, filter_id, flags, length(cd_values), Ref(Cuint.(cd_values)))
 end
 
 # Set multiple filters
-function set_filter(p::Properties, filter::Tuple, more_filters...)
+function set_filter(p::Properties, filter::Tuple, additional_filters...)
     set_filter(p::Properties, filter...)
-    for f in more_filters
+    for f in additional_filters
       set_filter(p::Properties, f...)
     end
 end
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -840,6 +840,7 @@ function _prop_set!(p::Properties, name::Symbol, val, check::Bool = true)
                name === :compress    ? h5p_set_deflate(p, val...) :
                name === :deflate     ? h5p_set_deflate(p, val...) :
                name === :external    ? h5p_set_external(p, val...) :
+               name === :filter      ? set_filter(p, val...) :
                name === :layout      ? h5p_set_layout(p, val...) :
                name === :shuffle     ? h5p_set_shuffle(p, val...) :
                name === :track_times ? h5p_set_obj_track_times(p, val...) : # H5P_OBJECT_CREATE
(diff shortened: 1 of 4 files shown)


src/HDF5.jl
```diff
@@ -1924,6 +1925,19 @@ function get_chunk(dset::Dataset)
     ret
 end
 
+# Set a single filter
+function set_filter(p::Properties, filter_id, flags, cd_values...)
+    h5p_set_filter(p::Properties, filter_id, flags, length(cd_values), Ref(UInt32.(cd_values)))
```
musm, line 1930:
```suggestion
    h5p_set_filter(p::Properties, filter_id, flags, length(cd_values), Ref(Cuint.(cd_values)))
```

src/HDF5.jl
```diff
@@ -1924,6 +1925,19 @@ function get_chunk(dset::Dataset)
...
+end
+
+# Set multiple filters
+function set_filter(p::Properties, filter::Tuple, more_filters...)
+    set_filter(p::Properties, filter...)
+    for f in more_filters
```
musm, line 1936:
```suggestion
function set_filter(p::Properties, filter::Tuple, additional_filters...)
    set_filter(p::Properties, filter...)
    for f in additional_filters
```

musm (approved):

musm:
Looks like there's some issue with CI (might need to rebase on top of master), but otherwise LGTM.

musm:
will merge in 24 hours sans objections

david-macmahon:
I just pushed the suggested changes (though not via the GitHub interface).  As the commit log message says, better names are better.  Thanks!

musm:
Any clue on the https://github.com/JuliaIO/HDF5.jl/pull/840/checks?check_run_id=2826317219 failure, it might be  real?

musm:
Looks like it's a convert error although I'm not sure why it's isolated to julia 1.3
```
  MethodError: no method matching unsafe_convert(::Type{Ptr{UInt32}}, ::Base.RefValue{Tuple{UInt32}})
```

src/HDF5.jl
```diff
@@ -1924,6 +1925,19 @@ function get_chunk(dset::Dataset)
     ret
 end
 
+# Set a single filter
+function set_filter(p::Properties, filter_id, flags, cd_values...)
+    h5p_set_filter(p::Properties, filter_id, flags, length(cd_values), Ref(Cuint.(cd_values)))
```
musm, line 1930:
```suggestion
    h5p_set_filter(p::Properties, filter_id, flags, length(cd_values), Cuint[cd_values...])
```

musm: Looks like for some reason 1.3.1 has issues with the `unsafe_convert` of the Tuple to Ptr{UInt32}

musm: I'm guessing this was added in https://github.com/JuliaLang/julia/pull/34199

david-macmahon: Yeah, looks like that was officially added in 1.5: https://github.com/JuliaLang/julia/blob/release-1.5/NEWS.md

What to do?  It would be easy enough to throw an informative exception proactively if someone tries to use this functionality on 1.3.

david-macmahon: Oh, just seeing your suggestion now.  That will allocate, right?  I wonder if `Ref{Cuint}(cd_values...)` also ends up allocating.

musm: The suggested change should work in a backwards compatible way

musm: Right, allocation would be a perf issue with the proposed change.

david-macmahon: Yes, it turns out `Cuint.(cd_values)` allocates more bytes that `Cuint[c_values...]`!  Change pushed.

src/HDF5.jl
```diff
@@ -1924,6 +1925,19 @@ function get_chunk(dset::Dataset)
     ret
 end
 
+# Set a single filter
+function set_filter(p::Properties, filter_id, flags, cd_values...)
+    h5p_set_filter(p::Properties, filter_id, flags, length(cd_values), Cuint[cd_values...])
```
musm, line 1930:
```suggestion
    # can pass the tuple cd_values directly starting from Julia 1.5+. We use an array for compatibility with 1.3 
    h5p_set_filter(p::Properties, filter_id, flags, length(cd_values), Cuint[cd_values...])
```

musm: I think  a minor comment would be good to add so that when we drop 1.3 support, we could update this

david-macmahon: FWIW, passing the tuple directly (without wrapping in a `Ref`) leads to this on 1.6.1:
```
MethodError: no method matching unsafe_convert(::Type{Ptr{UInt32}}, ::Tuple{UInt32})
```

david-macmahon:
Thanks for your great detective work @musm!

PR #845, opened 2021-06-22
Move C API to submodule

simonbyrne:
This moves all the C API functions and types to an `API` submodule (this was part of #834, but changed to not depend on #833).

The tests aren't yet passing as `h5d_read_chunk` in `src/api/helpers.jl` depends on `get_chunk_length` and `get_chunk_index` (defined in `src/api_midlevel.jl`). It doesn't really make sense to have the C API depend on higher-level API functions, so one of these should probably be moved: which would make more sense?


simonbyrne:
>  `h5d_read_chunk` in `src/api/helpers.jl` depends on `get_chunk_length` and `get_chunk_index` (defined in `src/api_midlevel.jl`). It doesn't really make sense to have the C API depend on higher-level API functions, so one of these should probably be moved: which would make more sense?

Given that these functions perform allocations and use keyword arguments, I decided it made more sense to move them to the midlevel (and deprecated some function signatures which were a bit confusing and unnecessary).

simonbyrne:
I think this is ready. @musm any thoughts?

src/deprecated.jl
```diff
@@ -19,21 +19,6 @@ function exists end
 ### Changed in PR #844
 @deprecate silence_errors(f::Function) f()
 
-### Changed in PR #845
```
musm: I decided to delete this since it was never on  a tagged release and we just merged these changes in the last week. So no one on a tagged version would be effected by this change.

simonbyrne:
One question is whether we should move `src/error.jl` into `src/api/`? That's the only place where the lower-level API accesses the higher one.

musm:
It's a good question. I was thinking the same, for the reason you mention.  The question is whether we should use `H5Error` for all exceptions in the HDF5 and API module,  if that's the case then we have to keep it as is. If we want to restrict the error to the inner API, , perhaps we should  create an `H5LibError <: H5AbstractError` as a subtype?

musm:
Hmm, on second thought, I'm actually thinking it would be better to leave the custom exception handling in the API module, since it is strictly being used to handle internal hdf5 exceptions and wouldn't make sense to use outside.

musm:
Let me know what you think about the last commit. 

Given that the error is isolated in its own `API` module., we could  even reuse the same `H5Error` name in the upper level module `HDF5` without issues, as a higher level `Exception` to deal with errors in that module, whereas `HDF5.API.H5Error` makes it clear it handles the internal `API` level errors.

src/api/error.jl
```diff
@@ -35,9 +53,9 @@ const SHORT_ERROR = Ref(true)
 
 function Base.showerror(io::IO, err::H5Error)
     n_total = length(err)
-    print(io, "H5Error: ", err.msg)
+    print(io, "$(typeof(err)): ", err.msg)
```
musm, line 56: I used `typeof` here so that the full module path is printed:
image


I think this is a good idea in acse we decide to use the same exception error name in the main module.

musm:
I think this PR is good to go, unless you think we should handle the internal errors in a different manner.

simonbyrne:
LGTM

PR #846, opened 2021-06-24
Fix docs due to module changes

musm:
Due to PR https://github.com/JuliaIO/HDF5.jl/pull/845 we have some issues with the docs that need resolving. I'm not sure it is optimally setup, but at least the docs are building without errors.

### DIFF
--- a/docs/src/api_bindings.md
+++ b/docs/src/api_bindings.md
@@ -3,7 +3,7 @@
 gen/gen_wrappers.jl script -->
 ```
 ```@meta
-CurrentModule = HDF5
+CurrentModule = HDF5.API
 ```
 
 # Low-level library bindings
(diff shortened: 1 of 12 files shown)


src/mpio.jl
```diff
@@ -1,6 +1,14 @@
 using .MPI
 import Libdl
 
+const depsfile = joinpath(@__DIR__, "..", "deps", "deps.jl")
```
musm, line 4: The check here requires we check `libhdf5`, but since we moved that to `API` submodule I had to re-add this here.

musm:
cc @simonbyrne

src/mpio.jl
musm: The tests should've shown this as an error, so this requires some debugging.

docs/src/api_bindings.md
```diff
@@ -23,553 +23,553 @@ documented here.
...
 ## [`H5`](https://portal.hdfgroup.org/display/HDF5/Library) — General Library Functions
-- [`h5_close`](@ref HDF5.h5_close)
-- [`h5_dont_atexit`](@ref HDF5.h5_dont_atexit)
-- [`h5_free_memory`](@ref HDF5.h5_free_memory)
-- [`h5_garbage_collect`](@ref HDF5.h5_garbage_collect)
-- [`h5_get_libversion`](@ref HDF5.h5_get_libversion)
-- [`h5_is_library_threadsafe`](@ref HDF5.h5_is_library_threadsafe)
-- [`h5_open`](@ref HDF5.h5_open)
-- [`h5_set_free_list_limits`](@ref HDF5.h5_set_free_list_limits)
+- [`API.h5_close`](@ref API.h5_close)
```
simonbyrne, line 26: You should be able to just set `CurrentModule = HDF5.API`

musm: Apparently, you had to also had to remove the exiting `HDF5` prefix to get  `CurrentModule = HDF5.API` to work.

musm:
I'm getting an odd method h5p_set_fapl_mpio matching error. If it's not immediately obvious I'll try to test later with a linux machine.

simonbyrne:
Ah, this line:
https://github.com/JuliaIO/HDF5.jl/blob/9e45452df63bc975d43b8fbd2a1a9c016fdbbae9/src/HDF5.jl#L804
should be
```julia
    name === :fapl_mpio     ? set_fapl_mpio(p, val...) : 
```

musm:
Great, that fixes it. I'm not too keen on how the `has_parallel` methods operate right now, but those are best left to a seperate PR. In particular, the loading of the `deps.jl` file in the `mpio.jl` file is pretty ugly.  It's loaded there and we also load it in the `API` submodule, and the `__init__()` checks the `API.check_deps()` method.  I reckon doing it all in the main `HDF5.jl` file makes the most sense, and handling the check there, then simply loading the deps file again in the `API` submodule, without checks.

simonbyrne:
>  In particular, the loading of the `deps.jl` file in the `mpio.jl` file is pretty ugly. It's loaded there and we also load it in the `API` submodule, and the `__init__()` checks the `API.check_deps()` method. I reckon doing it all in the main `HDF5.jl` file makes the most sense, and handling the check there, then simply loading the deps file again in the `API` submodule, without checks.

I think you only need it to get the library: in that case, you can just use `API.libhdf5`: I've made this change as part of #847.

PR #847, opened 2021-06-25
Refactor Properties interface

simonbyrne:
This supersedes #834: 
- it defines separate`Properties` types for each HDF5 property list class
  - this allows dispatch to be more specific on which properties object is used.
  - it also makes it easier to add new properties (there are still quite a few that are not exposed)
- Switches property access from `getindex`/`setindex!` to `getproperty`/`setproperty!`: I felt that this was more consistent with using keywords as constructors (rather than `=>`). It also allows tab-completion.
- Option setters/getters now take and return symbol names (e.g. `alloc_time=:early` instead of `alloc_time=HDF5.H5D_ALLOC_TIME_EARLY`, or Booleans where appropriate (`create_intermediate_group=true`).
- compression/validation filters are now handled via a `FilterPipeline` object which wraps a `Properties` object, and acts like a vector. Filters themselves are objects and can be `push!`ed onto the pipeline.
- Drivers are also objects: currently only two are exposed `POSIX` (the default) and `MPIO` (for MPI).
- Passing keyword options to multiple property lists will result in an error if a keyword is not found in any of the lists. This found at least one error: in `create_external_dataset`, `dcpl` was being passed as a keyword parameter, and was silently dropped as there was no key of that name.


simonbyrne:
@musm I think this is ready, let me know what you think.

src/api/functions.jl
```diff
@@ -2440,3 +2517,80 @@ function h5z_register(filter_class)
     return nothing
 end
 
+"""
+    h5fd_core_init() -> hid_t
+
+See `libhdf5` documentation for [`H5FD_core_init`](https://portal.hdfgroup.org/display/HDF5/H5FD_CORE_INIT).
```
musm, line 2523: Doesn't look like any of these hd5f_* methods are defined in the docs.

src/HDF5.jl
```diff
@@ -19,7 +19,8 @@ copy_object, open_object, delete_object,
 create_datatype, commit_datatype, open_datatype,
 create_property,
 group_info, object_info,
-dataspace, datatype
+dataspace, datatype,
+Filters, Drivers
```
musm, line 23: Did you really intend to export these types? Currently the types in HDF5 have relied on module scoping. Probably just add these to L39 https://github.com/JuliaIO/HDF5.jl/pull/847/files#diff-3df1923ba1a5ee433e521ee7d3f75d066f7674b297db94ecda3ad89f7fc3e9a2R39

musm:
> @musm I think this is ready, let me know what you think.

Looks great! Much improved API. Bear with me as I slowly review this PR, hopefully will be done before the weekend is over.

src/HDF5.jl
simonbyrne: I thought it would be useful to export these so that users could do, e.g.
```
f(..., filters=[Filters.Shuffle(), Filters.Deflate(9)])
```

src/api/functions.jl
simonbyrne: They're not: they're API is exposed via macros, e.g.:
https://github.com/HDFGroup/hdf5/blob/1d680fe04c545d601678d876ad22dea7bfc31edd/src/H5FDcore.h#L23
the only place I see them in the docs is https://portal.hdfgroup.org/display/HDF5/H5P_GET_DRIVER

simonbyrne:
No rush.

simonbyrne:
Bump?

musm:
Sorry things cropped up. Will finish the review soon.

simonbyrne:
any more thoughts?

simonbyrne:
Bump?

musm:
HI @simonbyrne, I have to apologize for letting this slip by. Last few months have been busy on my personal side and unfortunately hasn't left me with much time to get back to Julia work. My goal for this month is to finish this review and your other PRs, which I hope will happen sooner this month rather than later. I'll make it a priority. Thanks again for you patience.

simonbyrne:
@musm any more comments?

src/properties.jl
```diff
@@ -0,0 +1,668 @@
...
+function class_setproperty!(::Type{GroupCreateProperties}, p::Properties, name::Symbol, val)
+    name === :local_heap_size_hint ? API.h5p_set_local_heap_size_hint(p, val) :
+    class_setproperty!(superclass(GroupCreateProperties), p, name, val)
+end
+
+"""
+    FileCreateProperties(;kws...)
+
+Properties used when creating a new `Group`. Inherits from
+[`GroupCreateProperties`](@ref),  with additional properties:
```
musm, line 285:
```suggestion
Properties used when creating a new `File`. Inherits from
[`ObjectCreateProperties`](@ref),  with additional properties:
```

musm:
Almost done, just going through the last file.

src/properties.jl
```diff
@@ -0,0 +1,668 @@
...
+            enum = $api_get_property(p)
+            return $get_expr
+        end
+        function $(esc(set_property!))(p::Properties, val)
+            property = $(QuoteNode(property))
+            enum = $set_expr
+            return $api_set_property(p, enum)
+        end
+        function $(esc(set_property!))(p::Properties, enum::Integer)
+            # deprecate?
```
musm, line 194: I agree we should deprecate this, but it seems tricky to do this since it requires the inverse dict mapping and calling the `@deprecate` macro within the quoted block of this return.

simonbyrne: Yeah, I couldn't see an easy way to do it.

src/properties.jl
```diff
@@ -0,0 +1,668 @@
...
+
+@enum_property(layout,
+    :compact    => API.H5D_COMPACT,
+    :contiguous => API.H5D_CONTIGUOUS,
+    :chunked    => API.H5D_CHUNKED,
+    :virtual    => API.H5D_VIRTUAL)
+
+
+# filters getters/setters
+get_filters(p::Properties) = Filters.FilterPipeline(p)
```
musm, line 401: These filters getters and setters should be defined Properties not DatasetCreateProperties ?

simonbyrne: We could be more specific in some cases, but since we can't represent the inheritance with subtypes, it isn't possible to define them all this way, e.g. getters/setters defined on `ObjectCreateProperties` wouldn't work on `FileCreateProperties`.

In the end, since the getters and setters are really an internal API, I figured it was more effort than it was worth.

musm: Oh right all the getters/setters are defined on the abstract type. If they weren't then it would mess up in the inheritance chain and not find them. Ok I think understand this then.

musm: Lastly, wouldn't calling `get_filters` always reset the FilterPipeline?

musm: By reset I mean creat a new `FilterPipeline` object thus dropping the existing pipeline?

musm: Oh nvm I now understand what this is doing.

src/HDF5.jl
```diff
@@ -704,17 +704,19 @@ function split1(path::AbstractString)
...
 function create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::Datatype, dspace::Dataspace; pv...)
-    dcpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(API.H5P_DATASET_CREATE; pv...)
-    dxpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(API.H5P_DATASET_XFER; pv...)
-    dapl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(API.H5P_DATASET_ACCESS; pv...)
+    dcpl = DatasetCreateProperties()
+    dxpl = DatasetTransferProperties()
+    dapl = DatasetAccessProperties()
+    pv = setproperties!(dcpl,dxpl,dapl; pv...)
+    isempty(pv) || error("invalid keyword options")
     haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
```
musm, line 720:
```suggestion
    haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
    dcpl = DatasetCreateProperties()
    dxpl = DatasetTransferProperties()
    dapl = DatasetAccessProperties()
    pv = setproperties!(dcpl,dxpl,dapl; pv...)
    isempty(pv) || error("invalid keyword options")
```

src/HDF5.jl
```diff
@@ -704,18 +704,20 @@ function split1(path::AbstractString)
...
     haskey(parent, path) && error("cannot create group: object \"", path, "\" already exists at ", name(parent))
     Group(API.h5g_create(parent, path, lcpl, gcpl, API.H5P_DEFAULT), file(parent))
 end
 
 # Setting dset creation properties with name/value pairs
 function create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::Datatype, dspace::Dataspace; pv...)
-    dcpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(API.H5P_DATASET_CREATE; pv...)
-    dxpl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(API.H5P_DATASET_XFER; pv...)
-    dapl = isempty(pv) ? DEFAULT_PROPERTIES : create_property(API.H5P_DATASET_ACCESS; pv...)
     haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
```
musm, line 715: I moved this up since, it would make sense to error before constructing any properties objects. Hope that's ok with you.

musm:
Sorry this took so long but it looks great and I'm very happy with the improvements. It would be good to improve the coverage of the Filters module at some point in the future.

src/properties.jl
```diff
@@ -0,0 +1,668 @@
...
+
+const DEFAULT_PROPERTIES = GenericProperties()
+# These properties are initialized in __init__()
+const ASCII_LINK_PROPERTIES = LinkCreateProperties()
+const UTF8_LINK_PROPERTIES = LinkCreateProperties()
+const ASCII_ATTRIBUTE_PROPERTIES = AttributeCreateProperties()
+const UTF8_ATTRIBUTE_PROPERTIES = AttributeCreateProperties()
+
+_link_properties(::AbstractString) = UTF8_LINK_PROPERTIES
+_attr_properties(::AbstractString) = UTF8_ATTRIBUTE_PROPERTIES
```
musm, line 668: Can we just remove these and pass in the global constants directly.  To me the way these are used don't offer any advantages and is less clear than using the global constants. Thoughts on this?

src/properties.jl
simonbyrne: Yeah, `FilterPipeline` is just a wrapper object.

src/properties.jl
simonbyrne: Yes, that would probably make sense (I'm not 100% sure what I was thinking here)

src/properties.jl
```diff
@@ -0,0 +1,668 @@
...
+end
+
+
+@propertyclass LinkAccessProperties API.H5P_LINK_ACCESS
+
+@propertyclass GroupAccessProperties API.H5P_GROUP_ACCESS
+superclass(::Type{GroupAccessProperties}) = LinkAccessProperties
+
+@propertyclass DatatypeAccessProperties API.H5P_DATATYPE_ACCESS
+superclass(::Type{DatatypeAccessProperties}) = LinkAccessProperties
```
musm, line 621: What was the reason that these all use `LinkAccessProperties` as their superclass?

simonbyrne: I think I queried them via `H5P_GET_CLASS_PARENT` to see what they did internally. Would probably make sense to add that as a test.

PR #865, opened 2021-09-02
Add move_link

tomhaber:
Move a link to a new location in the file. If source is a hard link, this effectively renames the object (also #330).
Maybe `move_link` would be a better name

I would like to add a version that copies and deletes an object if the source and destination files are different, but I'm not sure how to check this. HDF5 has an internal `H5F_same_shared` for this purpose.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -15,7 +15,7 @@ h5open, h5read, h5write, h5rewrite, h5writeattr, h5readattr,
 create_attribute, open_attribute, read_attribute, write_attribute, delete_attribute, attributes,
 create_dataset, open_dataset, read_dataset, write_dataset,
 create_group, open_group,
-copy_object, open_object, delete_object, move_object,
+copy_object, open_object, delete_object, move_link,
 create_datatype, commit_datatype, open_datatype,
 create_property,
 group_info, object_info,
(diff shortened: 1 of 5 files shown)


musm:
Sorry for the late response. Generally looks good, would you mind to add a test for this?

tomhaber:
I've added a pretty basic test. Doesn't cover the error case when moving between files, since I would still like to add it

musm:
> Exercise care in moving links as it is possible to render data in a file inaccessible with H5L_MOVE. **If the link being moved is on the only path leading to an HDF5 object, that object may become permanently inaccessible in the file.**

I was looking over the documentation and I don't quite understand this scenario. Maybe you have a better idea of how this happens?

tomhaber:
The only thing I can come up with is something like this:

```julia
h5open(fn, "cw") do io
  io["moveme"] = [1,2,3]
  create_group(io, "moveto")
  
  # move group inside of itself
  move_link(io, "moveto", io["moveto"], "foo")
end
```

That will make `moveto` inaccessible

musm:
I see, so moving a group to itself basically renders all the data in that group inaccessible. 
If we are going to export the function, perhaps we should add a check to prevent this scenario.

tomhaber:
The problem is that Julia can also hold references to objects/groups and I think you can only detect "real" problems when those go out of scope (which is hard to detect in Julia).

I could see some scenarios where things might be temporarily in a bad state but then resolved. For example
```Julia
move_link(ds, "foo", io, "moveto")
```
would revert the situation.

`delete_object` can have the same strange effects: e.g. deleting a group and keeping references to objects inside of it. Especially weird if you keep a reference to the group itself

src/HDF5.jl
```diff
@@ -863,6 +863,10 @@ delete_object(obj::Object) = delete_object(parent(obj), ascii(split(name(obj),"/
 copy_object(src_parent::Union{File,Group}, src_path::AbstractString, dst_parent::Union{File,Group}, dst_path::AbstractString) = API.h5o_copy(checkvalid(src_parent), src_path, checkvalid(dst_parent), dst_path, API.H5P_DEFAULT, _link_properties(dst_path))
 copy_object(src_obj::Object, dst_parent::Union{File,Group}, dst_path::AbstractString) = API.h5o_copy(checkvalid(src_obj), ".", checkvalid(dst_parent), dst_path, API.H5P_DEFAULT, _link_properties(dst_path))
 
+# Move links
+move_link(src::Union{File,Group}, src_name::AbstractString, dest::Union{File,Group}, dest_name::AbstractString=src_name, lapl::Properties=DEFAULT_PROPERTIES, lcpl::Properties=DEFAULT_PROPERTIES) = API.h5l_move(checkvalid(src), src_name, checkvalid(dest), dest_name, lcpl, lapl)
+move_link(parent::Union{File,Group}, src_name::AbstractString, dest_name::AbstractString, lapl::Properties=DEFAULT_PROPERTIES, lcpl::Properties=DEFAULT_PROPERTIES) = API.h5l_move(checkvalid(parent), src_name, parent, dest_name, lcpl, lapl)
```
musm, line 868:
```suggestion
move_link(src::Union{File,Group}, src_name::AbstractString, dest::Union{File,Group}, dest_name::AbstractString=src_name, lapl::LinkAccessProperties = LinkAccessProperties(), lcpl::LinkCreateProperties = LinkCreateProperties()) =
    API.h5l_move(checkvalid(src), src_name, checkvalid(dest), dest_name, lcpl, lapl)

move_link(parent::Union{File,Group}, src_name::AbstractString, dest_name::AbstractString, lapl::LinkAccessProperties = LinkAccessProperties(), lcpl::LinkCreateProperties = LinkCreateProperties())  =
    API.h5l_move(checkvalid(parent), src_name, parent, dest_name, lcpl, lapl)
```

musm:
Thanks for being patient!

PR #880, opened 2021-12-08
Add bzip2, lz4, and zstd independent compression filter packages

mkitti:
This branch splits the compression plugin code into independent subdirectory packages. Consider merging this into #875.

The main purpose of this pull request is to test CI.


licenses/H5Zbzip2_LICENSE.txt
```diff
@@ -0,0 +1,29 @@
+Copyright Notice and Statement for PyTables Software Library and Utilities:
```
musm, line 1: Delete these and just link `src/filters/H5Zlz4/THIRDPARTY.txt` in the readme. Anyone copying the  `HDF5.jl` would have them included so this would be redundant.

mkitti: I deleted these in 2ab4410 and moved the readme to THIRDPARTY.md at the root level.

mkitti:
All tests are now passing, including nightly.

src/filters/H5Zbzip2/Project.toml
```diff
@@ -0,0 +1,12 @@
+name = "H5Zbzip2"
+uuid = "094576f2-1e46-4c84-8e32-c46c042eaaa2"
+version = "0.16.0"
```
musm, line 3: The idea will be to sync the version with the main HDF5.jl package version?

I don't have a problem with this, but as you probably already know, these 'independent' packages could also just maintain their own version number that get bumped when we modify the subpackages. 

I think the main benefit of the current approach is simplicity in bumping version numbers across all packages. OTOH, if we keep an independent version number I think then that reduces precompilation when we make new releases. 
From judging changes in the original blosc file, changes to the filter packages are likely to be infrequent.

src/filters/H5Zbzip2/src/H5Zbzip2.jl
```diff
@@ -137,32 +137,63 @@ end # function H5Z_filter_bzip2
...
-        dest,
-        destLen,
-        source,
-        sourceLen,
-        blockSize100k,
-        verbosity,
-        workFactor
-    )
+    @static if CodecBzip2.WIN32
+        return ccall(
```
musm, line 141: good catch... the dreaded stcall

src/filters/H5Zbzip2/Project.toml
mkitti: Yes. We can only tag the whole repository, so we might as well keep it in sync with the tags.

src/filters/H5Zblosc/src/H5Zblosc.jl
```diff
@@ -91,10 +91,16 @@ function blosc_filter(flags::Cuint, cd_nelmts::Csize_t,
...
-        outbuf_size, cbytes, blocksize = Blosc.cbuffer_sizes(unsafe_load(buf))
-        outbuf = Libc.malloc(outbuf_size)
-        outbuf == C_NULL && return Csize_t(0)
-        status = Blosc.blosc_decompress(unsafe_load(buf), outbuf, outbuf_size)
+        in = unsafe_load(buf)
+        outbuf_size, cbytes, blocksize = Blosc.cbuffer_sizes(in)
+        # We need to preserve the output of cbuffer_sizes under Julia 1.8 (as of 2021/12/11)
+        # Otherwise their Refs get freed and the buffer becomes corrupted
+        # If this fails, consider copying `in` or wrapping that array.
+        GC.@preserve outbuf_size cbytes blocksize begin
```
mkitti, line 99: Without this line, H5Zblosc appears to error with Julia 1.8 in development.

`cbuffer_sizes` calls `blosc_cbuffer_sizes` which returns a size values via pointers / `Refs`.

https://github.com/JuliaIO/Blosc.jl/blob/8ab4e3cf60dd45ee598ad879caf99948bd2a2718/src/Blosc.jl#L129-L137
```julia
function cbuffer_sizes(buf)
    s1 = Ref{Csize_t}()
    s2 = Ref{Csize_t}()
    s3 = Ref{Csize_t}()
    ccall((:blosc_cbuffer_sizes,libblosc), Cvoid,
          (Ptr{UInt8}, Ref{Csize_t}, Ref{Csize_t}, Ref{Csize_t}),
          buf, s1, s2, s3)
    return (s1[], s2[], s3[])
end
```

Somehow it appears that the `Refs` are getting garbage collected and the memory in the actual buffer `in` gets corrupted. This does not make sense to me, but this seems to have worked fine until Julia 1.8, so perhaps there is a bug.

test/external.jl
```diff
@@ -50,5 +54,6 @@ close(source_file)
 
 rm(fn1)
 # rm(fn2)
+@debug "external tests did not delete" fn2
```
mkitti, line 57: We cannot delete `fn2` or run those above tests because some HDF5 handle is still open. Because of the external link, HDF5 is set to use "weak" closing, so closing the file is not sufficient. Ultimately `h5close` at the end of runtests.jl does close the files and allows for the tempfile to be deleted.

test/runtests.jl
```diff
@@ -33,4 +54,7 @@ if HDF5.has_parallel()
     include("mpio.jl")
 end
 
+# Clean up after all resources
+HDF5.API.h5_close()
```
mkitti, line 58: Need to clean up after external.jl

test/swmr.jl
```diff
@@ -114,4 +116,9 @@ end
 end
 
 rm(fname) # cleanup file created by swmr tests
+
+if nprocs() > 1
+    rmprocs(procs)
```
mkitti, line 121: This solves the stall on Julia 1.8 nightly. See https://github.com/JuliaIO/HDF5.jl/pull/881

src/filters/H5Zblosc/src/H5Zblosc.jl
mkitti: This has been resolved in https://github.com/JuliaLang/julia/pull/43408 (pending merge)

src/filters/H5Zbzip2/Project.toml
thchr: You can tag a sub-package associated with a subdirectory of a repository via `@JuliaRegistrator register subdir=...`. Makie e.g., uses this approach.

mkitti: What I meant is a tag and release from the git perspective.

musm: Let's keep all the subpackages at version `0.1.0` and tag them individually. I think the benefit is we avoid precompilation of the subpackages after new HDF5.jl releases, and most of these filter packages won't be changing frequently, so I believe that makes more sense and follows semver, instead of bumping the submodules in instances were  HDF5.jl was changes but the subpackages were untouched. 


For releases I've been using the registrator bot. It looks like it should be straightforward to accommodate subpackages with the registrator from @thchr comment.

test/runtests.jl
musm: Can we remove this change and because the external file issue is a separate bug.  I propose  pushing this is a separate commit.    (Logically trying to keep  PRs to a related set of features/bugfixes)

test/swmr.jl
musm: I propose  pushing this is a separate commit.

src/filters/H5Zblosc/src/H5Zblosc.jl
musm: Oh okay, so this is unrelated to https://github.com/JuliaIO/Blosc.jl/issues/86 or does the fix also require making the changes according to your comment in https://github.com/JuliaIO/Blosc.jl/issues/86 ?

musm: Looks like the underlying fix has been merged.  I propose we remove mitigating patch here as it won't be an issue henceforth.

test/swmr.jl
mkitti: You want the test/swmr.jl in separate branch/PR or do you really mean a distinct commit?

musm: Sorry I mean a different PR, if it wasn't clear.

mkitti: I put it into #882

test/runtests.jl
mkitti: I put it into #882

musm:
Any chance you could rebase with master.

musm:
I had this comment here: https://github.com/JuliaIO/HDF5.jl/pull/880/files#r767927849 
and also 
https://github.com/JuliaIO/HDF5.jl/pull/880/files#r767937926 
to remove the mitigating patch as it's now been fixed in Julia. master

src/filters/H5Zbzip2/Project.toml
mkitti: Well we can tell the registrator bot whatever version we want, but the question is how to do you mark these versions in Git? Do you plan on making an individual tag for each sub-package?

test/external.jl
mkitti: I removed the debug message

src/filters/H5Zbzip2/Project.toml
musm: Ah, I see your point now, since these are not git submodules...

mkitti:
These two pull requests should fix the remaining downstream errors.

https://github.com/JuliaIO/MAT.jl/pull/170 (fixes SparseMatrixCSC errors for Julia 1.7)
https://github.com/JuliaIO/JLD.jl/pull/305 (adds H5Zblosc, so set_blosc! exists)

Anything else. You are welcome to make commits directly.

musm:
> Anything else? You are welcome to make commits directly.

You'll have to enable "allow commits" in the PR since I don't have access.

src/filters/H5Zbzip2/Project.toml
musm: Thinking about this again, does the git tag version even matter? No. 

So, what'll happen is that the  git tag version will be synced with the HDF5.jl version as is currently the case.
The git tag version will not agree with the tagged version of the filter subpackages, which will have their own independent versions as specified in their own Project.toml. This is the case since we aren't using git submodules, which is more trouble than it's worth.  

To me this makes sense and is desired behavior since we won't have to tag all the filter packages superfluously every time a change is made in the main HDF5.jl module that doesn't modify the filter submodules.

src/filters/H5Zblosc/src/H5Zblosc.jl
mkitti: You can tag a commit as H5Zblosc_v0.1.0 when you make a release. The main convenience of keeping the versions somewhat aligned is making it easy to know what versions are compatible with what other versions.

mkitti:
> > Anything else? You are welcome to make commits directly.
> 
> You'll have to enable "allow commits" in the PR since I don't have access.

I invited you to the repository. I guess it's trickier since it's coming from an organizational account.

mkitti:
All tests should be passing now.

src/filters/H5Zblosc/src/H5Zblosc.jl
musm: That is true. OTOH, I think it might not be too difficult to determine if a non-backwards compatible change is made to the filter subpackages. At worst, we could always bump the compatible HDF5 version to the latest one in the filter subpackage  every time a change is made.

Project.toml
```diff
@@ -12,16 +11,23 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
...
 
 [compat]
-Blosc = "0.7.1"
 Compat = "3.1.0"
 HDF5_jll = "~1.10.5, ~1.12.0"
 Requires = "1.0"
 julia = "1.3"
 
 [extras]
+Blosc = "a74b3585-a348-5f62-a45c-50e91977d574"
```
musm, line 20:
```suggestion
```

Do we still need Blosc in the main HDF5.jl source?

mkitti: I thought it might be part of the tests still, but perhaps not. Let's omit it and see what happens.

src/filters/filters.jl
```diff
@@ -106,8 +292,32 @@ function Base.append!(filters::FilterPipeline, extra)
...
+    quote
+        @eval begin
+            using Pkg
+            Pkg.develop(PackageSpec(; path=joinpath(dirname(pathof(HDF5)), "filters", "H5Zblosc")))
+            Pkg.develop(PackageSpec(; path=joinpath(dirname(pathof(HDF5)), "filters", "H5Zbzip2")))
+            Pkg.develop(PackageSpec(; path=joinpath(dirname(pathof(HDF5)), "filters", "H5Zlz4")))
+            Pkg.develop(PackageSpec(; path=joinpath(dirname(pathof(HDF5)), "filters", "H5Zzstd")))
+        end
+    end
+end
```
mkitti, line 321: Here is the macro which devs the filter packages into the test environment.

musm:
Looks like there may be an aws issue downloading the non ubuntu binaries?

musm:
@mkitti  any chance you could review my changes.

musm:
I think we need to in a follow up PR or this PR,  either: 

(1)  deprecate setting `:blosc` for properties and warn users to load in the external H5Zblosc package to use the blosc filter.  

(2) Alternatively, we could add hooks for :zstd, :lz4, :bzip2  like we do for the blosc package
i.e. add the following for the external filters.
```julia
    @require H5Zblosc="c8ec2601-a99c-407f-b158-e79c03c2f5f7" @eval begin
        set_blosc!(p::Properties, val::Bool) = val && push!(Filters.FilterPipeline(p), H5Zblosc.BloscFilter())
        set_blosc!(p::Properties, level::Integer) = push!(Filters.FilterPipeline(p), H5Zblosc.BloscFilter(level=level))
    end
```

 @simonbyrne  maybe you also have an opinion on which option is more desirable from a ux perspective.

mkitti:
I generally think that we should deprecate the older syntax `blosc = 3` or otherwise come up with a way to implement properties in an extensible manner. 

Rather than this
https://github.com/JuliaIO/HDF5.jl/blob/00dcb1395acb6a9e4ea90ec3ac0eb9f586412760/src/properties.jl#L438-L452

an extensible system would work as follows

```julia
# Defined in HDF5.jl
function class_setproperty!(::Type{DatasetCreateProperties}, p::Properties, name::Symbol, val)
   class_setproperty!(::Type{DatasetCreateProperties}, p, Val(name), val)
end

# Defined in H5Zblosc.jl
import HDF5: class_setproperty!
class_setproperty!(::Type{DatasetCreateProperties}, p::Properties, Val{:blosc}, val) = set_blosc!(p, val)
```

That said, I think the `filters` interface is much better and is already extensible. Filters should have an order to them in a pipeline. Thus, I would lean towards deprecating the properties interface for filters.

test/runtests.jl
```diff
@@ -1,60 +1,64 @@
...
-end
+using HDF5
+using Test
+using Pkg
+
+filter_path = joinpath(dirname(pathof(HDF5)), "filters")
+Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zblosc")))
+Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zbzip2")))
+Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zlz4")))
+Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zzstd")))
```
mkitti, line 8: I did like the idea of having this in some kind of callable closure. I anticipate downstream projects may want to do something similar via CI. Perhaps though it could it its own file under `test/`?

mkitti:
Overall the changes look fine to me. We just need to register the packages for some of the tests to work.

mkitti:
It will probably be easier to iterate and test this once the subpackages are registered. I suggest registering the subpackages well before we register HDF5 0.16 itself. Then at least we can `dev` everything in the normal fashion.

musm:
Sure,  is  dev'ing not working in some cases for you ? I've just manually kept calling 
```
Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zblosc")))
```
when need be...

In any case, we'll have to register the subpackages first.  I think I'll merge this PR now and we then need to iterate on the new filters interface. I do agree that the second approach is much more flexible, so we'd have to appropriately deprecate existing calls.

musm:
Thank you very much @mkitti  for your effort and patience with these PRs!  I'm really excited for these changes in the upcoming v0.16 release. Also sorry to drag out some of the licensing issues.

musm:
I've submitted registration for all the subpackages, they will take a day or two since they are blocked until a manual merge is done, as they depend on v0.16 which is yet released.

mkitti:
We could ask for manual merge if needed.

PR #882, opened 2021-12-13
Fix nightly tests, add debugging

mkitti:
Cherry-picked from 740983ed4bc99ec56ba09aa1f6915d789066cec5

445e038: Use `rmprocs` to clean up added process, fix nightly stall
4fbcb99: Add debug messages to test, use h5_close after tests finish to clean up temp files
c1e5761: Enable debugging in Github Actions test step

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -75,6 +75,8 @@ jobs:
           arch: ${{ matrix.arch }}
       - uses: julia-actions/julia-buildpkg@latest
       - uses: julia-actions/julia-runtest@latest
+        env:
+          JULIA_DEBUG: Main
       - uses: julia-actions/julia-processcoverage@latest
       - uses: codecov/codecov-action@v1
         with:
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -30,4 +30,7 @@ if HDF5.has_parallel()
     include("mpio.jl")
 end
 
+# Clean up after all resources
+HDF5.API.h5_close()
+
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -4,24 +4,44 @@ using Pkg
 
 @info "libhdf5 v$(HDF5.API.h5_get_libversion())"
 
+# To debug HDF5.jl tests, uncomment the next line
+# ENV["JULIA_DEBUG"] = "Main"
+
 @testset "HDF5.jl" begin
 
+@debug "plain"
 include("plain.jl")
+@debug "compound"
 include("compound.jl")
+@debug "custom"
 include("custom.jl")
+@debug "reference"
 include("reference.jl")
+@debug "dataspace"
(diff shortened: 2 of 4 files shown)


musm:
Presumably the plan is to drop the last two commits after nightlies pass?

mkitti:
I would keep them. The one to drop might be the last one if you really wanted a quiet CI. However, having the debug messages on during CI makes it easier to catch problems. It would have been much easier to catch the stall issue earlier if these messages were always present. Instead CI just failed silently and we were not able to isolate the problem.

musm:
Also FYI https://github.com/JuliaIO/HDF5.jl/runs/4511327515?check_suite_focus=true

.github/workflows/CI.yml
```diff
@@ -75,6 +75,8 @@ jobs:
           arch: ${{ matrix.arch }}
       - uses: julia-actions/julia-buildpkg@latest
       - uses: julia-actions/julia-runtest@latest
+        env:
+          JULIA_DEBUG: Main
```
musm, line 79: Main?

mkitti: That's the module where the tests are run.

musm:
Could you remove the `@debug` statements, I don't think there's a benefit to leaving them in and on by default, otherwise LGTM. Ideally nested testsets would print out explicitly. Instead we could just remove the outer `HDF5.jl` testset so that the individual testsets print out.

mkitti:
> Could you remove the `@debug` statements, I don't think there's a huge benefit to leaving them in and on by default, otherwise LGTM. If we run into bugs, adding them during the PR phase to identify and fix bugs is likely the best course of action. Ideally nested testsets would print out explicitly. Instead we could just remove the outer `HDF5.jl` testset so that the individual testsets print out, see #883.

Could we remove the `@debug` statements in another PR after we merge the current set of PRs? The test logging in #883 is not sufficient to identify the problem in nightly. In fact, based on the tests that did report there, I suspect there is another bug. Meanwhile, nightly continues to be in flux.

test/runtests.jl
```diff
@@ -30,4 +50,7 @@ if HDF5.has_parallel()
     include("mpio.jl")
 end
 
+# Clean up after all resources
+HDF5.API.h5_close()
```
musm, line 54: Could you create a new PR with this change or edit this commit to remove this line and add it as  a new commit. (It helps to have the commits logically separated and in this case the debug statements from the resource clean up fix).

musm: Usually for something like this  a git rebase interactive is helpful.  `git rebase -i HEAD~6` then go  to this commit and edit it (you can also rearrange the commits etc.) 

The reason I'm asking is I'd like to merge this PR without squashing the PR since there are several logically independent commits in the PR, which are useful to preserve in the git history.

musm:
Thanks!

PR #883, opened 2021-12-14
Remove outer testset grouping

musm:
In order to printout the result from each testset separately

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -4,6 +4,8 @@ using Pkg
 
 @info "libhdf5 v$(HDF5.API.h5_get_libversion())"
 
+@testset verbose=True "HDF5.jl" begin
+    
 include("plain.jl")
 include("compound.jl")
 include("custom.jl")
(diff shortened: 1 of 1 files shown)


mkitti:
The nightly tests failed here. Is it obvious why?

musm:
I understand your point. The debug messages are certainly helpful for ironing out more intricate library/c related errors. After  the bug is fixed it makes sense to me to remove the debug messages or at least turn them off. Why leave them in there by default for every PR/testset run, when the testsets would catch most of the main errors received in PRs and for regression testing? To your point they didn't print out detailed debugging information after master made changes that triggered failures in HDF5.jl. In any case, it's a pattern we don't see by default in other packages in the Julia ecosystem, even if they would help in identifying a potential bug that could arise in  the future.  I just think it's unnecessarily verbose for the majority of times and for the instance it is helpful it makes sense to turn on the debugging flag separately to iron out the issue in a more targeting manner.

mkitti:
`@testset verbose=true` does not work in earlier versions of Julia.

PR #885, opened 2022-01-21
Move filter packages out of main HDF5.jl source directory

musm:
If the package filters are installed under src, this would end up
installing the same source twice. Thus we move them out into their own
independent folder.


mkitti (approved):
looks good to me

Issue #884, opened 2021-12-28
0.16 changes `filter` keyword to `filters`


david-macmahon:
It looks like the `filter` keyword was changed to `filters` in 1c2ccb5.  Is this is an intentional or inadvertent breaking change from 0.15?

texadactyl:
I wonder if "filter" is the intended keyword and "filters" is simply being ignored .
It's hard to tell considering the lack of filter coverage in the documentation.

mkitti:
I think that was quite an intentional change by @simonbyrne . That's why the coming version is 0.16 and not 0.15.8

musm:
I think the idea was that since you can pass in a  series of   compression filters, it would make sense to change the keyword argument to `filters`. 

I wonder if there's a style guide recommendation here whether to prefer the plural or singular. Since you can pass in a single filter `filters=Filters.Deflate(3)` or a list as in `filters=[Filters.Shuffle(), Filters.Deflate(3)]`.

mkitti:
Would it be worth enabling a backwards compatible filter keyword that just uses the the new infrastructure? So `filter=x` just translated to `filters=[x]`?

musm:
Yes that's my plan, but like to first confirm we are sticking to the `filters` name.  I'll try to dig around to see if there's a style guide suggestion on this.  I don't have a strong preference on whether the singular or plural makes sense. For the singular case, you can think of passing in the list of filters as a single object (a single list). In that way the singlar could be argued to also make as much sense as the plural.

david-macmahon:
Since one can pass `Tuple{Int}` to employ a single filter or `Tuple{Tuple{Integer}}` to create a composite filter from a sequence of filters, there will always be singular/plural mismatch potential: either `<singular_keyword>=<plural_filter_specs>` or `<plural_keyword>=<singular_filter_spec>`.

I think the singular keyword `filter` is preferable (less unpreferable?) because one can think of multiple filters applied in sequence as forming a single composite filter.

I suppose we could get really pedantic and force the users to use the singular `filter` keyword for a `Tuple{Integer}` singular filter specification and the `filters` keyword for a `Tuple{Tuple{Integer}}` plural filter specification.  But that seems overly detailed IMHO.

musm:
@david-macmahon  I'm also in favor or your logic, and was leaning towards that. I'll go ahead and make a PR with the changes tonight.

musm:
[...]

PR #887, opened 2022-01-21
Add filter to filters deprecation for property rename

musm:
close https://github.com/JuliaIO/HDF5.jl/issues/884

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -26,7 +26,7 @@ for name in names(API; all=true)
     end
 end
 
-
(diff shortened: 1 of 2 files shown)


musm:
cc @simonbyrne

mkitti:
We should probably add a test for the `filter` property while it should still work.

musm:
We typically haven't gone as far as to add tests for deprecated calls. I've personally locally tested that the `filter` keyword works as intended, hopefully that should be enough for now.

PR #889, opened 2022-01-21
Add fallback for set_blosc!

simonbyrne:
Add an informative error message for old blosc keyword argument.


musm:
LGTM,

I honestly think we should deprecate  the property interface in v0.16  for setting filters e.g. `:blosc` in favor of the newer  filters interface as its more flexible and extensible.

musm:
We probably should also add a note to the end of the list in this file https://github.com/JuliaIO/HDF5.jl/blob/master/src/deprecated.jl

musm:
Let me know what you think  or if you have any pushback on  the proposal to deprecate in favor of the new filter pipeline for v0.16.

simonbyrne:
I don't have a strong opinion either way: the argument for keeping it is convenience, for getting rid of it is consistency and being explicit.

mkitti (approved):

PR #891, opened 2022-01-25
remove eval in front of requires blocks

KristofferC:
Having the `@eval` there makes the `@require` pattern matching for `include` fail and the pacakge thereby is unusuable when used together with PackageCompiler (example log: https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/dd0c14b_vs_dd0c14b/YaoQX.primary.log)

```
┌ Warning: Error requiring `MPI` from `HDF5`
│   exception =
│    LoadError: SystemError: opening file "/home/pkgeval/.julia/packages/HDF5/pIJra/src/mpio.jl": No such file or directory
│    Stacktrace:
│      [1] systemerror(p::String, errno::Int32; extrainfo::Nothing)
```

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -1611,8 +1611,8 @@ function __init__()
     ASCII_ATTRIBUTE_PROPERTIES.char_encoding = :ascii
     UTF8_ATTRIBUTE_PROPERTIES.char_encoding = :utf8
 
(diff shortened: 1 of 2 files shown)


musm:
Are there any solution that are compatible with both PackageCompiler and precompilation of the glue code using Requires?

KristofferC:
I am confused. You aren't using a separate "glue" package here. You are just using `include`. `@eval` will have no influence here.

PR #892, opened 2022-01-26
Fix chunkstorage test by switching to Int rather than Int64

mkitti:
Let's see if this fixes the test


mkitti:
@musm ready for review here with green checkmarks all around.

musm:
Great  yay

Issue #894, opened 2022-01-28
HISTORY.md is outdated


laborg:
I think `HISTORY.md` should be either removed or updated. Having an outdated version lying around isn't useful and might confuse users.

musm:
Thanks for the reminder.  I should be more on top of updating that file..

musm:
closed by  #895

PR #895, opened 2022-01-28
Update HISTORY.md

musm:


musm:
close https://github.com/JuliaIO/HDF5.jl/issues/894

PR #899, opened 2022-02-04
Support BITSHUFFLE option of Blosc >= v0.7.3

david-macmahon:
Blosc 0.7.3 now treats the blosc "shuffle" option as an Integer rather
than a Bool.  This allows the use of all three of blosc's shuffling
options: no shuffling (0), byte shuffling (1), or bit shuffling (2).
The `shuffle` field of the BloscFilter struct is not restricted to Bool,
but it was effectively being converted to Bool via `!=(0)` in the
`blosc_filter()` function of the H5Zblosc plugin before passing to the
`blosc_compress()` function.  This prevented the use of bit shuffling
because `2 != 0` is `true` which is the same as 1 (byte shuffling).

The H5Zblosc plugin no longer converts the given value to Bool and also
validates the value against the three supported shuffle option values.
These values are also imported into the H5Zblosc module so users can
refer to them as one of:

- H5Zblosc.NOSHUFFLE
- H5Zblosc.SHUFFLE
- H5Zblosc.BITSHUFFLE

closes https://github.com/JuliaIO/HDF5.jl/issues/285

### DIFF
--- a/filters/H5Zblosc/Project.toml
+++ b/filters/H5Zblosc/Project.toml
@@ -8,5 +8,5 @@ HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 
 [compat]
 HDF5 = "0.16"
-Blosc = "0.7.1"
+Blosc = "0.7.3"
 julia = "1.3"
--- a/filters/H5Zblosc/src/H5Zblosc.jl
+++ b/filters/H5Zblosc/src/H5Zblosc.jl
@@ -8,6 +8,8 @@ import HDF5.Filters: filterid, register_filter, filtername, filter_func, filter_
 
 export H5Z_FILTER_BLOSC, blosc_filter, BloscFilter
 
+# Import Blosc shuffle constants
+import Blosc: NOSHUFFLE, SHUFFLE, BITSHUFFLE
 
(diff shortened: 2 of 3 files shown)


musm (approved):

musm:
closes https://github.com/JuliaIO/HDF5.jl/issues/285

musm:
Thank you!

mkitti:
We really should consider a strategy of shifting everything to some version of Blosc2 possibly starting with incorporation of [Blosc2.jl](https://github.com/waralex/Blosc2.jl) into JuliaIO.

musm:
Agreed

musm:
https://github.com/waralex/Blosc2.jl says it's WIP, I wonder what's still missing. Is he open potentially to moving to JuliaIO?

mkitti:
With regard to this filter we should also review how similar it is to these reference implementations:
https://github.com/Blosc/hdf5-blosc/blob/master/src/blosc_filter.c https://github.com/HDFGroup/hdf5_plugins/blob/master/BLOSC/src/H5Zblosc.c

david-macmahon:
These are all great comments vis a vis Blosc2, but perhaps they would be better placed in a new Blosc2 feature request issue so they don't get lost here.  Just saying...

mkitti:
The operating issue is https://github.com/waralex/Blosc2.jl/issues/6

PR #900, opened 2022-02-04
External dataset support

mkitti:
This pull request implements `h5p_get_external` and `h5p_get_external_count`.

We introduce a new type `off_t` which is just `Int`. We may need to take a closer look at that.

Also we create methods such that the `datatype` of a `Datatype` is just the instance. Likewise for `dataspace`.

Tests are added.

```julia
julia> fn_external = tempname()
"C:\\Users\\KITTIS~1\\AppData\\Local\\Temp\\jl_TiEf6nMequ"

julia> dset = HDF5.create_external_dataset(hfile, "ext", fn_external, Int, (10,20))
🔢 HDF5.Dataset: /ext (file: C:\Users\KITTIS~1\AppData\Local\Temp\jl_ux3rDdRhn6 xfer_mode: 0)

julia> dcpl = HDF5.get_create_properties(dset)
HDF5.DatasetCreateProperties(
  alloc_time      = :late,
  external        = (name = "C:\\Users\\KITTIS~1\\AppData\\Local\\Temp\\jl_TiEf6nMequ", offset = 0, size = 0x0000000000000640),
  filters         = HDF5.Filters.Filter[],
  layout          = :contiguous,
  filter          = HDF5.Filters.Filter[],
  obj_track_times = true,
)

julia> HDF5.API.h5p_get_external_count(dcpl)
1

julia> HDF5.API.h5p_get_external(dcpl)
(name = "C:\\Users\\KITTIS~1\\AppData\\Local\\Temp\\jl_TiEf6nMequ", offset = 0, size = 0x0000000000000640)

julia> HDF5.API.h5p_get_external(dcpl, 0)
(name = "C:\\Users\\KITTIS~1\\AppData\\Local\\Temp\\jl_TiEf6nMequ", offset = 0, size = 0x0000000000000640)
```


mkitti:
@musm any thoughts about this one?

musm:
Sorry I need  this weekend to finish this review, I had some comments last I looked at this.

musm:
No I just had some issues with the helpers.jl file.

I can merge this and fix that when I have time, sorry I didn't get to it this weekend.

src/properties.jl
```diff
@@ -626,6 +626,19 @@ superclass(::Type{DatatypeAccessProperties}) = LinkAccessProperties
 @propertyclass DatasetAccessProperties API.H5P_DATASET_ACCESS
 superclass(::Type{DatasetAccessProperties}) = LinkAccessProperties
 
+class_propertynames(::Type{DatasetAccessProperties}) = (
+    :efile_prefix,
+)
+
+function class_getproperty(::Type{DatasetAccessProperties}, p::Properties, name::Symbol)
```
musm, line 633: Maybe also test `efile_prefix` getter and setters directly?

src/api/helpers.jl
```diff
@@ -352,6 +352,28 @@ function h5p_get_dxpl_mpio(dxpl_id)
...
+    prefix_size = h5p_get_efile_prefix(plist, prefix, prefix_size)
+    return unsafe_string(pointer(prefix), prefix_size) 
+end
+
+function h5p_get_external(plist, idx = 0)
+    name_size = Csize_t(1024)
+    name = zeros(UInt8, name_size)
+    offset = Ref{off_t}(0)
+    sz = Ref{hsize_t}(0)
+    h5p_get_external(plist, idx, name_size, name, offset, sz)
```
musm, line 367: Frustratingly I don't see a way where we can query the required buffer size for `name` .

mkitti: My suggestion is to check if the string is null terminated, then double the size of the buffer.

musm: That's what we have done before, and probably the best idea. Hopefully they are not using strcmp for the copies, because it could be susceptible to buffer overflow. I recall there is one other method in the hdf5 library where we had this issue.

PR #901, opened 2022-02-05
Add filters documentation

mkitti:
Fix https://github.com/JuliaIO/HDF5.jl/issues/893

### DIFF
--- a/docs/src/filters.md
+++ b/docs/src/filters.md
@@ -1,0 +1,112 @@
+# Filters
+
+HDF5 supports filters for compression and validation: these are applied sequentially to
+each chunk of a dataset when writing data, and in reverse order when reading data.
+
+```@meta
+CurrentModule = HDF5
+```
+
+These can be set by passing a filter or vector of filters as a `filters` property to
+[`DatasetCreateProperties`](@ref).
+
+## Filter Interface
+
+```@meta
+CurrentModule = HDF5.Filters
+```
+
+The filter interface is used to describe filters and obtain information on them.
+
+```@docs
+Filters
+Filter
+FilterPipeline
+UnknownFilter
+FILTERS
+filterid
+encoder_present
+decoder_present
+filtername
+can_apply_func
+can_apply_cfunc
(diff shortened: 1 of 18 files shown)


mkitti:
@simonbyrne would you like to take a look at the documentation for the HDF5 filters interface?

musm (approved):

mkitti:
Let's just merge this. It is is easier to review and iterate after CI has generated and deployed the docs.

docs/src/filters.md
```diff
@@ -0,0 +1,112 @@
...
+decoder_present
+filtername
+can_apply_func
+can_apply_cfunc
+set_local_func
+set_local_cfunc
+filter_func
+filter_cfunc
+register_filter
+```
```
simonbyrne, line 38: This is mostly internals: while good to document, it should probably be moved to the bottom.

simonbyrne:
You can enable push_preview in documented which will build preview docs for PRs

musm:
> You can enable push_preview in documented which will build preview docs for PRs

https://github.com/JuliaIO/HDF5.jl/blob/master/docs/make.jl#L39 

It's enabled, but won't work across forks.

mkitti:
We'll either need to dev the the plugin packages into the documentation environment or release new versions of them. Which do you prefer?

docs/src/filters.md
mkitti: Moved in [4fff354](https://github.com/JuliaIO/HDF5.jl/pull/901/commits/4fff354772ecc80a1d51e0c32558a86f409f07c9)

musm:
Hmm, I don't have a strong preference for either. 

I think I'll just bump every package filter package and HDF5 in the next round, and that'll sync them up.

simonbyrne:
In other cases where I've had sub packages, I've found it easier to check in the Manifest.

mkitti:
OK, I pushed my branch within the repository with a Manifest.toml: #903.

PR #902, opened 2022-02-05
Fix #896

### LINKED ISSUE #896: 0.16 breaks "external" filters
In 0.15.7, one could use filters that are available in the the HDF5 plugin directory (even if they were not previously known or supported by `HDF5.jl`) by passing `filter=(FILTER_ID, FLAGS, cd_values...)` as a keyword to `create_dataset`.  In 0.16.1, this leads to a would-be infinite loop:
```
ERROR: LoadError: StackOverflowError:
Stacktrace:
      [1] append!
        @ ~/.julia/dev/HDF5/src/filters/filters.jl:292 [inlined]
      [2] push!(a::HDF5.Filters.FilterPipeline{HDF5.DatasetCreateProperties}, iter::Int64)
        @ Base ./array.jl:986
--- the last 2 lines are repeated 79982 more times ---
 [159967] append!
        @ ~/.julia/dev/HDF5/src/filters/filters.jl:292 [inlined]
```
Is there a new way to work with these external filters in v0.16.x?

mkitti:
Create additional methods to interpret a `Tuple` as an `UnknownFilter`.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -53,14 +53,12 @@ function set_shuffle!(p::Properties, ::Tuple{})
     set_shuffle!(p, true)
 end
 
-### Fix Issue #896
+### Changed in PR #902
 import Base: append!, push!
 import .Filters: UnknownFilter
-
 @deprecate append!(filters::Filters.FilterPipeline, extra::NTuple{N, Integer}) where N append!(filters, [UnknownFilter(extra...)])
 @deprecate push!(p::Filters.FilterPipeline, f::NTuple{N, Integer}) where N push!(p, UnknownFilter(f...))
 @deprecate UnknownFilter(t::Tuple) UnknownFilter(t...)
-# End Fix Issue #896
 
 # see src/properties.jl for the following deprecated keywords
 # :compress
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -53,6 +53,15 @@ function set_shuffle!(p::Properties, ::Tuple{})
     set_shuffle!(p, true)
 end
 
+### Fix Issue #896
+import Base: append!, push!
+import .Filters: UnknownFilter
+
+@deprecate append!(filters::Filters.FilterPipeline, extra::NTuple{N, Integer}) where N append!(filters, [UnknownFilter(extra...)])
+@deprecate push!(p::Filters.FilterPipeline, f::NTuple{N, Integer}) where N push!(p, UnknownFilter(f...))
+@deprecate UnknownFilter(t::Tuple) UnknownFilter(t...)
+# End Fix Issue #896
+
 # see src/properties.jl for the following deprecated keywords
 # :compress
 # :fapl_mpio
--- a/src/filters/filters.jl
+++ b/src/filters/filters.jl
@@ -309,14 +309,6 @@ function Base.push!(p::FilterPipeline, f::UnknownFilter)
     end
 end
 
-# Fix Issue #896
-import Base: append!, push!
-
-@deprecate append!(filters::FilterPipeline, extra::NTuple{N, Integer}) where N append!(filters, [UnknownFilter(extra...)])
-@deprecate push!(p::FilterPipeline, f::NTuple{N, Integer}) where N push!(p, UnknownFilter(f...))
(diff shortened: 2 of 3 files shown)


src/filters/filters.jl
```diff
@@ -287,12 +291,15 @@ function Base.delete!(filters::FilterPipeline, ::Type{F}) where {F<:Filter}
...
     return filters
 end
-function Base.append!(filters::FilterPipeline, extra)
+function Base.append!(filters::FilterPipeline, extra::Union{AbstractVector{<:Filter}, NTuple{N, Filter} where N})
     for filter in extra
         push!(filters, filter)
     end
     return filters
 end
+function Base.append!(filters::FilterPipeline, extra::NTuple{N, Integer}) where N
```
musm, line 300: I guess this is only added to support existing use cases.

Would it make sense to wrap this is a deprecation warning and force users to manually wrap things in `UnkownFilter`.

I'm leaning towards not doing that and just leaving this as in the PR?  Just a thought.

mkitti: We have already deprecated the upstream interface, so I think that warning should be sufficient. Keeping this just prevents the rather nasty stack overflow situation.

musm: This should've been in reference to the addition
```
function Base.push!(p::FilterPipeline, f::Tuple)
```

mkitti: I deprecated all the `Tuple` based methods. The tests give the following warnings.

```
┌ Warning: `filter=(307, 0)` keyword option is deprecated, use `filters=(307, 0)` instead
│   caller = setproperty!(p::HDF5.DatasetCreateProperties, name::Symbol, val::Tuple{Int32, Int64}) at properties.jl:47
└ @ HDF5 C:\Users\kittisopikulm\.julia\dev\HDF5\src\properties.jl:47
┌ Warning: `append!(filters::FilterPipeline, extra::NTuple{N, Integer}) where N` is deprecated, use `append!(filters, [UnknownFilter(extra...)])` instead.
│   caller = set_filters! at properties.jl:403 [inlined]
└ @ Core C:\Users\kittisopikulm\.julia\dev\HDF5\src\properties.jl:403
```

musm:
LGTM, could we finally move the deprecations to https://github.com/JuliaIO/HDF5.jl/blob/master/src/deprecated.jl  makes things a lot easier during major version updates. 

Thanks!

mkitti:
Deprecated statements moved. This should be ready to go now.

musm:
Could you enable edits from other users in the PR?

src/deprecated.jl
```diff
@@ -53,6 +53,15 @@ function set_shuffle!(p::Properties, ::Tuple{})
...
 end
 
+### Fix Issue #896
+import Base: append!, push!
+import .Filters: UnknownFilter
+
+@deprecate append!(filters::Filters.FilterPipeline, extra::NTuple{N, Integer}) where N append!(filters, [UnknownFilter(extra...)])
+@deprecate push!(p::Filters.FilterPipeline, f::NTuple{N, Integer}) where N push!(p, UnknownFilter(f...))
+@deprecate UnknownFilter(t::Tuple) UnknownFilter(t...)
+# End Fix Issue #896
```
musm, line 63:
```suggestion
```

src/deprecated.jl
```diff
@@ -53,6 +53,15 @@ function set_shuffle!(p::Properties, ::Tuple{})
     set_shuffle!(p, true)
 end
 
+### Fix Issue #896
+import Base: append!, push!
+import .Filters: UnknownFilter
+
```
musm, line 59:
```suggestion
```

src/deprecated.jl
```diff
@@ -53,6 +53,15 @@ function set_shuffle!(p::Properties, ::Tuple{})
     set_shuffle!(p, true)
 end
 
+### Fix Issue #896
```
musm, line 56:
```suggestion
### Changed in PR #902
```

musm (approved):

mkitti:
It is coming from a fork in another org, so I'm not sure I can do it directly.. But you should still have access to that repository.

PR #903, opened 2022-02-05
Add Manifest.toml for documentation generation

mkitti:


musm:
So the Manifest.toml fixes the issue since it specifies the relative directories to pull in?

docs/Manifest.toml
```diff
@@ -0,0 +1,344 @@
...
+
+[[deps.H5Zzstd]]
+deps = ["CodecZstd", "HDF5"]
+path = "../filters/H5Zzstd"
+uuid = "f6f2d980-1ec6-471c-a70d-0270e22f1103"
+version = "0.1.0"
+
+[[deps.HDF5]]
+deps = ["Compat", "HDF5_jll", "Libdl", "Mmap", "Random", "Requires"]
+path = ".."
```
musm, line 125: Ah I see the relative paths  are in the manifest file. Ok I think I understand how this is working.  

Will this then always just use the latest  commits for each package?

mkitti: We are essentially "dev"ing them into the project.

mkitti:
https://juliaio.github.io/HDF5.jl/previews/PR903/filters/

docs/Manifest.toml
musm: Ok makes sense. Yeah I guess only one package that would've happened automatically, but with the subpacakges it seems like we have to force it to 'dev' them throught a Manifest.toml file.

Issue #905, opened 2022-02-06
Setting HDF5_PLUGIN_PATH via ENV has no effect on HDF5


mkitti:
Since the HDF5 library uses `getenv` and Julia sets environmental variables via `SetEnvironmentVariableW` on Windows, changing `HDF5_PLUGIN_PATH` via `ENV` will not be seen by the HDF5 library.

```julia
julia> using LibHDF5


julia> ENV["HDF5_PLUGIN_PATH"]
"C:\\old\\path"

julia> ENV["HDF5_PLUGIN_PATH"] = raw"C:\new\path"
"C:\\new\\path"

julia> function get_hdf5_plugin_path()
           buf_size = Csize_t(1024)
           path_buf = zeros(Cchar, buf_size)
           buf_size = H5PLget(0, pointer(path_buf), buf_size)
           unsafe_string(pointer(path_buf), buf_size)
       end
get_hdf5_plugin_path (generic function with 1 method)

julia> using LibHDF5 # https://github.com/mkitti/LibHDF5.jl

julia> get_hdf5_plugin_path()
"C:\\old\\path"

```
See https://github.com/JuliaLang/julia/issues/44054

The solution at the moment is to use a different mechanism on Windows to set the the environmental variable. Note that this can only be done once before the HDF5 library is initialized.

```
julia> using LibHDF5

julia> ccall(:_putenv_s, Cint, (Cstring,Cstring), "HDF5_PLUGIN_PATH", raw"C:\new\path")
0

julia> function get_hdf5_plugin_path()
           buf_size = Csize_t(1024)
           path_buf = zeros(Cchar, buf_size)
           buf_size = H5PLget(0, pointer(path_buf), buf_size)
           unsafe_string(pointer(path_buf), buf_size)
       end
get_hdf5_plugin_path (generic function with 1 method)

julia> get_hdf5_plugin_path()
"C:\\new\\path"

julia> ccall(:_putenv_s, Cint, (Cstring, Cstring), "HDF5_PLUGIN_PATH", raw"C:\new\path\newer")
0

julia> get_hdf5_plugin_path()
"C:\\new\\path"
```

My proposal is that we may need to create some initialization code to ensure that environment variables read by HDF5 on Windows are set in an environment table where they can be read.

mkitti:
Current plan here is to build an interface around `h5pl_*` instead of using the environmental variable.

PR #907, opened 2022-02-09
Add env sync for windows

mkitti:
Fixes https://github.com/JuliaIO/HDF5.jl/issues/905

Due to See https://github.com/JuliaLang/julia/issues/44054, the HDF5 library does not see environmental variables changed by assigning values to `ENV` while on Windows.

To fix this, set all environmental variables starting with HDF5 using [`_wputenv_s`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=msvc-170) on `__init__` while on Windows.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -42,6 +42,9 @@ h5doc(name) = "[`$name`](https://portal.hdfgroup.org/display/HDF5/$(name))"
 
 include("api/api.jl")
 include("properties.jl")
+@static if Sys.iswindows()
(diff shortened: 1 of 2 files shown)


mkitti:
Withdrawing this pull request w.r.t https://github.com/JuliaLang/julia/issues/44054 and HDFFV-10143.

I will reimplement this using the HDF5 plugin interface.

https://portal.hdfgroup.org/display/HDF5/Plugins

PR #908, opened 2022-02-10
Add plugin interface via LibHDF5, expand H5Z

mkitti:
This pull request adds the HDF5 Plugin interface. The plugin interface allows one to programmatically modify a list of plugin paths from which plugins may be dynamically loaded

Rather than setting `HDF5_PLUGIN_PATH` via an environmental variable, the path can be set via `h5pl_append`, `h5pl_insert`
`h5pl_prepend`, or `h5pl_replace`.

Additional `H5Z` functions are also added.


mkitti:
The extra `@bind` statements were generated by Clang.jl via https://github.com/mkitti/LibHDF5.jl/blob/main/src/bind.jl

src/api_midlevel.jl
```diff
@@ -185,3 +185,7 @@ function write_chunk(dataset_id, index::Integer, buf::Union{DenseArray,Base.Fast
     offset = [reverse(get_chunk_offset(dataset_id, index))...]
     write_chunk(dataset_id, offset, buf; dxpl_id = dxpl_id, filter_mask = filter_mask)
 end
+
+# Accept a Filter subtype as an argument
+API.h5z_get_filter_info(::Type{F}) where F <: Filters.Filter = API.h5z_get_filter_info(Filters.filterid(F))
+API.h5z_filter_avail(::Type{F}) where F <: Filters.Filter = API.h5z_filter_avail(Filters.filterid(F))
```
musm, line 191: These shouldn't overload the `API` function. Mid-level functions define their own function name, internally call `API`  methods,    have the `h5.._` prefix stripped, and sometimes include additional name normalization. 

One option is
```julia
get_filter_info(::Type{F}) where {F <: Filters.Filter} = API.h5z_get_filter_info(Filters.filterid(F))
get_filter_info(filter) = h5z_get_filter_info(filter)  # calls  api/helpers.jl method

filter_available(::Type{F}) where {F <: Filters.Filter} = API.h5z_filter_avail(Filters.filterid(F))
```

We might also want to include some brief documentation if they live in `midlevel` since these routines are sometimes part of the non-exported public API.

mkitti: I was originally thinking of them as helpers, but the imports were cleaner.

I was also thinking of moving this into the `Filters` module. Any thoughts about that?

musm: > I was also thinking of moving this into the Filters module. Any thoughts about that?

I don't have a strong opinion on this. 
On one hand, logically, it makes sense, especially if we are going to rename it.  

I'll leave the discretion up to you.

mkitti:
Only the nightly test are failing now.

mkitti:
Changes since yesterday:

Added midlevel functions `is_filter_available, `can_filter_encode`, and `can_filter_decode`.

Added a `Base.convert` from `Filters.Filter` to `<: Integer`
```
function Base.convert(::Type{I}, ::Type{F}) where {I <: Integer, F <: Filter}
    Base.convert(I, filterid(F))
end
```

src/api_midlevel.jl
```diff
@@ -185,3 +185,33 @@ function write_chunk(dataset_id, index::Integer, buf::Union{DenseArray,Base.Fast
...
+"""
+is_filter_available(filter_or_id) = API.h5z_filter_avail(filter_or_id)
+
+"""
+    can_filter_encode(filter_or_id)
+
+Given a subtype of `Filters.Filter` or the filter ID number as an integer,
+return `true` if the filter can encode or compress data.
+"""
+function can_filter_encode(filter_or_id)
```
musm, line 203: `can_filter_encode`/`can_filter_decode`  read  a bit strange

what about  ?
`filter_encoder_enabled`
`filter_decoder_enabled`

Ideally, with dispatch on the `Filter` type, we would use more generic names   but the fact they also take an `Integer` makes that less feasible.

musm: thoughts on this?

mkitti: Or maybe `is_encoder_enabled`?

musm: that works too, although 
image

I'm not sure if consistency with base naming matters that much, but just  thought to mention.

mkitti:
I need this as a prerequisite to testing external filter support.

PR #910, opened 2022-02-15
Fix deprecation in test file

musm:


test/filter.jl
```diff
@@ -95,7 +95,7 @@ close(f)
 h5open(fn, "w") do f
     data = rand(UInt8, 512, 16, 512)
     # Tuple of integers should become an Unknown Filter
-    ds, dt = create_dataset(f, "data", data, chunk=(256,1,256), filter=(H5Z_FILTER_BZIP2, 0))
+    ds, dt = create_dataset(f, "data", data, chunk=(256,1,256), filters=(H5Z_FILTER_BZIP2, 0))
```
musm, line 98: @mkitti  was the idea to  use the older filter syntax here with the tuple definition?

mkitti: We also need to figure if we will support this syntax with filters going forward.

mkitti: @simonbyrne I would like your perspective here as well.

The value of `filters` is canonically supposed to be an iterable of `Filters.Filter`. Should it be strictly an `AbstractVector` or would a `Tuple{Filters.Filter}` also work? Should we continue to support defining a filter as a tuple of integers via the `filters` or leave that for backwards compatibility with with `filter`?

The other way we could interpret an iterable of integers is as a bunch of filter id numbers. That seems slightly more consistent going forward.

musm: I also agree.  I think it just makes it more confusing to support this use case, honestly. The `depwarn` suggestion would be pretty straightforward with `ExternalFilter` that I think removing the tuple integer interface makes the most sense.

mkitti: So let's not make this change then as committed. Maybe just add a note that this is deprecated usage? We should also also throw an error if someone tries to use a tuple of integers with the `filters` keyword.

PR #911, opened 2022-02-15
Document ExternalFilter, turn UnknownFilter into an alias

mkitti:


mkitti:
This currently exports `ExternalFilter`.

mkitti:
Proposal: We should download some of the plugins here and test to see if they work:
https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.12/hdf5-1.12.1/plugins/

mkitti:
Perhaps we should even see if we can get the binary plugins into binary builder:
https://github.com/hdfGroup/hdf5_plugins

src/filters/filters.jl
```diff
@@ -46,7 +46,7 @@ See [test/filter.jl](https://github.com/JuliaIO/HDF5.jl/blob/master/test/filter.
 module Filters
 
 # builtin filters
-export Deflate, Shuffle, Fletcher32, Szip, NBit, ScaleOffset
+export Deflate, Shuffle, Fletcher32, Szip, NBit, ScaleOffset, ExternalFilter
```
musm, line 49: If we export `ExternalFilter` we probably should redefine show so that it doesn't print out 
`HDF5.Filters.UnkownFilter`

docs/src/filters.md
```diff
@@ -89,6 +89,74 @@ ZstdFilter
 
 Additional filters can be dynamically loaded by the HDF5 library. See [External Links](@ref) below for more information.
 
+### Using an UnknownFilter (aka ExternalFilter)
```
musm, line 92: Let's switch this around and have the docs mainly refer to `ExternalFilter` rather than `UnkownFilter`

src/filters/filters.jl
mkitti: Maybe we should just switch the alias then. `UnknownFilter` is an alias for `ExternalFilter` ?

musm: Agreed, if this is exported that would be easier and make more sense.

mkitti: Done in [3f5ec32](https://github.com/JuliaIO/HDF5.jl/pull/911/commits/3f5ec321f73322ecd566b1d206a028d82b9ad286)

src/filters/filters.jl
```diff
@@ -248,36 +248,62 @@ function register_filter(::Type{F}) where F <: Filter
...
 
-External filter type. Alias for `UnkownFilter` (see related documentation).
+Unknown filter type. Alias for [`ExternalFilter`](@ref) (see related documentation).
 Intended to support arbitrary, unregistered, external filters. Allows the
 quick creation of filters using internal/proprietary filters without subtyping
 `HDF5.Filters.Filter`.
 Users are instead encouraged to define subtypes on `HDF5.Filters.Filter`.
 """
-const ExternalFilter = UnknownFilter
+const Unknown = ExternalFilter
```
musm, line 306:
```suggestion
const UnknownFilter = ExternalFilter
```

docs/src/filters.md
```diff
@@ -30,6 +30,7 @@ Fletcher32
 Szip
 NBit
 ScaleOffset
+ExternalFilter
 UnknownFilter
```
musm, line 34:
```suggestion
```

src/filters/filters.jl
```diff
@@ -248,36 +248,62 @@ function register_filter(::Type{F}) where F <: Filter
...
 """
-    ExternalFilter
+    UnknownFilter
 
-External filter type. Alias for `UnkownFilter` (see related documentation).
+Unknown filter type. Alias for [`ExternalFilter`](@ref) (see related documentation).
 Intended to support arbitrary, unregistered, external filters. Allows the
 quick creation of filters using internal/proprietary filters without subtyping
 `HDF5.Filters.Filter`.
 Users are instead encouraged to define subtypes on `HDF5.Filters.Filter`.
```
musm, line 304:
```suggestion

```

src/filters/filters.jl
```diff
@@ -248,36 +248,62 @@ function register_filter(::Type{F}) where F <: Filter
 end
 
 """
-    UnknownFilter(filter_id::API.H5Z_filter_t, flags::Cuint, data::Vector{Cuint}, name::String, config::Cuint)
+    ExternalFilter(filter_id::API.H5Z_filter_t, flags::Cuint, data::Vector{Cuint}, name::String, config::Cuint)
+    ExternalFilter(filter_id, flags, data::Integer...)
+    ExternalFilter(filter_id, data::AbstractVector{<:Integer} = Cuint[])
 
-An unknown filter.
+An unknown filter. This is used to encode filters in the pipeline that are not known to HDF5.jl.
```
musm, line 255:
```suggestion
Intended to support arbitrary, unregistered, external filters. Allows the
quick creation of filters using internal/proprietary filters without subtyping
`HDF5.Filters.Filter`.
Users are instead encouraged to define subtypes on `HDF5.Filters.Filter`.
```

musm: Feel free to reword

musm:
Let's try to get this merged in the next week. I think it would be a good idea to tag 0.16.3 early-mid March.

mkitti:
Let's see if https://juliaio.github.io/HDF5.jl/previews/PR911/filters/#Creating-a-new-Filter-type
is still correct.

musm:
Hmm looks like the links are not working still?

mkitti:
I will take a look at this a few hours.

musm:
To remind myself is the only issue left that the HTML docs are not linking things correctly?

mkitti:
Yes, I believe so. I need to revisit some use of `@ref` I think.

mkitti:
I fixed the links. I think this is good to go.

PR #912, opened 2022-03-12
Support `track_order`

t-bltg:
This is a proposition for supporting `track_order` in `write`s and `read`s.

Since `Dict` in julia does not preserve ordering, reading a file with the correct order is a bit tricky and I have to use `OrderedDict`s, ~~and `const  global`s `IDX_TYPE` and `ORDER`~~. Comments for enhancing this `PR` are very welcome.

Related: https://github.com/h5py/h5py/issues/1471.

**write**
```julia
using HDF5

h5open("track_order.h5", "w"; track_order=true) do io
  io["b"] = 1
  io["a"] = 2
  g = create_group(io, "G"; track_order=true)
  write(g, "z", 3)
  write(g, "f", 4)
end
```

_alternatively_
```julia
using OrderedCollections, FileIO

save("track_order.h5", OrderedDict("b"=>1, "a"=>2, "G/z"=>3, "G/f"=>4))
```

**read Julia**
```julia
using OrderedCollections, FileIO

d = load("track_order.h5"; dict=OrderedDict())  # `track_order` is inferred from `OrderedDict`

@show keys(d)
```
_output_
```julia
keys(d) = Any["b", "a", "G/z", "G/f"]
```

(without `track_order`: `Any["G/f", "G/z", "a", "b"]`)

**read Python**
```python
import h5py

with h5py.File("track_order.h5", "r", track_order=True) as f:
    print(f["/"].keys())
    print(f["G"].keys())
```
_output_
```python
<KeysViewHDF5 ['b', 'a', 'G']>
<KeysViewHDF5 ['z', 'f']>
```

### DIFF
--- a/src/fileio.jl
+++ b/src/fileio.jl
@@ -1,6 +1,6 @@
 import .FileIO
 
-function loadtodict!(d::Dict, g::Union{File,Group}, prefix::String="")
+function loadtodict!(d::AbstractDict, g::Union{File,Group}, prefix::String="")
     for k in keys(g)
         v = g[k]
         if v isa Group
--- a/src/fileio.jl
+++ b/src/fileio.jl
@@ -20,11 +20,7 @@ else
     false, nothing
 end
 
-function _restore_track_order(saved)
-    restore, prev = saved
-    restore && (IDX_TYPE[] = prev)
-    nothing
-end
+_restore_track_order(restore, prev) = (restore && (IDX_TYPE[] = prev); nothing)
 
 # load with just a filename returns a flat dictionary containing all the variables
 function fileio_load(f::FileIO.File{FileIO.format"HDF5"}; kwargs...)
@@ -34,7 +30,7 @@ function fileio_load(f::FileIO.File{FileIO.format"HDF5"}; kwargs...)
     out = h5open(FileIO.filename(f), "r"; kw...) do file
         loadtodict!(d, file)
     end
-    _restore_track_order(saved)
+    _restore_track_order(saved...)
     out
 end
 
@@ -44,7 +40,7 @@ function fileio_load(f::FileIO.File{FileIO.format"HDF5"}, varname::AbstractStrin
     out = h5open(FileIO.filename(f), "r"; kwargs...) do file
         read(file, varname)
     end
-    _restore_track_order(saved)
+    _restore_track_order(saved...)
     out
 end
 
@@ -53,14 +49,13 @@ function fileio_load(f::FileIO.File{FileIO.format"HDF5"}, varnames::AbstractStri
     out = h5open(FileIO.filename(f), "r"; kwargs...) do file
         map(var -> read(file, var), varnames)
     end
-    _restore_track_order(saved)
+    _restore_track_order(saved...)
     out
 end
 
 # save all the key-value pairs in the dict as top-level variables
 function fileio_save(f::FileIO.File{FileIO.format"HDF5"}, dict::AbstractDict; kwargs...)
-    saved = _set_track_order(kwargs)
-    out = h5open(FileIO.filename(f), "w"; kwargs...) do file
+    h5open(FileIO.filename(f), "w"; kwargs...) do file
         for (k,v) in dict
             if !isa(k, AbstractString)
                 throw(ArgumentError("keys must be strings (the names of variables), got $k"))
@@ -68,6 +63,4 @@ function fileio_save(f::FileIO.File{FileIO.format"HDF5"}, dict::AbstractDict; kw
             write(file, String(k), v)
         end
     end
-    _restore_track_order(saved)
-    out
 end
--- a/src/fileio.jl
+++ b/src/fileio.jl
@@ -12,29 +12,55 @@ function loadtodict!(d::AbstractDict, g::Union{File,Group}, prefix::String="")
     return d
 end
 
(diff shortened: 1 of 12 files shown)


src/properties.jl
```diff
@@ -253,6 +253,21 @@ function class_setproperty!(::Type{ObjectCreateProperties}, p::Properties, name:
     class_setproperty!(superclass(ObjectCreateProperties), p, name, val)
 end
 
+function get_track_order(p::Properties)
+    attr = Ref{UInt32}()
+    API.h5p_get_attr_creation_order(p, attr)
```
musm, line 258: It would make sense to define in https://github.com/JuliaIO/HDF5.jl/blob/master/src/api/helpers.jl 
```julia
function h5p_get_attr_creation_order(p)
    attr = Ref{UInt32}()
    h5g_get_num_objs(p, attr)
    return attr[]
end
```
same for `h5p_get_link_creation_order`

musm:
If someone passes `OrderedDict`  it sounds like we should automatically be able to set `track_order=true`.

src/properties.jl
t-bltg: Thanks for pointing `src/api/helpers.jl`, I've added the helpers for both.

src/properties.jl
```diff
@@ -253,13 +253,10 @@ function class_setproperty!(::Type{ObjectCreateProperties}, p::Properties, name:
...
-    attr = Ref{UInt32}()
-    API.h5p_get_attr_creation_order(p, attr)
-    link = Ref{UInt32}()
-    API.h5p_get_link_creation_order(p, link)
-    attr[] != 0 && link[] != 0
-end
+get_track_order(p::Properties) = (
+    API.h5p_get_link_creation_order(p) != 0 &&
+    API.h5p_get_attr_creation_order(p) != 0
+)
```
musm, line 259:
```julia
get_track_order(p::Properties) = API.h5p_get_link_creation_order(p) != 0 && API.h5p_get_attr_creation_order(p) != 0
```
If this is going to be a short form function, you might as well make it a one-liner  :)

t-bltg:
> If someone passes OrderedDict it sounds like we should automatically be able to set track_order=true, for the save funciton.

Good point. The problem however, is that `OrderedDict` is not a dependency of `HDF5`, so we cannot test it as in `d isa OrderedDict`.

musm:
> Good point. The problem however, is that OrderedDict is not a dependency of HDF5, so we cannot test it as in d isa OrderedDict.

Ah right...

src/properties.jl
t-bltg: I thought `HDF5.jl` tried to adhere to some kind of `linewidth` code style convention, but as you wish ;)

musm:
We could take advantage of the following trick to support the OrderdDict saving use case:
https://github.com/JuliaPackaging/Requires.jl#multiple-requirements

src/properties.jl
musm: We are not that strict, usually we aim for readability instead of a hardcore line length limit

t-bltg: I would be nitpicking by saying that the multi-line version is more readable (with the `!=` alignment) ;)

t-bltg:
> We could take advantage of the following trick to support the OrderdDict saving use case:
https://github.com/JuliaPackaging/Requires.jl#multiple-requirements

Done. I've updated the tests accordingly.

src/fileio.jl
```diff
@@ -12,30 +13,55 @@ function loadtodict!(d::Dict, g::Union{File,Group}, prefix::String="")
...
     end
+    _restore_track_order(saved...)
+    out
 end
 
 # save all the key-value pairs in the dict as top-level variables
 function fileio_save(f::FileIO.File{FileIO.format"HDF5"}, dict::AbstractDict; kwargs...)
     h5open(FileIO.filename(f), "w"; kwargs...) do file
-        for (k,v) in dict
+        for (k, v) in dict
```
musm, line 64: Shouldn't this also  check
`track_order = isa(d, OrderedDict)`
then pass that into the `h5open` ?

musm: I might be confused on the expected outcomes when an `OrderedDict` is saved.

If someone saves an ordered dict, we do not automatically track its order, but shouldn't  we set it automatically?

If the user then reads this back in as a `Dict` it doesn't matter since `Dict` doesn't preserve order. If the user reads in into an `OrderedDict` and they forgot to set `track_order` then we read it back into an `OrderedDict`, but the ordering is not preserved.  Is that correct?

t-bltg: Done.

musm:
Excellent, latest changes look good to me. I'll give it a day or two to allow others to chime in if they want then merge. Thanks again!

src/fileio.jl
t-bltg: > If someone saves an ordered dict, we do not automatically track its order, but shouldn't we set it automatically?
It is now automatic.

Plus, as a safety measure when someone passes an `OrderedDict`, using `get!(...)` ensures that if the user passes `track_order=false` together with and `OD`, then the order won't be tracked in the corresponding `hdf5` file.

t-bltg:
> I'll give it a day or two to allow others to chime in if they want then merge

Perfect, thanks for helping me out here !

musm:
Thank you! This is nice feature.

PR #914, opened 2022-03-14
clean up `track_order` logic

t-bltg:
I realize that the `kwargs` mutation logic in https://github.com/JuliaIO/HDF5.jl/pull/912 was un-necessary.
This is now much cleaner code IMO. Sorry for writing sloppy code !


musm:
Indeed this is much simpler

PR #917, opened 2022-03-18
Add anonymous datasets

mkitti:
This adds [`HDF5.API.h5d_create_anon`](https://portal.hdfgroup.org/display/HDF5/H5D_CREATE_ANON) to allow for the creation of anonymous datasets.

If `nothing` is passed as the path or name to `create_dataset` or related functions, then an anonymous dataset is created. This mirrors what happens if `None` is passed to h5py: https://docs.h5py.org/en/stable/high/group.html#h5py.Group.create_dataset

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -598,10 +598,10 @@ function create_dataset(
     dapl::DatasetAccessProperties = DatasetAccessProperties(),
     pv...
 )
-    path !== nothing && haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
+    !isnothing(path) && haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
     pv = setproperties!(dcpl,dxpl,dapl; pv...)
     isempty(pv) || error("invalid keyword options")
-    if path === nothing
+    if isnothing(path)
         ds = API.h5d_create_anon(parent, dtype, dspace, dcpl, dapl)
     else
         ds = API.h5d_create(parent, path, dtype, dspace, _link_properties(path), dcpl, dapl)
@@ -1176,7 +1176,7 @@ function readmmap(obj::Dataset)
 end
 
 # Generic write
-function Base.write(parent::Union{File,Group}, name1::Union{AbstractString,Nothing}, val1, name2::Union{AbstractString, Nothing}, val2, nameval...) # FIXME: remove?
+function Base.write(parent::Union{File,Group}, name1::Union{AbstractString,Nothing}, val1, name2::Union{AbstractString,Nothing}, val2, nameval...) # FIXME: remove?
     if !iseven(length(nameval))
         error("name, value arguments must come in pairs")
     end
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -590,7 +590,7 @@ See also
 """
 function create_dataset(
     parent::Union{File,Group},
-    path::Union{AbstractString, Nothing},
+    path::Union{AbstractString,Nothing},
     dtype::Datatype,
     dspace::Dataspace;
     dcpl::DatasetCreateProperties = DatasetCreateProperties(),
@@ -601,16 +601,16 @@ function create_dataset(
     path !== nothing && haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
     pv = setproperties!(dcpl,dxpl,dapl; pv...)
     isempty(pv) || error("invalid keyword options")
-    if path isa Nothing
+    if path === nothing
         ds = API.h5d_create_anon(parent, dtype, dspace, dcpl, dapl)
     else
         ds = API.h5d_create(parent, path, dtype, dspace, _link_properties(path), dcpl, dapl)
     end
     Dataset(ds, file(parent), dxpl)
 end
-create_dataset(parent::Union{File,Group}, path::Union{AbstractString, Nothing}, dtype::Datatype, dspace_dims::Dims; pv...) = create_dataset(checkvalid(parent), path, dtype, dataspace(dspace_dims); pv...)
-create_dataset(parent::Union{File,Group}, path::Union{AbstractString, Nothing}, dtype::Datatype, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, dtype, dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
-create_dataset(parent::Union{File,Group}, path::Union{AbstractString, Nothing}, dtype::Type, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, datatype(dtype), dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
+create_dataset(parent::Union{File,Group}, path::Union{AbstractString,Nothing}, dtype::Datatype, dspace_dims::Dims; pv...) = create_dataset(checkvalid(parent), path, dtype, dataspace(dspace_dims); pv...)
+create_dataset(parent::Union{File,Group}, path::Union{AbstractString,Nothing}, dtype::Datatype, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, dtype, dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
+create_dataset(parent::Union{File,Group}, path::Union{AbstractString,Nothing}, dtype::Type, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, datatype(dtype), dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
 
 # Note that H5Tcreate is very different; H5Tcommit is the analog of these others
 create_datatype(class_id, sz) = Datatype(API.h5t_create(class_id, sz))
@@ -651,7 +651,7 @@ move_link(parent::Union{File,Group}, src_name::AbstractString, dest_name::Abstra
 Base.setindex!(dset::Dataset, val, name::AbstractString) = write_attribute(dset, name, val)
 Base.setindex!(x::Attributes, val, name::AbstractString) = write_attribute(x.parent, name, val)
 # Create a dataset with properties: obj[path, prop = val, ...] = val
-function Base.setindex!(parent::Union{File,Group}, val, path::Union{AbstractString, Nothing}; pv...)
+function Base.setindex!(parent::Union{File,Group}, val, path::Union{AbstractString,Nothing}; pv...)
     need_chunks = any(k in keys(chunked_props) for k in keys(pv))
     have_chunks = any(k == :chunk for k in keys(pv))
 
@@ -1176,7 +1176,7 @@ function readmmap(obj::Dataset)
 end
 
 # Generic write
-function Base.write(parent::Union{File,Group}, name1::Union{AbstractString, Nothing}, val1, name2::Union{AbstractString, Nothing}, val2, nameval...) # FIXME: remove?
+function Base.write(parent::Union{File,Group}, name1::Union{AbstractString,Nothing}, val1, name2::Union{AbstractString, Nothing}, val2, nameval...) # FIXME: remove?
     if !iseven(length(nameval))
         error("name, value arguments must come in pairs")
     end
@@ -1197,7 +1197,7 @@ end
 # Create datasets and attributes with "native" types, but don't write the data.
 # The return syntax is: dset, dtype = create_dataset(parent, name, data; properties...)
 
-function create_dataset(parent::Union{File,Group}, name::Union{AbstractString, Nothing}, data; pv...)
+function create_dataset(parent::Union{File,Group}, name::Union{AbstractString,Nothing}, data; pv...)
     dtype = datatype(data)
     dspace = dataspace(data)
     obj = try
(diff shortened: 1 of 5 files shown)


src/HDF5.jl
```diff
@@ -579,24 +580,37 @@ There are many keyword properties that can be set. Below are a few select keywor
...
+    parent::Union{File,Group},
+    path::Union{AbstractString, Nothing},
+    dtype::Datatype,
+    dspace::Dataspace;
+    dcpl::DatasetCreateProperties = DatasetCreateProperties(),
+    dxpl::DatasetTransferProperties = DatasetTransferProperties(),
+    dapl::DatasetAccessProperties = DatasetAccessProperties(),
+    pv...
+)
+    path !== nothing && haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
```
musm, line 601: worth to test this as well?

mkitti: Could you clarify what you would like to test? I already test passing `nothing` to the function.

The code for checking for an existing dataset already existed before this PR. Is there currently not a test for this?

mkitti: I think we already test the error for an existing dataset:
https://github.com/JuliaIO/HDF5.jl/blob/704d9e8c32b3b5d1239254d0155d7e14cad0c713/test/plain.jl#L1362-L1367

musm: Nevermind. When I read this, I thought you added this whole line, but that's not right heh

src/HDF5.jl
```diff
@@ -589,22 +590,27 @@ See also
...
     dxpl::DatasetTransferProperties = DatasetTransferProperties(),
     dapl::DatasetAccessProperties = DatasetAccessProperties(),
     pv...
 )
-    haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
+    path !== nothing && haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
     pv = setproperties!(dcpl,dxpl,dapl; pv...)
     isempty(pv) || error("invalid keyword options")
-    Dataset(API.h5d_create(parent, path, dtype, dspace, _link_properties(path), dcpl, dapl), file(parent), dxpl)
+    if path isa Nothing
```
mkitti, line 604:
```suggestion
    if path === nothing
```

Perhaps this is clearer? I'm not sure what the practical difference is.

musm: `isnothing(path)`

musm, line 604:
```suggestion
    if isnothing(path)
```
would be more canonical

musm:
Interesting feature, didn't know this was possible. Apparently the python docs suggest this can be helpful for linking later. I still kind of wonder what kind of use case that would assist.

src/HDF5.jl
```diff
@@ -589,22 +590,27 @@ See also
 """
 function create_dataset(
     parent::Union{File,Group},
-    path::AbstractString,
+    path::Union{AbstractString, Nothing},
```
musm, line 593:
```suggestion
    path::Union{AbstractString,Nothing},
```

src/HDF5.jl
```diff
@@ -589,22 +590,27 @@ See also
 """
 function create_dataset(
```
musm, line 591: Technically this is a new convention on styling the breaking long method signatures in the repo, but it's not that important.  We usually would align the second lines, similar to the majority of the julialang repo (I think).

src/HDF5.jl
mkitti: `isnothing` does not exist in older Julia whereas `===` does.

src/HDF5.jl
mkitti: I'm using https://github.com/invenia/BlueStyle#method-definitions as a style reference. In particular, I'm using the format they have here:

```julia
# Yes:
function foobar(
    df::DataFrame,
    id::Symbol,
    variable::Symbol,
    value::AbstractString,
    prefix::AbstractString="",
)
    # code
end
```

I'm not sure if anyone else has tried to codify a style for Julia. `JuliaLang/Julia` is not very consistent stylistically, but I think Blue extracts the best practices.

src/HDF5.jl
mkitti: https://docs.julialang.org/en/v1/base/base/#Base.isnothing is only available with Julia 1.1. Have we broken compat with Julia 1.0?

src/HDF5.jl
```diff
@@ -589,22 +590,27 @@ See also
...
-create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::Datatype, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, dtype, dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
-create_dataset(parent::Union{File,Group}, path::AbstractString, dtype::Type, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, datatype(dtype), dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
+create_dataset(parent::Union{File,Group}, path::Union{AbstractString, Nothing}, dtype::Datatype, dspace_dims::Dims; pv...) = create_dataset(checkvalid(parent), path, dtype, dataspace(dspace_dims); pv...)
+create_dataset(parent::Union{File,Group}, path::Union{AbstractString, Nothing}, dtype::Datatype, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, dtype, dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
+create_dataset(parent::Union{File,Group}, path::Union{AbstractString, Nothing}, dtype::Type, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, datatype(dtype), dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
```
mkitti, line 613:
```suggestion
create_dataset(parent::Union{File,Group}, path::Union{AbstractString,Nothing}, dtype::Datatype, dspace_dims::Dims; pv...) = create_dataset(checkvalid(parent), path, dtype, dataspace(dspace_dims); pv...)
create_dataset(parent::Union{File,Group}, path::Union{AbstractString,Nothing}, dtype::Datatype, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, dtype, dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
create_dataset(parent::Union{File,Group}, path::Union{AbstractString,Nothing}, dtype::Type, dspace_dims::Tuple{Dims,Dims}; pv...) = create_dataset(checkvalid(parent), path, datatype(dtype), dataspace(dspace_dims[1], max_dims=dspace_dims[2]); pv...)
```

src/HDF5.jl
```diff
@@ -645,7 +651,7 @@ move_link(parent::Union{File,Group}, src_name::AbstractString, dest_name::Abstra
 Base.setindex!(dset::Dataset, val, name::AbstractString) = write_attribute(dset, name, val)
 Base.setindex!(x::Attributes, val, name::AbstractString) = write_attribute(x.parent, name, val)
 # Create a dataset with properties: obj[path, prop = val, ...] = val
-function Base.setindex!(parent::Union{File,Group}, val, path::AbstractString; pv...)
+function Base.setindex!(parent::Union{File,Group}, val, path::Union{AbstractString, Nothing}; pv...)
```
mkitti, line 654:
```suggestion
function Base.setindex!(parent::Union{File,Group}, val, path::Union{AbstractString,Nothing}; pv...)
```

src/HDF5.jl
```diff
@@ -1170,7 +1176,7 @@ function readmmap(obj::Dataset)
 end
 
 # Generic write
-function Base.write(parent::Union{File,Group}, name1::AbstractString, val1, name2::AbstractString, val2, nameval...) # FIXME: remove?
+function Base.write(parent::Union{File,Group}, name1::Union{AbstractString, Nothing}, val1, name2::Union{AbstractString, Nothing}, val2, nameval...) # FIXME: remove?
```
mkitti, line 1179:
```suggestion
function Base.write(parent::Union{File,Group}, name1::Union{AbstractString,Nothing}, val1, name2::Union{AbstractString, Nothing}, val2, nameval...) # FIXME: remove?
```

src/HDF5.jl
```diff
@@ -1191,7 +1197,7 @@ end
 # Create datasets and attributes with "native" types, but don't write the data.
 # The return syntax is: dset, dtype = create_dataset(parent, name, data; properties...)
 
-function create_dataset(parent::Union{File,Group}, name::AbstractString, data; pv...)
+function create_dataset(parent::Union{File,Group}, name::Union{AbstractString, Nothing}, data; pv...)
```
mkitti, line 1200:
```suggestion
function create_dataset(parent::Union{File,Group}, name::Union{AbstractString,Nothing}, data; pv...)
```

src/HDF5.jl
```diff
@@ -1213,7 +1219,7 @@ function create_attribute(parent::Union{File,Object}, name::AbstractString, data
 end
 
 # Create and write, closing the objects upon exit
-function write_dataset(parent::Union{File,Group}, name::AbstractString, data; pv...)
+function write_dataset(parent::Union{File,Group}, name::Union{AbstractString, Nothing}, data; pv...)
```
mkitti, line 1222:
```suggestion
function write_dataset(parent::Union{File,Group}, name::Union{AbstractString,Nothing}, data; pv...)
```

src/HDF5.jl
```diff
@@ -1562,7 +1568,7 @@ function write_dataset(dataset::Dataset, memtype::Datatype, buf::AbstractArray,
     stride(buf, 1) != 1 && throw(ArgumentError("Cannot write arrays with a different stride than `Array`"))
     API.h5d_write(dataset, memtype, API.H5S_ALL, API.H5S_ALL, xfer, buf)
 end
-function write_dataset(dataset::Dataset, memtype::Datatype, str::AbstractString, xfer::DatasetTransferProperties=dataset.xfer)
+function write_dataset(dataset::Dataset, memtype::Datatype, str::Union{AbstractString, Nothing}, xfer::DatasetTransferProperties=dataset.xfer)
```
mkitti, line 1571:
```suggestion
function write_dataset(dataset::Dataset, memtype::Datatype, str::Union{AbstractString,Nothing}, xfer::DatasetTransferProperties=dataset.xfer)
```

src/HDF5.jl
mkitti: I'm not sure where to increase the ident to. Identing it more seems to contradict this part of the Blue style guide.
```julia
# No: Indented too much.
function foobar(
        df::DataFrame,
        id::Symbol,
        variable::Symbol,
        value::AbstractString,
        prefix::AbstractString="",
    )
    # code
end
```

src/HDF5.jl
```diff
@@ -1170,7 +1176,7 @@ function readmmap(obj::Dataset)
 end
 
 # Generic write
-function Base.write(parent::Union{File,Group}, name1::AbstractString, val1, name2::AbstractString, val2, nameval...) # FIXME: remove?
+function Base.write(parent::Union{File,Group}, name1::Union{AbstractString,Nothing}, val1, name2::Union{AbstractString, Nothing}, val2, nameval...) # FIXME: remove?
```
mkitti, line 1179:
```suggestion
function Base.write(parent::Union{File,Group}, name1::Union{AbstractString,Nothing}, val1, name2::Union{AbstractString,Nothing}, val2, nameval...) # FIXME: remove?
```

src/HDF5.jl
musm: https://github.com/JuliaIO/HDF5.jl/blob/master/Project.toml#L17

musm: Yes

src/HDF5.jl
musm: We don't try to follow the Blue style guide and neither does the main Julia repo. So I wouldn't feel constrained by that guide. After all, it's just one convection.

mkitti: Feel free to suggest an indentation.

musm: I suggest we don't do anything and keep it as is for now.

src/HDF5.jl
```diff
@@ -589,22 +590,27 @@ See also
...
+    path::Union{AbstractString,Nothing},
     dtype::Datatype,
     dspace::Dataspace;
     dcpl::DatasetCreateProperties = DatasetCreateProperties(),
     dxpl::DatasetTransferProperties = DatasetTransferProperties(),
     dapl::DatasetAccessProperties = DatasetAccessProperties(),
     pv...
 )
-    haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
+    path !== nothing && haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
```
musm, line 601:
```suggestion
    !isnothing(path) && haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
```

src/HDF5.jl
```diff
@@ -589,22 +590,27 @@ See also
...
     pv = setproperties!(dcpl,dxpl,dapl; pv...)
     isempty(pv) || error("invalid keyword options")
-    Dataset(API.h5d_create(parent, path, dtype, dspace, _link_properties(path), dcpl, dapl), file(parent), dxpl)
+    if path === nothing
```
musm, line 604:
```suggestion
    if isnothing(path)
```

musm (approved):

src/HDF5.jl
mkitti, line 601:
```suggestion
    isnothing(path) || haskey(parent, path) && error("cannot create dataset: object \"", path, "\" already exists at ", name(parent))
```

src/HDF5.jl
mkitti, line 604:
```suggestion
    if isnothing(path)
```

mkitti (commented):
Use `isnothing`.

mkitti:
In my application I'm using this to create a binary template for LabView to fill in with data. The anonymous datasets allows me to space out the real datasets at regular intervals so it is easy to fill in. This way LabView just knows that needs to know to write dataset at regular offsets and it can "create" a HDF5 file without using the HDF5 library.

PR #918, opened 2022-03-19
HDF5 Type order precision offset

mkitti:
This pull request expands the options to create types with different orders (endianness), precision, and offsets.

<strike>It also changes how unknown types are printed.</strike>

```julia
using HDF5

function create_h5_uint24()
    dt = HDF5.API.h5t_copy(HDF5.API.H5T_STD_U32LE)
    HDF5.API.h5t_set_size(dt, 3)
    HDF5.API.h5t_set_precision(dt, 24)
    return HDF5.Datatype(dt)
end
```

### DIFF
--- a/src/api/types.jl
+++ b/src/api/types.jl
@@ -304,6 +304,7 @@ const H5S_SEL_HYPERSLABS = 2
 const H5S_SEL_ALL        = 3
 
 # type classes (C enum H5T_class_t)
+const H5T_NO_CLASS     = hid_t(-1)
 const H5T_INTEGER      = hid_t(0)
 const H5T_FLOAT        = hid_t(1)
 const H5T_TIME         = hid_t(2)  # not supported by HDF5 library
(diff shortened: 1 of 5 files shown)


src/api/types.jl
```diff
@@ -316,13 +317,29 @@ const H5T_ENUM         = hid_t(8)
...
+
 # Character types
 const H5T_CSET_ASCII   = 0
 const H5T_CSET_UTF8    = 1
 
 # Sign types (C enum H5T_sign_t)
+const H5T_SGN_ERROR = Cint(-1) # error
 const H5T_SGN_NONE     = Cint(0)  # unsigned
 const H5T_SGN_2        = Cint(1)  # 2's complement
+const H5T_NSGN = Cint(2)        # sentinel: this must be last!
```
musm, line 342:
```suggestion
const H5T_NSGN         = Cint(2)        # sentinel: this must be last!
```

mkitti:
I originally wrote this a while ago, forgot about, and then discovered it was missing when I need it again.

I'm not quite happy with the type display.

src/show.jl
```diff
@@ -68,6 +68,9 @@ function Base.show(io::IO, dtype::Datatype)
     if isvalid(dtype)
         API.h5t_committed(dtype) && print(io, name(dtype), " ")
         print(io, API.h5lt_dtype_to_text(dtype))
```
musm, line 70: I'd remove this `print` statement and combine it with the latter one.

mkitti:
I am withdrawing the changes to how types are printed for now. I'll revisit that in another pull request if needed.

PR #920, opened 2022-03-23
Add Core driver and other driver API

mkitti:
This creates the core driver which allows HDF5 files to be created in memory.


mkitti:
Interesting... CI is not running

src/drivers/drivers.jl
```diff
@@ -20,6 +20,50 @@ end
...
+        r_backing_store[],
+        r_write_tracking[],
+        r_page_size[]
+    )
+end
+
+function set_driver!(p::Properties, driver::Core)
+    API.h5p_set_fapl_core(p, driver.increment, driver.backing_store)
+    API.h5p_set_core_write_tracking(p, driver.write_tracking, driver.page_size)
+end
```
musm, line 66:
```suggestion
    return nothing
end
```

mkitti: I'm fine with this. Is the general convention to return `nothing` from the property setters? The other thing I could see returning is the the `Properties` instance, `p`?

musm: > I'm fine with this. Is the general convention to return nothing from the property setters? The other thing I could see returning is the the Properties instance, p?

I don't think we are using the properties instance. I guess it might be useful, but I think it makes sense to just return `nothing` to keep things simple.

mkitti: In one of the other pull requests, I kept finding myself doing the following to make sure the set worked. I wonder if this is a more common pattern?

```
julia> HDF5.set_fill_time!(dcpl, :alloc)

julia> dcpl
HDF5.DatasetCreateProperties(
  alloc_time      = :late,
  fill_time       = :alloc,
  fill_value      = 3.0,
  filters         = HDF5.Filters.Filter[],
  layout          = :contiguous,
  filter          = HDF5.Filters.Filter[],
  obj_track_times = true,
)

julia> HDF5.set_fill_time!(dcpl, :never)

julia> dcpl
HDF5.DatasetCreateProperties(
  alloc_time      = :late,
  fill_time       = :never,
  fill_value      = 3.0,
  filters         = HDF5.Filters.Filter[],
  layout          = :contiguous,
  filter          = HDF5.Filters.Filter[],
  obj_track_times = true,
)
```

musm:
LGTM

PR #921, opened 2022-03-23
Add fill_time and fill_value dataset props

mkitti:
This adds dataset creation properties for `fill_time` and `fill_value`.


src/api_midlevel.jl
```diff
@@ -185,3 +185,17 @@ function write_chunk(dataset_id, index::Integer, buf::Union{DenseArray,Base.Fast
     offset = [reverse(get_chunk_offset(dataset_id, index))...]
     write_chunk(dataset_id, offset, buf; dxpl_id = dxpl_id, filter_mask = filter_mask)
 end
+
+function get_fill_value(plist_id, ::Type{T}) where T
```
musm, line 189: Would the second argument ever not be `Float64`?

mkitti: One might want an `Int64` at some point.

mkitti: I defaulted to `Float64` since that should cover the most common cases.

musm: Ok I see that that is the case in the test examples. And it doesn't seem like something we could infer.

mkitti: Right. I cannot find a way to find the type unless we have a reference to the associated dataset.

musm:
LGTM

mkitti:
What's with CI these days?

mkitti:
thanks

musm:
At some point we should make an announcement on discourse with all these new v0.16 improvements 😄

mkitti:
I'm thinking of at least one more on file free space strategy.
https://docs.hdfgroup.org/hdf5/rfc/FileSpaceManagement.pdf
https://docs.hdfgroup.org/archive/support/HDF5/docNewFeatures/NewFeaturesFileSpaceMgmtDocs.html
https://docs.hdfgroup.org/hdf5/develop/group___f_c_p_l.html#ga167ff65f392ca3b7f1933b1cee1b9f70

PR #923, opened 2022-03-25
Update index.md

rkube:
Libraries are either in `lib` or `lib64`, see https://github.com/JuliaIO/HDF5.jl/blob/master/deps/build.jl

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -27,7 +27,7 @@ Starting from Julia 1.3, the HDF5 binaries are by default downloaded using the `
 
 ### Using custom or system provided HDF5 binaries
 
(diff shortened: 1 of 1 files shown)


mkitti:
I concur.
https://github.com/JuliaIO/HDF5.jl/blob/d8287f201b10885a2d06847cd241899690dde062/deps/build.jl#L19-L20

mkitti (approved):

musm:
Thanks for the fix. Is it  in `lib` on 32-bit systems?

musm:
This answered my question: https://unix.stackexchange.com/questions/74646/difference-between-lib-lib32-lib64-libx32-and-libexec

PR #924, opened 2022-03-26
Add filespace management API calls

mkitti:
This is a work in progress.


musm:
LGMT if you want to merge this as is.

src/api/helpers.jl
```diff
@@ -224,6 +224,35 @@ function h5f_get_vfd_handle(file_id, fapl)
...
+function h5f_get_free_sections!(file_id, type, sect_info::Vector{H5F_sect_info_t})
+    h5f_get_free_sections(file_id, type, length(sect_info), sect_info)
+    return sect_info
+end
+
+function h5p_get_file_space_strategy(plist_id)
+    strategy = Ref{H5F_fspace_strategy_t}()
+    persist = Ref{hbool_t}()
+    threshold = Ref{hsize_t}()
+    h5p_get_file_space_strategy(plist_id)
```
musm, line 246: Looks like you missed passing in the refs to the method call?

src/api/helpers.jl
```diff
@@ -253,6 +253,20 @@ function h5p_get_file_space_page_size(plist_id)
...
 end
 
+function h5p_set_file_space_strategy(plist_id; strategy = nothing, persist = nothing, threshold = nothing)
+    current = h5p_get_file_space_strategy(plist_id)
+    params = (
+        strategy = isnothing(strategy) ? current[:strategy] : strategy,
+        persist = isnothing(persist) ? current[:persist] : persist,
+        threshold = isnothing(threshold) ? current[:threshold] : threshold
+    )
+    return h5p_set_file_space_strategy(plist_id, params)
```
musm, line 263: I think it would be simpler and just as effective to simply pass them in:
```
h5p_set_file_space_strategy(plist_id, strategy, persist, threshold)
```
As opposed to adding the additional method below, which I don't think is going to be used  elsewhere.

musm: That is true, although, if we really wanted to expose this  extended helper method, we would do it outside the `helpers.jl` file in the midlevel layer and rename the function; the methods in this  file are really supposed to be thin  overlays to the low-level api calls and really  intended for internal or low-level calls, so I don't see a huge benefit to the addition.  

Unless there's a direct need, I think it would be  better to keep things simple and only add additional method extensions as we find a direct need.

src/api/helpers.jl
```diff
@@ -224,6 +224,43 @@ function h5f_get_vfd_handle(file_id, fapl)
...
+    nsects = h5f_get_free_sections(file_id, type, 0, C_NULL)
+    return h5f_get_free_sections(file_id, type, nsects)
+end
+
+function h5f_get_free_sections(file_id, type, nsects)
+    sect_info = Vector{H5F_sect_info_t}(undef, nsects)
+    return h5f_get_free_sections!(file_id, type, sect_info)
+end
+
+function h5f_get_free_sections!(file_id, type, sect_info::Vector{H5F_sect_info_t})
```
mkitti, line 237: Maybe this should not have a `!`

mkitti:
@musm This is ready to go. I'm just not sure if `h5f_get_free_sections!` should exist or if it should have a `!`.

src/api/helpers.jl
musm: I'd combine them
```julia
function h5f_get_free_sections(file_id, type, nsects)
    sect_info = Vector{H5F_sect_info_t}(undef, nsects)
    h5f_get_free_sections(file_id, type, length(sect_info), sect_info)
    return sect_info
end
```

src/api/helpers.jl
```diff
@@ -224,6 +224,43 @@ function h5f_get_vfd_handle(file_id, fapl)
     return file_handle[]
 end
 
+function h5f_get_free_sections(file_id, type)
+    nsects = h5f_get_free_sections(file_id, type, 0, C_NULL)
+    if nsects > 0
+        return h5f_get_free_sections(file_id, type, nsects)
+    else
+        return H5F_sect_info_t[]
```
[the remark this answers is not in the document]
mkitti: It returns a `Vector` of `H5F_sect_info_t` describing the address and size of free sections in the HDF5 file. The first call retrieves the number of free sections. If there are no free sections, it is an error to request 0 free sections with a non-null pointer. Therfore we have to return an empty `Vector` of `H5F_sect_info_t` if `nsects == 0`.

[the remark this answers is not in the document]
musm: Ahh, that makes sense. Thanks for the clarification.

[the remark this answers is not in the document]
musm: I missed the definition below and was looking at the original low-level call.... Is there a reason we are separating out these two methods?  Can we just combine them together again?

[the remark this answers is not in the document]
mkitti: We could, but I think this more obviously shows that there are two and three argument versions of this method without having to consider a union type of `Nothing` and `Int`. I think this is simpler.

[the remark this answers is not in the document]
musm: How would the union show up?  It looks like it always returns an `H5F_sect_info_t`
It seems like the main use case for `nsects !== h5f_get_free_sections(file_id, type, 0, C_NULL)` is to return fewer entries of the vector.  Is that helpful in some cases? 

I'm looking over the documentation: https://confluence.hdfgroup.org/display/HDF5/H5F_GET_FREE_SECTIONS
It also looks like that for the method on L236. If you pass in nsects+1  where `nsects = h5f_get_free_sections(file_id, type, 0, C_NULL)`  It will only fill in the first `nsects` entries of the `sect_info` vector. Might have to add a check to ensure against this.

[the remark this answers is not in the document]
mkitti: Yes, it is useful if you would like to avoid allocating new memory and would like to reuse an existing vector rather than creating a new one.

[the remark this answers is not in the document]
mkitti: or rather in this specific case, if you already know how many free sections there from a previous call.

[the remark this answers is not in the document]
musm: > The third example shows that the first call to H5Fget_free_sections() returns the total number of free-space sections in nsects for the H5FD_MEM_BTREE free-space manager in the file that is associated with fcpl. Even though the second call to H5Fget_free_sections() requests nsects+1sections, the routine retrieves free-space section information in sect_info for only nsects sections. The value in ret is the same as nsects.

I think we need to check `nsects <= ` the nominal number by throwing an argument error or resizing the array. Otherwise  it'll return the undef entries.

src/api/helpers.jl
```diff
@@ -224,6 +224,49 @@ function h5f_get_vfd_handle(file_id, fapl)
...
+If `sect_info` is provided, a view, a `SubArray`, will be returned.
+"""
+function h5f_get_free_sections(file_id, type)
+    nsects = h5f_get_free_sections(file_id, type, 0, C_NULL)
+    sect_info = Vector{H5F_sect_info_t}(undef, nsects)
+    if nsects > 0
+        h5f_get_free_sections(file_id, type, nsects, sect_info)
+    end
+    return sect_info
+end
```
mkitti, line 240: I revised this so that the 2-arg version will return a `Vector`.

src/api/helpers.jl
```diff
@@ -224,6 +224,49 @@ function h5f_get_vfd_handle(file_id, fapl)
...
+
+function h5f_get_free_sections(file_id, type, sect_info::AbstractVector{H5F_sect_info_t})
+    nsects = length(sect_info)
+    nsects = h5f_get_free_sections(file_id, type, nsects, sect_info)
+    return @view(sect_info[1:nsects])
+end
```
mkitti, line 246: The 3-arg version takes an `AbstractVector` as the third argument and will return a view, a `SubArray`.

src/api/helpers.jl
```diff
@@ -224,19 +224,25 @@ function h5f_get_vfd_handle(file_id, fapl)
...
+    return sect_info
 end
 
-function h5f_get_free_sections(file_id, type, nsects)
-    sect_info = Vector{H5F_sect_info_t}(undef, nsects)
-    h5f_get_free_sections(file_id, type, nsects, sect_info)
-    return sect_info
+function h5f_get_free_sections(file_id, type, sect_info::AbstractVector{H5F_sect_info_t})
+    nsects = length(sect_info)
+    nsects = h5f_get_free_sections(file_id, type, nsects, sect_info)
```
musm, line 244: Does this method always return the same `nsects` as passed in?

mkitti: No. See your post above:
https://github.com/JuliaIO/HDF5.jl/pull/924#discussion_r839846588

src/api/helpers.jl
musm: Ok let me get this straight. If a user passes in a vector of length `n` where `n > nsects` this will fill in those entries just fine. Now if a user passes in a vector with `n< nsects` this will still work as only those `n` entries will be filled.  I think that's right?

mkitti: Correct

musm: Ok great, thanks for changing this and the other helper method.

musm: Makes sense to me and is closer with the hdf5 library signature which this file aims to simplify for the Julia side. So I think they're both positive changes.

PR #926, opened 2022-03-30
Resize h5p_get_external if no null is found

musm:


src/api/helpers.jl
```diff
@@ -361,18 +361,26 @@ function h5p_get_efile_prefix(plist)
...
+            name_size *= 2
+            name = Base.StringVector(name_size)
+        else
+            resize!(name, null_id-1)
+            break
+        end
+    end
     @static if Sys.iswindows() && sizeof(Int) == 4
         sz[] &= 0xffffffff
     end
```
musm, line 382: @mkitti  could you explain why this is necessary?

src/api/helpers.jl
```diff
@@ -361,18 +361,26 @@ function h5p_get_efile_prefix(plist)
...
     sz = Ref{hsize_t}(0)
-    h5p_get_external(plist, idx, name_size, name, offset, sz)
-    # name may not be null terminated according to H5P_GET_EXTERNAL documentation
-    nul_idx =  findfirst(==(0x00), name)
-    name_size = nul_idx === nothing ? name_size : nul_idx-1
+    name_size = 64
+    name = Base.StringVector(name_size)
+    null_id = nothing
+    while true
+        h5p_get_external(plist, idx, name_size+1, name, offset, sz)
```
musm, line 370: There might be an alternative way to get the name through a different hdf5 method. 

For example, that's what we did here:
https://github.com/JuliaIO/HDF5.jl/blob/master/src/api/helpers.jl#L591

src/api/helpers.jl
mkitti: If we take that out, I think this fails on 32-bit Windows. I'm guessing that one of our types is coded wrong for 32-bit windows.

src/api/helpers.jl
```diff
@@ -361,18 +361,25 @@ function h5p_get_efile_prefix(plist)
...
-    nul_idx =  findfirst(==(0x00), name)
-    name_size = nul_idx === nothing ? name_size : nul_idx-1
+    name_size = 64
+    name = Base.StringVector(name_size)
+    while true
+        h5p_get_external(plist, idx, name_size+1, name, offset, sz)
+        null_id = findfirst(==(0x00), name)
+        if isnothing(null_id)
+            name_size *= 2
+            name = Base.StringVector(name_size)
```
mkitti, line 373:
```suggestion
            resize!(name, name_size)
```

Why allocate

musm: Yep, that was dumb. Thank you.

src/api/helpers.jl
```diff
@@ -361,18 +361,25 @@ function h5p_get_efile_prefix(plist)
 end
 
 function h5p_get_external(plist, idx = 0)
-    name_size = Csize_t(1024)
-    name = zeros(UInt8, name_size)
-    offset = Ref{off_t}(0)
+    offset = Ref{Int64}(0)
```
mkitti, line 364: I think this should still be `off_t` following the C header:
https://portal.hdfgroup.org/display/HDF5/H5P_GET_EXTERNAL

musm: whops, thanks, I was locally testing and forgot to put it back in.

musm, line 364:
```suggestion
    offset = Ref{off_t}(0)
```

src/api/helpers.jl
mkitti: Why `name_size+1`? The library is going to overrun the buffer.

mkitti:
```suggestion
        h5p_get_external(plist, idx, name_size, name, offset, sz)
```

mkitti:
```suggestion
        h5p_get_external(plist, idx, name_size, name, offset, sz)
```

musm: Stringvector includes an implied null

musm: It might be - not +, but I'll check later.

musm: The size passed in the library needs to include a slot for the null byte. Whereas when you use string vector the length does not include the null byte (it's implied). This is why we need to pass in `name_size+1`.

musm: Let me double check that for this method, the above even matters since it's not returning the required buffer size for this hdf5 method.

musm: Ok right this case it doesn't actually matter since the library wants the full array length.

src/api/helpers.jl
musm: Oddly enough, however, the type is  hsize_t here: https://github.com/HDFGroup/hdf5/blob/78fe6751b5a682111e8bc2bf7ce944c55d52ba56/src/H5Pdcpl.c#L2974

musm: I'm thinking to change the API signature to UInt, instead of this hack.

mkitti: Is our `hsize_t` definition wrong?
https://github.com/JuliaIO/HDF5.jl/blob/ad2f86d38d0298de7e187d7e09bb26e4312fb562/src/api/types.jl#L9

musm: From the source code of libhdf5 that is the correct definition. It's a `typedef unsigned long long`

It's just that this method `h5p_get_external ` has issues on win32 with  `sz`  being UInt64 on a 64-bit windows. Changing 
```
@bind h5p_get_external(plist::hid_t, idx::Cuint, name_size::Csize_t, name::Ptr{Cuchar}, offset::Ptr{off_t}, size::Ptr{hsize_t})::herr_t "Error getting external file properties"
```
to
```
@bind h5p_get_external(plist::hid_t, idx::Cuint, name_size::Csize_t, name::Ptr{Cuchar}, offset::Ptr{off_t}, size::Ptr{UInt})::herr_t "Error getting external file properties"
```
fixes the problem. Despite the fact that the libhdf5 is using `hsize_t`  😕

mkitti: This is probably a bug in HDF5 on Win32. Frankly, it still looks like a hack to me because I think 32-bit windows is the only 32-bit platform we test on.

mkitti: Well your hack looks slightly better than my hack, so might as well commit it.

musm: I'd like to double check on 32-bit Linux, I think we have the capability to test it, but I probably disabled it due to redundancy with the win32 system.

mkitti: We don't have a HDF5_jll to test it on.

musm: welp....

In any case, I don't think it hurts too much to use `UInt` for this entry anyways... The external file data would have to bigger than `UInt32` for this to be a problem...

mkitti: Did we find a configuration that will pass CI?

musm: I think so, I just haven't gotten around to changing the PR.

src/api/helpers.jl
```diff
@@ -361,18 +361,22 @@ function h5p_get_efile_prefix(plist)
...
-    nul_idx =  findfirst(==(0x00), name)
-    name_size = nul_idx === nothing ? name_size : nul_idx-1
-    @static if Sys.iswindows() && sizeof(Int) == 4
-        sz[] &= 0xffffffff
+    sz = Ref{UInt}(0)
+    name_size = 64
+    name = Base.StringVector(name_size)
+    while true
+        h5p_get_external(plist, idx, name_size, name, offset, sz)
+        null_id = findfirst(==(0x00), name)
```
mkitti, line 370: Is an infinite loop possible here if `findfirst` keeps returning `nothing`?

mkitti:
Are we still stuck here?

musm:
I'll rebase and merge. Sadly, I wasn't able to get this to work without having to add
```
    @static if Sys.iswindows() && sizeof(Int) == 4
        sz[] &= 0xffffffff
    end
```

mkitti:
Frankly, I don't think we should spend too much time worrying about 32-bit Windows.

PR #928, opened 2022-04-21
Initialise drivers at module load time

jipolanco:
This PR fixes an issue I experienced when using HDF5 + MPI which was probably introduced by #920.

In short, when MPI.jl is loaded, the `MPIO` driver *replaces* the `Core` driver in the `Drivers.DRIVERS` dictionary, instead of just adding itself to the list. In other words, as soon as one loads MPI.jl, the `Core` driver is no longer available.

As I understand it, this happens because the `DRIVERS` dictionary is currently filled at *precompilation* time with the serial drivers (`Core` and `POSIX`). That is, functions like `HDF5.API.h5fd_core_init()` are called at precompilation time, when they should be called after the HDF5 C library is initialised (i.e. in `__init__`). Note that this is the case for the `MPIO` driver, which is initialised at module load time if MPI.jl has been loaded. In this case, the C library has not yet initialised the serial drivers, so it is free to choose driver ids that may conflict with those in the `DRIVERS` dictionary.

Illustration (using MPI-enabled HDF5 libraries):

```julia
julia> using HDF5

julia> Drivers.DRIVERS
Dict{Int64, Any} with 2 entries:
  576460752303423489 => Core
  576460752303423488 => POSIX

julia> using MPI

julia> Drivers.DRIVERS
Dict{Int64, Any} with 2 entries:
  576460752303423489 => MPIO     # this driver has **replaced** the Core driver
  576460752303423488 => POSIX
```

In practice, this is annoying if one opens a file in serial mode:

```julia
julia> using MPI

julia> using HDF5

julia> ff = h5open(tempname(), "w")
🗂️ HDF5.File: (read-write) /tmp/jl_NscWgElJxE

julia> fapl = HDF5.get_access_properties(ff)
HDF5.FileAccessProperties(  # NOTE: the driver is not among the properties!
  alignment       = (0x0000000000000001, 0x0000000000000001),
  fclose_degree   = :strong,
  libver_bounds   = (:earliest, v"1.10.0"),
)

julia> Drivers.get_driver(fapl)
ERROR: HDF5.API.H5Error: Error in h5p_get_fapl_core (not annotated)
libhdf5 Stacktrace:
 [1] H5Pget_fapl_core: Property lists/Bad value
     incorrect VFL driver
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/HDF5/wzCtc/src/api/error.jl:18 [inlined]
 [2] h5p_get_fapl_core(fapl_id::HDF5.FileAccessProperties, increment::Base.RefValue{UInt64}, backing_store::Base.RefValue{UInt32})
   @ HDF5.API ~/.julia/packages/HDF5/wzCtc/src/api/functions.jl:1348
 [3] get_driver(p::HDF5.FileAccessProperties, #unused#::Type{HDF5.Drivers.Core})
   @ HDF5.Drivers ~/.julia/packages/HDF5/wzCtc/src/drivers/drivers.jl:55
 [4] get_driver(p::HDF5.FileAccessProperties)
   @ HDF5.Drivers ~/.julia/packages/HDF5/wzCtc/src/drivers/drivers.jl:18
 [5] top-level scope
   @ REPL[5]:1
```

(For some reason, things work correctly if one inverts the loading order of MPI and HDF5.)


mkitti:
Looks good to me.

Previously, `DRIVERS` was getting populated during precompilation. Doing this during `__init__` instead makes sense since it depends on runtime.

mkitti (approved):

musm:
New release includes this patch fix. Thank you!

PR #934, opened 2022-05-08
Add create_dataset convenience forms for dataspace

mkitti:
This pull requests adds some convenience forms for `create_dataset` when specifying a dataspace. It allows the dimensions to specified as a single tuple or as `Int`s. For example, 

```julia
    h5open(fn, "w") do f
        create_dataset(f, "test", Int, (128, 32))
        create_dataset(f, "test2", Float64, 128, 64)
    end
```


musm:
Makes sense

PR #935, opened 2022-05-09
Avoid possible segfault when loading system HDF5

### LINKED ISSUE #929: Segmentation fault in building HDF5 in project
Recently, my HDF5 fails to build while running `build` from the Julia package manager. I do not know how to solve this problem, but I get the following error message:

```
    Building HDF5 ────→ `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/36df177c1ce5f399a8de959e5f4b75216fe6c834/build.log`
ERROR: Error building `HDF5`: 

signal (11): Segmentation fault: 11
in expression starting at /Users/chiel/.julia/packages/HDF5/THb0i/deps/build.jl:8

signal (4): Illegal instruction: 4
in expression starting at /Users/chiel/.julia/packages/HDF5/THb0i/deps/build.jl:8
```

On line 8 of `build.jl` is the statement
```
new_contents = if libpath === nothing
```

I don't know whether the problem is on the HDF5 side or mine, but I am stuck with this for a few weeks now and have no idea how to solve it. I use a system provided MPI-enabled HDF5. My project is [MicroHH.jl](https://github.com/Chiil/MicroHH.jl)

jipolanco:
Should close #929

It probably makes sense to propose this variant of `find_library` to base Julia.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -5,6 +5,31 @@ const depsfile = joinpath(@__DIR__, "deps.jl")
 libpath = get(ENV, "JULIA_HDF5_PATH",
                 get(ENV, "JULIA_HDF5_LIBRARY_PATH", nothing)) # legacy env variable for compatibility
 
+# We avoid calling Libdl.find_library to avoid possible segfault when calling
+# dlclose (#929).
+# The only difference with Libdl.find_library is that we allow custom dlopen
+# flags via the `flags` argument.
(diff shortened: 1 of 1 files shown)


musm:
Good hunting!

> It probably makes sense to propose this variant of find_library to base Julia.

👍

Issue #409, opened 2017-05-23
Loading (parts of) HDF5 datasets into existing arrays?


oschulz:
From what I can see, there's currently no easy way to load an HDF5 dataset (or parts of it) into an existing array (to avoid memory allocation/ GC costs, esp. in multi-threaded applications). We could provide methods for `setindex!` and `view` to support for something like

```julia
target[:] = ds
```

and (to read a fragment of a DS into a fragment of an array)

```julia
target[a:b, ...] = view(ds, c:d, ...)
```

oschulz:
Something like

```julia
similar(::Type{Array}, ds::HDF5Dataset)
```

might also come in handy in that context.

andyferris:
Many data sets (e.g. Unchunked, uncompressed ones) will be memory mapped, in which case the OS is already providing a view of the data on disk. 

Still, something to cover the general case would be awesome.

oschulz:
> Many data sets (e.g. Unchunked, uncompressed ones) will be memory mapped

Sure - but I'll have to deal with large files with chunked and compressed datasets quite often (and therefore, have to process them out-of-core, in chunks). If people are fine with the proposal above, I can implement it and do a PR. I just wanted to gauge acceptance, first.

andyferris:
It seems very reasonable to me. Perhaps an `AbstractArray` type which is the full view of the dataset, and use e.g. `SubArray` for subsets/views thereof?

timholy:
If you need to keep track of which chunk you're working on, but still read only small chunks (i.e., never have the "full view" available), then an `OffsetArray` is handy:

```julia
function read_chunk!(buf, ds, inds...)
    buf[:] = view(ds, inds...)
    OffsetArray(buf, inds)
end
```

oschulz:
> ... then an OffsetArray is handy

Thanks, Tim, that could indeed come in very useful. To what extent are arrays with not-one-based indexing are supported by Base and common packages, currently?

Issue #930, opened 2022-05-04
HDF5.Dataset as an AbstractArray


mkitti:
Should `HDF5.Dataset` be an `AbstractArray`?

Alternatively, should we build a wrapper?
```julia
julia> struct DatasetArrayWrapper{T, N} <: AbstractArray{T,N}
           parent::HDF5.Dataset
       end

julia> DatasetArrayWrapper(parent::HDF5.Dataset) = DatasetArrayWrapper{eltype(parent), ndims(parent)}(parent)
DatasetArrayWrapper

julia> Base.size(w::DatasetArrayWrapper) = size(w.parent)

julia> Base.getindex(w::DatasetArrayWrapper, I...) = getindex(w.parent, I...)

julia> Base.getindex(w::DatasetArrayWrapper, I::Vararg{Int}) = getindex(w.parent, I...)
```

We almost have the entire AbstractArray interface implemented for `HDF5.Dataset`:
https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array

musm:
How does the wrapper benefit over a direct subtyping of `Dataset <: AbstractArray`  

Logically, makes sense that it would be a subtype. I think the original rationale might have been to limit it's interface. 
Are there any potential unintended side effects of this change?

mkitti:
We probably should implement AbstractDiskArray.jl. Perhaps stealing this, but querying more of the information rather than caching it like this implementation: https://github.com/AStupidBear/HDF5Utils.jl/blob/master/src/diskarrays.jl

musm:
Do you think it would be best to make it a direct subtype:
```
HDF5.Dataset <: AbstractDiskArray 
```
vs. what HDF5Utils is doing where
```
HDF5DiskArray <: AbstractDiskArray
```
and then defining
```
function HDF5DiskArray(dset::HDF5Dataset)
...
```
Which I think they are only  doing since they don't own the `HDF5.Dataset` type.

mkitti:
Yes, we should look into making it a direct subtype. This is a 0.17 topic though.

In contrast to the HDF5Utils.jl implementation, I think our implementation should be lazy.

Another issue if we make it a direct subtype is that we will need to parameterize it. We will need to consider how to mitigate the compilation latency in this case.

rafaqz:
Please do subtype `AbstractDiskArray`. You can also just use the macros to apply the interface to you own types without subtyping, but subtyping is simpler for dispatch.

[...]

PR #937, opened 2022-05-13
Add copyto! and similar for Datasets

### LINKED ISSUE #580: Permit reading values into a preexisting array buffer
Is there a way of reading from hdf5 directly into a preexisting array buffer? 

Edit: It would also be useful to expose a version of `getindex` or ` copyto!` or something for retrieving part of an array - that let you pass in the dest array instead of [allocating it every time](https://github.com/JuliaIO/HDF5.jl/blob/master/src/HDF5.jl#L1781).

mkitti:
Closes #580, #409

Here we implement <strike>`read!`</strike>, `copyto!`, and `similar` to allow data to be read into existing array buffers rather than allocating new buffers.


src/views.jl
```diff
@@ -0,0 +1,41 @@
...
+# Long term consideration:
+# If Dataset and Attribute become an AbstractArray, perhaps these should just be SubArrays ?
+"""
+    DatasetView
+
+This is an experimental data type and may change. Use `view` rather than `DatasetView` directly.
+"""
+struct DatasetView
+    parent::Dataset
+    indices
```
simonbyrne, line 12: Would it make sense to store the dataspace?

mkitti: I think this is worthwhile eventually, but I think we should defer this to a larger overhaul of `Dataset`.

In particular, we should consider parameterizing `Dataset` and making it an `AbstractArray` or `AbstractDiskArray` for version 0.17.

mkitti: See #930

mkitti: I'm leaning towards removing the view feature from this pull request.

simonbyrne: I think it's a good idea, but I would suggest making it actually a functional part: i.e. calling `read!(dataset, array, I...)` would internally call `read!(view(dataset, I...), array)`

mkitti: I agree. I just think the `view` component is much more complicated than the other parts.

src/HDF5.jl
```diff
@@ -957,6 +957,19 @@ function Base.read(obj::DatasetOrAttribute, ::Type{T}, I...) where T
     return val
 end
 
+"""
+    read!(obj::DatasetOrAttribute, output_buffer::AbstractArray{T} [, I...]) where T
+
+Read [part of] a dataset or attribute into a preallocated output buffer.
+The output buffer must be convertible to a pointer and have a contiguous layout.
+"""
+function Base.read!(obj::DatasetOrAttribute, buf::AbstractArray{T}, I...) where T
```
musm, line 966: I'm curious why do we need this method? Seems like `copyto!` alone would be sufficient as per the AbstractArray interface.

src/HDF5.jl
```diff
@@ -1062,6 +1102,89 @@ function generic_read(obj::DatasetOrAttribute, filetype::Datatype, ::Type{T}, I.
     end
 end
 
+
+"""
+    similar(obj::DatasetOrAttribute, [::Type{T}], [I::Integer...])
+
+Return a `Array{T}` or `Matrix{UInt8}` to that can contain [part of] the dataset.
+"""
+function Base.similar(obj::DatasetOrAttribute, I::Integer...)
```
musm, line 1111: Question, why do we need to pass the indices here?

mkitti: `Base.similar` accepts an optional size value. These are not indices. I suppose we should rename the argument.
https://docs.julialang.org/en/v1/base/arrays/#Base.similar

src/HDF5.jl
```diff
@@ -989,7 +1017,14 @@ function generic_read(obj::DatasetOrAttribute, filetype::Datatype, ::Type{Opaque
 end
 
 # generic read function
+function generic_read!(buf::Union{AbstractMatrix{UInt8}, AbstractArray{T}}, obj::DatasetOrAttribute, filetype::Datatype, ::Type{T}, I...) where T
```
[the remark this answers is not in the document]
mkitti: Yes. Given that `generic_read` is defined in this package, we get to define the argument order. Since the buffer is the variable being altered, I put it first.

src/HDF5.jl
mkitti: We currently implement `Base.read`, so why not `Base.read!`?

src/HDF5.jl
musm: I was more specifically looking at the AbstractArray interface https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array, where there are 4 optional `similar` methods that we could define:
<html>
<body>


similar(A) | similar(A, eltype(A), size(A)) | Return a mutable array with the same shape and element type
-- | -- | --
similar(A, ::Type{S}) | similar(A, S, size(A)) | Return a mutable array with the same shape and the specified element type
similar(A, dims::Dims) | similar(A, eltype(A), dims) | Return a mutable array with the same element type and size dims
similar(A, ::Type{S}, dims::Dims) | Array{S}(undef, dims) | Return a mutable array with the specified element type and size


</body>
</html>

src/HDF5.jl
musm: I don't think we should define this method, because we  are moving more and more having the `HDF5.Dataset` type  satisfying  the  `AbstractArray` interface. Users would expect  `copyto!`  as that is the  standard name for this function, thus additionally adding `Base.read!` seems unnecessary.  

I understand that we are defining methods on `Base.read` throughout the `HDF5` module, but I don't see that we should extend that and add `Base.read!` 

The definition in IO/Networking, also suggests that this wouldn't be the best overload, since it's used by default for reading from a filename or stream. 
https://docs.julialang.org/en/v1/base/io-network/#Base.read!

src/HDF5.jl
mkitti: When we do implement AbstractArray in 0.17, then we do not need this signature. We could just implement the version with Dims. However, since we are not implementing AbstractArray in 0.16, we probably do need to retain a method with the above signature since we do not have a `similar(a::AbstractArray{T}, dims::Union{Integer, AbstractUnitRange}...)` to cover that case for us.

I will rename this to `Base.similar(obj::DatasetOrAttribute, dims::Dims)`. We probably sill need `Base.similar, dims::Integer...)` as well, until we actually do implement `AbstractArray`.

src/HDF5.jl
mkitti: `Base.read!` has been removed.

src/HDF5.jl
```diff
@@ -1104,38 +1061,81 @@ end
...
-function Base.similar(obj::DatasetOrAttribute, I::Integer...)
-    dtype = datatype(obj)
-    T = get_jl_type(dtype)
-    val = similar(obj, dtype, T, I...)
-    close(dtype)
-    return val
+function Base.similar(
+    obj::DatasetOrAttribute,
+    ::Type{T},
+    dims::Integer...;
```
musm, line 1073: Why not have `dims::Dims`  here and below?

src/HDF5.jl
```diff
@@ -989,11 +1011,133 @@ function generic_read(obj::DatasetOrAttribute, filetype::Datatype, ::Type{Opaque
...
+    return _generic_read(obj, filetype, T, buf, I...)
+end
 function generic_read(obj::DatasetOrAttribute, filetype::Datatype, ::Type{T}, I...) where T
+    return _generic_read(obj, filetype, T, nothing, I...)
+end
+function _generic_read(obj::DatasetOrAttribute, filetype::Datatype, ::Type{T},
+    buf::Union{AbstractMatrix{UInt8}, AbstractArray{T}, Nothing}, I...) where T
+
+    sz, scalar, dspace = _size_of_buffer(obj, I)
+    memtype = _memtype(filetype, T)
```
musm, line 1024: Maybe move this below the buffer size checks?

musm:
Overall LGTM sans minor comments. Latest changes are much improved.

src/HDF5.jl
musm: @simonbyrne  did you intend to comment on the other thread?   Is there a specific part of https://github.com/JuliaIO/HDF5.jl/pull/937#discussion_r888036072 you disagree with?  
The goal is not to remove `copyto!` and replace it with `read!`. Originally, @mkitti  had `read!` as an alias for `copyto!`, which we require in order to satisfy part of the `AbstractArray` interface. My comment on `read!` was that this was both redundant with `copyto!` and unexpected since `read!` is used in base Julia for reading from an IO stream/filename, which is very different than what `copyto!` accomplishes.

mkitti:
I'm not entirely sure about this `normalize` keyword for `similar`. For many circumstances it may not matter.

mkitti:
My plan after this is to revisit views of `Dataset` in #946

musm:
Awesome, thanks! The improvements and clean up to the generic read function are defintely welcome.

PR #943, opened 2022-05-29
move attributes to new file, add docs

simonbyrne:
As part of my intermittent effort to clean up the code, this moves all the attributes-related stuff to a new file, and adds some docs.

### DIFF
--- a/docs/src/attributes.md
+++ b/docs/src/attributes.md
@@ -1,0 +1,23 @@
+# Attributes
+
+```@docs
+HDF5.Attribute
+attributes
+```
+
+## Mid-level Interface
+
+```@docs
+open_attribute
+create_attribute
+read_attribute
+write_attribute
+delete_attribute
+```
+
+## Convenience interface
+
+```@docs
+h5readattr
+h5writeattr
+```
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1,0 +1,232 @@
+"""
+    HDF5.Attribute
+
+A HDF5 attribute: this is a piece of metadata attached to an HDF5 `Group` or
+`Dataset`. It acts like a `Dataset`, in that it has a defined datatype and
+dataspace, and can `read` and `write` data to it.
+
+See also
+- [`open_attribute`](@ref)
+- [`create_attribute`](@ref)
+- [`read_attribute`](@ref)
+- [`write_attribute`](@ref)
+- [`delete_attribute`](@ref)
+"""
+mutable struct Attribute
+    id::API.hid_t
+    file::File
+
+    function Attribute(id, file)
+        dset = new(id, file)
+        finalizer(close, dset)
+        dset
+    end
+end
+Base.cconvert(::Type{API.hid_t}, attr::Attribute) = attr
+Base.unsafe_convert(::Type{API.hid_t}, attr::Attribute) = attr.id
+function Base.close(obj::Attribute)
+    if obj.id != -1
+        if obj.file.id != -1 && isvalid(obj)
+            API.h5a_close(obj)
+        end
+        obj.id = -1
+    end
+    nothing
+end
+name(attr::Attribute) = API.h5a_get_name(attr)
+
+
+datatype(dset::Attribute) = Datatype(API.h5a_get_type(checkvalid(dset)), file(dset))
+dataspace(attr::Attribute) = Dataspace(API.h5a_get_space(checkvalid(attr)))
+
+function Base.write(obj::Attribute, x)
+    dtype = datatype(x)
+    try
+        write_attribute(obj, dtype, x)
+    finally
+        close(dtype)
+    end
+end
+
+"""
+    read_attribute(parent::Union{File,Group,Dataset,Datatype}, name::AbstractString)
+
+Read the value of the named attribute on the parent object.
+
+# Example
+```julia-repl
+julia> HDF5.read_attribute(g, "time")
+2.45
+```
+"""
+function read_attribute(parent::Union{File,Group,Dataset,Datatype}, name::AbstractString)
+    obj = open_attribute(parent, name)
+    try
+        return read(obj)
+    finally
+        close(obj)
+    end
+end
+read_attribute(attr::Attribute, memtype::Datatype, buf) = API.h5a_read(attr, memtype, buf)
+
+"""
+    open_attribute(parent::Union{File,Group,Dataset,Datatype}, name::AbstractString)
+
+Open the [`Attribute`](@ref) named `name` on the object `parent`.
+"""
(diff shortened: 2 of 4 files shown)


simonbyrne:
@musm @mkitti

mkitti:
Looks pretty good to me.

Should the Dataset, File, and Group related stuff really end up in attributes? Should we parcel those out to files as well?

Is there a bigger plan for a file tree?

simonbyrne:
> Should the Dataset, File, and Group related methods end up in attributes? Should we parcel those out to files as well?

I'm not sure what you mean: can you give an example?

> Is there a bigger plan for a file tree?

I'd like to split up the functionality so it is easier to navigate and understand the API. At the moment, it is not even clear what the user-facing APIs are, let alone where they're defined. I originally tried to split them out with #833, but that was too ambitious, so I decided a piecemeal approach might be easier to manage.

This also makes it a bit easier to refactor and identify smaller changes. For example, after looking at this, there are a couple of things we might to consider to refactor (but I'll leave for separate PRs):

- make `Attributes` a subtype of `AbstractDict{String, Attribute}`
- add additional methods to `Attributes` (e.g. `delete!`) to make it full-featured (so that the mid-level API is not required)
- deprecate "datasets act like attributes" with `getindex`/`setindex!` of strings (calling `attributes(obj)["name"]` isn't that hard).

mkitti:
> deprecate \"datasets act like attributes\" with getindex/setindex! of strings (calling attributes(obj)[\"name\"] isn't that hard).

While I kind of agree, I'm also not sure why this is a problem. Is there any ambiguity that this causes at the moment?

mkitti:
> make Attributes a subtype of AbstractDict{String, Attribute}

I think we should target this for 0.17.

simonbyrne:
> While I kind of agree, I'm also not sure why this is a problem. Is there any ambiguity that this causes at the moment?

1. it's inconsistent with other objects (Files or Groups), which use it for accessing sub-objects.
2. when reading code, it's a lot easier to understand what `attributes(dataset)["name"]` does.

(anyway, that was why i wanted to leave it to a new PR)

mkitti:
> I originally tried to split them out with #833,

I agree with changes in #833 more than I disagree with them. Using #833 as a practical example, perhaps prioritizing what to split out and why in an issue might be helpful.

src/attributes.jl
```diff
@@ -0,0 +1,232 @@
...
+
+Read the value of the named attribute on the parent object.
+
+# Example
+```julia-repl
+julia> HDF5.read_attribute(g, "time")
+2.45
+```
+"""
+function read_attribute(parent::Union{File,Group,Dataset,Datatype}, name::AbstractString)
```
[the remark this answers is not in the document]
simonbyrne: I'll admit that there aren't any clear rules, but there are a couple of reasons to keep it here:

- they all use the attribute (`h5a_XXX`) C API functions
- attributes are an "extra" thing stuck on top of HDF5 (i.e. you can define groups, datasets etc without referring to attributes at all)
- in this particular case, it is a method that applies to multiple objects (so if you were to put it with those, it would need to be defined multiple times)

docs/src/attributes.md
```diff
@@ -0,0 +1,23 @@
...
+write_attribute
+delete_attribute
+```
+
+## Convenience interface
+
+```@docs
+h5readattr
+h5writeattr
+```
```
simonbyrne, line 23: I don't know what to call these functions (honestly, these might be another good candidate for deprecation)

mkitti:
> > While I kind of agree, I'm also not sure why this is a problem. Is there any ambiguity that this causes at the moment?
> 
> 1. it's inconsistent with other objects (Files or Groups), which use it for accessing sub-objects.
> 2. when reading code, it's a lot easier to understand what `attributes(dataset)["name"]` does.
> 
> (anyway, that was why i wanted to leave it to a new PR)

Is it possible to have an attribute or dataset with the same name in a group?

One technical advantage I see to your approach is that we would likely improve type stability overall by doing this.

I do think `attributes(dataset)["name"]` is a bit verbose and hard to intuit though.

mkitti:
Overall, I would be kn favor of merging this sooner rather than later.

mkitti (approved):

simonbyrne:
> Is it possible to have an attribute or dataset with the same name in a group?

Yes:
```
julia> h5 = h5open("xx.h5", "cw")
🗂️ HDF5.File: (read-write) xx.h5

julia> attributes(h5)["aa"] = [1.0]
1-element Vector{Float64}:
 1.0

julia> h5["aa"] = [2.0]
1-element Vector{Float64}:
 2.0

julia> h5
🗂️ HDF5.File: (read-write) xx.h5
├─ 🏷️ aa
└─ 🔢 aa
```

mkitti:
> > Is it possible to have an attribute or dataset with the same name in a group?
> 
> Yes:

I'm convinced. The interface needs to be deprecated.

mkitti:
Let's set a deadline of May 4th to merge this. I just want to give it more time because of the US holiday so everyone can see it.

src/attributes.jl
```diff
@@ -0,0 +1,232 @@
...
+    file::File
+
+    function Attribute(id, file)
+        dset = new(id, file)
+        finalizer(close, dset)
+        dset
+    end
+end
+Base.cconvert(::Type{API.hid_t}, attr::Attribute) = attr
+Base.unsafe_convert(::Type{API.hid_t}, attr::Attribute) = attr.id
```
musm, line 26: 👍

musm:
Looks great to me, I'm also in strong approval of all the suggestions in 

> 
> * make `Attributes` a subtype of `AbstractDict{String, Attribute}`
> * add additional methods to `Attributes` (e.g. `delete!`) to make it full-featured (so that the mid-level API is not required)
> * deprecate "datasets act like attributes" with `getindex`/`setindex!` of strings (calling `attributes(obj)["name"]` isn't that hard).

On the last bullet, I agree.  It's not very  clear  why dataset act like attributes and it's certainly far more descriptive to make users call attributes on the dataset if that's what they want to work with.

simonbyrne:
After looking at this a bit more, I thing one other change that may make sense is that `getindex(::Attribute, ::String)` should actually `read` the attribute value, rather than returning an `Attribute` object (since there is no way to `read` only part of an attribute).

simonbyrne:
This appears to be what h5py does: https://docs.h5py.org/en/stable/quick.html#attributes

mkitti:
Maybe we should have a `attrs` that returns a `AbstractDict{String, Any}` and acts like the h5py interface.

Meanwhile, perhaps we should parameterize attribute by its Julia type and have it work like a `Ref`.

PR #944, opened 2022-05-29
Add dset_no_attrs_hint support, h5f coverage up to v1.12

mkitti:
The primary focus of this pull request is to add
* `HDF5.API.h5f_get_dset_no_attrs_hint`
* `HDF5.API.h5f_set_dset_no_attrs_hint`

This pull request adds helpers for
* `HDF5.API.h5f_get_dset_no_attrs_hint`
* `HDF5.API.h5p_get_dset_no_attrs_hint`

The `DatasetCreateProperties` gains a `no_attrs_hint` property.

Increased support is added for H5F* functions in general, covering known functions up to HDF5 v1.12.

Apart of that effort, this pull request converts `H5F_libver_*` into an enum from a series of constants.

A practical benefit of the `dset_no_attrs_hint` enhancements is a smaller HDF5 file with data:
```julia
julia> h5open("test.h5", "w"; libver_bounds = :latest, meta_block_size = 294) do f
           HDF5.API.h5f_set_dset_no_attrs_hint(f, true)
           f["test"] = 0x1
           f["test"] |> HDF5.API.h5d_get_offset |> Int
       end
294

julia> filesize("test.h5")
295

julia> h5open("test.h5", "w"; libver_bounds = :latest, meta_block_size = 294) do f
           HDF5.API.h5f_set_dset_no_attrs_hint(f, false)
           f["test"] = 0x1
           f["test"] |> HDF5.API.h5d_get_offset |> Int
       end
479

julia> filesize("test.h5")
480
```


mkitti:
Only failure is codecov on functions.jl since I just added a bunch of bindings from LibHDF5.jl

simonbyrne:
Looks like also a failure using the system hdf5 (which uses 1.10 I think)

mkitti:
Good point.  We need to gate at 1.10.5 

https://docs.hdfgroup.org/hdf5/v1_12/group___h5_f.html#gacbf3ba8b36750c42b49740567a9732c4

Also does anyone know how to make doxygen URLs more readable?

simonbyrne:
Isn't this the same information as https://portal.hdfgroup.org/display/HDF5/H5F_GET_DSET_NO_ATTRS_HINT?

mkitti:
It is, I'm just worried that the newer doxygen based docs will replace the older docs at some point.  Doxygen is geared more towards C++ though so they hash the method name and the argument types.

mkitti:
> Looks like also a failure using the system hdf5 (which uses 1.10 I think)

I fixed this. The remaining check fail is now truly code coverage.

mkitti:
I broke and fixed this again. Main fail is due to codecov on `src/api/functions.jl` since I added the h5f bindings.

src/properties.jl
```diff
@@ -496,6 +497,11 @@ function class_getproperty(::Type{DatasetCreateProperties}, p::Properties, name:
     name === :external    ? API.h5p_get_external(p) :
     name === :filters     ? get_filters(p) :
     name === :layout      ? get_layout(p) :
+    name === :no_attrs_hint ?
```
simonbyrne, line 500: add to the docstring for `DatasetCreateProperties`

simonbyrne (approved):

src/properties.jl
mkitti: done in [a523356](https://github.com/JuliaIO/HDF5.jl/pull/944/commits/a5233563d8fd83c1e0476b5470365f374ea0e3b4)

test/fileio.jl
```diff
@@ -60,4 +60,37 @@ let fn = tempname() * ".h5"
...
+  @test filesize(fn) == threshold + 1
+  h5open(
+    fn, "w";
+    libver_bounds = :latest,
+    meta_block_size = threshold
+  ) do f
+    HDF5.API.h5f_set_dset_no_attrs_hint(f, false)
+    @test !HDF5.API.h5f_get_dset_no_attrs_hint(f)
+    f["test"] = 0x1
+    # We expect that with the hint, the offset will actually be 300
```
musm, line 90: This comment needs to change

docs/src/api_bindings.md
```diff
@@ -169,37 +169,97 @@ h5e_walk
 ---
 
 ## [[`H5F`](https://portal.hdfgroup.org/display/HDF5/Files) — File Interface](@id H5F)
```
musm, line 171: Presumably these were all determined using the generator you have developed?  Thus I haven't reviewed them for correctness, since I think the generator has been working without issues.

test/fileio.jl
mkitti: I first figured out what the minimum size would be, and picked a round number greater than that. 

https://twitter.com/markkitti/status/1530811272966586368?t=kyNK_e2Gwr87Xuv9mhYVog&s=19

docs/src/api_bindings.md
mkitti: Yes they are coming from https://github.com/mkitti/LibHDF5.jl/blob/712b6e306a15de37f748727b37676aca70ea0664/src/bind.jl

That is generated by this code:
https://github.com/mkitti/LibHDF5.jl/blob/main/gen/bind_generator.jl

test/fileio.jl
musm, line 90:
```suggestion
    # We expect that with the hint, the offset will be greater than 300
```

mkitti: Is it clear that we mean with the hint set to `false`?

PR #948, opened 2022-06-04
Define AttibuteDict

simonbyrne:
This follows on from https://github.com/JuliaIO/HDF5.jl/pull/943#issuecomment-1140504666

It defines `attrs(obj)` which returns an `AttibuteDict`: this is mostly the same as `Attributes` object, except:
- it is a subtype of `AbstractDict{String, Any}`, and supports the relevant features (`keys`, `pairs`, `values`)
- `getindex` will `read` instead of returning an `Attribute`
- it supports `delete!`
- it supports overwriting attributes

TODO:
- [ ] update docs
- [ ] deprecate `attributes`

### DIFF
--- a/src/api/helpers.jl
+++ b/src/api/helpers.jl
@@ -53,7 +53,11 @@ end
 # `@cfunction` with `$f`.
 # This helper translates between the two preferred forms for each respective language.
 function h5a_iterate_helper(loc_id::hid_t, attr_name::Ptr{Cchar}, ainfo::Ptr{H5A_info_t}, @nospecialize(f::Any))::herr_t
-    return f(loc_id, attr_name, ainfo)
+    try
+        return herr_t(f(loc_id, attr_name, ainfo))
+    catch e
+        return herr_t(-1)
+    end
 end
 
 """
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1,3 +1,5 @@
+# mid-level API
+
 """
     HDF5.Attribute
 
@@ -77,7 +79,6 @@ Open the [`Attribute`](@ref) named `name` on the object `parent`.
 open_attribute(parent::Union{File,Object}, name::AbstractString, aapl::AttributeAccessProperties=AttributeAccessProperties()) =
     Attribute(API.h5a_open(checkvalid(parent), name, aapl), file(parent))
 
-
 """
     create_attribute(parent::Union{File,Object}, name::AbstractString, dtype::Datatype, space::Dataspace)
     create_attribute(parent::Union{File,Object}, name::AbstractString, data)
@@ -102,24 +103,25 @@ function create_attribute(parent::Union{File,Object}, name::AbstractString, dtyp
     return Attribute(attrid, file(parent))
 end
 
-
+# generic method
+write_attribute(attr::Attribute, memtype::Datatype, x) = API.h5a_write(attr, memtype, x)
+# specific methods
 function write_attribute(attr::Attribute, memtype::Datatype, str::AbstractString)
     strbuf = Base.cconvert(Cstring, str)
     GC.@preserve strbuf begin
         buf = Base.unsafe_convert(Ptr{UInt8}, strbuf)
-        API.h5a_write(attr, memtype, buf)
+        write_attribute(attr, memtype, buf)
     end
 end
 function write_attribute(attr::Attribute, memtype::Datatype, x::T) where {T<:Union{ScalarType,Complex{<:ScalarType}}}
     tmp = Ref{T}(x)
-    API.h5a_write(attr, memtype, tmp)
+    write_attribute(attr, memtype, tmp)
 end
 function write_attribute(attr::Attribute, memtype::Datatype, strs::Array{<:AbstractString})
     p = Ref{Cstring}(strs)
-    API.h5a_write(attr, memtype, p)
+    write_attribute(attr, memtype, p)
 end
 write_attribute(attr::Attribute, memtype::Datatype, ::EmptyArray) = nothing
-write_attribute(attr::Attribute, memtype::Datatype, x) = API.h5a_write(attr, memtype, x)
 
 """
     write_attribute(parent::Union{File,Object}, name::AbstractString, data)
@@ -127,19 +129,26 @@ write_attribute(attr::Attribute, memtype::Datatype, x) = API.h5a_write(attr, mem
 Write `data` as an [`Attribute`](@ref) named `name` on the object `parent`.
 """
 function write_attribute(parent::Union{File,Object}, name::AbstractString, data; pv...)
-    obj, dtype = create_attribute(parent, name, data; pv...)
+    attr, dtype = create_attribute(parent, name, data; pv...)
     try
-        write_attribute(obj, dtype, data)
+        write_attribute(attr, dtype, data)
     catch exc
         delete_attribute(parent, name)
         rethrow(exc)
     finally
-        close(obj)
+        close(attr)
         close(dtype)
     end
     nothing
 end
 
+"""
+    rename_attribute(parent::Union{File,Object}, oldname::AbstractString, newname::AbstractString)
+
+Rename the [`Attribute`](@ref) of the object `parent` named `oldname` to `newname`.
+"""
+rename_attribute(parent::Union{File,Object}, oldname::AbstractString, newname::AbstractString) =
+    API.h5a_rename(checkvalid(parent), oldname, newname)
 
 """
     delete_attribute(parent::Union{File,Object}, name::AbstractString)
@@ -186,6 +195,84 @@ function h5readattr(filename, name::AbstractString)
     end
     dat
 end
+
+
+struct AttributeDict <: AbstractDict{String,Any}
+    parent::Object
+end
+
+"""
+    attrs(object::Union{File,Group,Dataset,Datatype})
+
+The attributes dictionary of `object`. Returns an `AttributeDict`, a `Dict`-like
+object for accessing the attributes of `object`.
+
+```julia
+attrs(object)["name"] = value  # create/overwrite an attribute
+attr = attrs(object)["name"]   # read an attribute
+delete!(attrs(object), "name") # delete an attribute
+keys(attrs(object))            # list the attribute names
+```
+"""
+function attrs(parent::Object)
+    return AttributeDict(parent)
+end
+attrs(file::File) = attrs(open_group(file, "."))
+
+Base.haskey(attrdict::AttributeDict, path::AbstractString) = API.h5a_exists(checkvalid(attrdict.parent), path)
+Base.length(attrdict::AttributeDict) = Int(object_info(attrdict.parent).num_attrs)
+
+function Base.getindex(x::AttributeDict, name::AbstractString)
+    haskey(x, name) || throw(KeyError(name))
+    read_attribute(x.parent, name)
+end
+function Base.get(x::AttributeDict, name::AbstractString, default)
+    haskey(x, name) || return default
+    read_attribute(x.parent, name)
+end
+function Base.setindex!(attrdict::AttributeDict, val, name::AbstractString)
+    if haskey(attrdict, name)
+        # in case of an error, we write first to a temporary, then rename
+        _name = tempname()
+        try
+            write_attribute(attrdict.parent, _name, val)
+            delete_attribute(attrdict.parent, name)
+            rename_attribute(attrdict.parent, _name, name)
+        finally
+            haskey(attrdict, _name) && delete_attribute(attrdict.parent, _name)
+        end
+    else
+        write_attribute(attrdict.parent, name, val)
+    end
+end
+Base.delete!(attrdict::AttributeDict, path::AbstractString) = delete_attribute(attrdict.parent, path)
+
+function Base.keys(attrdict::AttributeDict)
+    # faster than iteratively calling h5a_get_name_by_idx
+    checkvalid(attrdict.parent)
+    keyvec = sizehint!(String[], length(attrdict))
+    API.h5a_iterate(attrdict.parent, IDX_TYPE[], ORDER[]) do _, attr_name, _
+        push!(keyvec, unsafe_string(attr_name))
+        return false
+    end
+    return keyvec
+end
+
+function Base.iterate(attrdict::AttributeDict)
+    # constuct key vector, then iterate
+    # faster than calling h5a_open_by_idx
+    iterate(attrdict, (keys(attrdict), 1))
+end
+function Base.iterate(attrdict::AttributeDict, (keyvec, n))
+    iter = iterate(keyvec, n)
+    if isnothing(iter)
+        return iter
+    end
+    key, nn = iter
+    return (key => attrdict[key]), (keyvec, nn)
+end
+
+
 
 
 struct Attributes
--- a/src/show.jl
+++ b/src/show.jl
@@ -63,6 +63,16 @@ function Base.show(io::IO, attr::Attributes)
     print(io, "Attributes of ", attr.parent)
 end
 
+Base.show(io::IO, attrdict::AttributeDict) = summary(io, attrdict)
+function Base.summary(io::IO, attrdict::AttributeDict)
+    print(io, "AttributeDict of ", attrdict.parent)
+    if isvalid(attrdict.parent)
+        n = length(attrdict)
+        print(io, " with ", n, n == 1 ? " attribute" : " attributes")
+    end
+end
+
+
 const ENDIAN_DICT  = Dict(
     API.H5T_ORDER_LE    => "little endian byte order",
     API.H5T_ORDER_BE    => "big endian byte order",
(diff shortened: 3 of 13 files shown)


mkitti:
I'm not sure if we need to deprecate `attributes`. I think there might be some use in making that API type safe by parameterizing `Attribute`.

simonbyrne:
> I'm not sure if we need to deprecate `attributes`. I think there might be some use in making that API type safe by parameterizing `Attribute`.

It's kind of confusing to keep both, and everything it does can be better handled by the mid-level APIs.

mkitti:
Even if we are to deprecate `attributes`, we still need to test that the old interface continues to work until we remove it.

For the moment, we need to duplicate the tests, not replace them. We can remove the tests when we actually remove the API.

src/attributes.jl
```diff
@@ -188,45 +193,82 @@ function h5readattr(filename, name::AbstractString)
 end
 
 
-struct Attributes
-    parent::Union{File,Object}
+struct AttributeDict <: AbstractDict{String,Any}
+    parent::Object
```
mkitti, line 197: Is `Union{File, Object}` redundant?

src/attributes.jl
```diff
@@ -188,45 +193,82 @@ function h5readattr(filename, name::AbstractString)
 end
 
 
-struct Attributes
-    parent::Union{File,Object}
+struct AttributeDict <: AbstractDict{String,Any}
```
mkitti, line 196: Let's consider parameterizing this so that we can have a more specific type than `Any`, which can be the default is not specified.

src/attributes.jl
```diff
@@ -188,45 +193,82 @@ function h5readattr(filename, name::AbstractString)
...
+attrs(file::File) = attrs(open_group(file, "."))
 
-Base.isvalid(obj::Attributes) = isvalid(obj.parent)
+Base.haskey(attrdict::AttributeDict, path::AbstractString) = API.h5a_exists(checkvalid(attrdict.parent), path)
+Base.length(attrdict::AttributeDict) = Int(object_info(attrdict.parent).num_attrs)
 
-function Base.getindex(x::Attributes, name::AbstractString)
+function Base.getindex(x::AttributeDict, name::AbstractString)
     haskey(x, name) || throw(KeyError(name))
-    open_attribute(x.parent, name)
```
mkitti, line 208: We can add the AttributeDict methods without removing the older interface. This still needs to work even when deprecated.

mkitti (changes requested):
Overall I like the idea of having a `AbstractDict{String, Any}` interface emulating h5py. The interface is simple to use. However, we should also figure out a way to provide a Julian interface with additional type safe features.

We should retain the tests for the current `attributes` API until removal. We cannot remove this without a deprecation period and at least a minor version bump.

I would also prefer to have deprecation in another PR while we evaluate the matter. It would also help to narrow the diff. We should consider if there might be a Julian way to emulate the C++ highfive interface with parametric types:
https://bluebrain.github.io/HighFive/class_high_five_1_1_attribute.html

One consideration is we could also specify the value type so that we can have an `AttributeDict{T} <: AbstractDict{String, T} where T` rather than the value being `Any`. Accessing attributes via this interface should use `convert` to coerce the type.

In summary, I think the easiest path here would be to focus on adding the new interface and make it optionally type safe.

src/api/helpers.jl
```diff
@@ -53,7 +53,11 @@ end
 # `@cfunction` with `$f`.
 # This helper translates between the two preferred forms for each respective language.
 function h5a_iterate_helper(loc_id::hid_t, attr_name::Ptr{Cchar}, ainfo::Ptr{H5A_info_t}, @nospecialize(f::Any))::herr_t
-    return f(loc_id, attr_name, ainfo)
+    try
+        return herr_t(f(loc_id, attr_name, ainfo))
+    catch e
```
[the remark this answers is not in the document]
simonbyrne: it would probably be better to pass the error back out via a reference, and rethrow it from outside

src/attributes.jl
simonbyrne: Attributes attached to a file are really attached to the root group, so i made this explicit.

src/attributes.jl
simonbyrne: I don't see how that could be made to work, or really,what advantage it would give. Can you expand on what you had in mind? What would happen if you attempted to read or write an attribute of a different type?

My intention was to provide a simple, easy-to-use interface. I think if you want to specify the return type, it would be better to do in the mid-level interface, e.g. adding a return type to `read_attribute`.

simonbyrne:
> I would also prefer to have deprecation in another PR while we evaluate the matter. It would also help to narrow the diff. 

Okay, I can move the deprecations to a new PR.

> We should consider if there might be a Julian way to emulate the C++ highfive interface with parametric types:
> https://bluebrain.github.io/HighFive/class_high_five_1_1_attribute.html

I admit to not having much experience with C++, but this looks similar to the mid-level interface, combined with adding a type parameter or output array to `read_attribute`

src/attributes.jl
mkitti: Perhaps `AttributeDict{T}` could use `convert(T, ...)`? If `convert` throws an error, then so be it.

simonbyrne: It adds a lot of complexity and potential brittleness, and I'm still skeptical as to why it would be helpful? The files I've seen with attributes use a mix of different types (typically strings, integers and vectors of integers).

mkitti: Here's my prototype diff:
https://github.com/JaneliaSciComp/HDF5.jl/commit/c3a5db33b1cf96e245c6ae8bf5edaf61a10073e6
I added a `try` to `Base.iterate` to skip over items where `convert` fails.

Type stability can be helpful. Since I'm accessing the HDF5 file in Julia rather than Python, I would like to take advantage of the type system. In many cases, I may have prior knowledge of the type I'm expecting the attribute to be, or I can glean that from the API. This provides an easy way to enforce knowledge of the type, allowing subsequent code to be type stable without having to use a completely different API. It also throws an error if my assumptions are violated.

Having the parameter default to `Any` produces the exact same results you have now. Allowing for a parameter allows the same API to be extended to encourage type stability.


```julia
julia> attrs(h5f)
AttributeDict of HDF5.Group: / (file: test.h5) with 4 attributes:
  "bye"     => "3.0"
  "hello"   => 5
  "someint" => 3
  "test"    => 5.5

julia> attrs(h5f, Int)
AttributeDict of HDF5.Group: / (file: test.h5) with 4 attributes:
  "hello"   => 5
  "someint" => 3

julia> attrs(h5f, String)
AttributeDict of HDF5.Group: / (file: test.h5) with 4 attributes:
  "bye" => "3.0"

julia> attrs(h5f, Float64)
AttributeDict of HDF5.Group: / (file: test.h5) with 4 attributes:
  "hello"   => 5.0
  "someint" => 3.0
  "test"    => 5.5

julia> attrs(h5f, Int)["test"]
ERROR: InexactError: Int64(5.5)
Stacktrace:
 [1] Int64
   @ ./float.jl:812 [inlined]
 [2] convert(#unused#::Type{Int64}, x::Float64)
   @ Base ./number.jl:7
 [3] getindex(x::HDF5.AttributeDict{Int64}, name::String)
   @ HDF5 ~/.julia/dev/HDF5/src/attributes.jl:224
 [4] top-level scope
   @ REPL[12]:1
```

There a few things to explore here, such as passing the parameter as the `memtype` to `read_attribute`, but I think this would allow this API to go a lot further. Since we are talking about parameterizing `Dataset` as well, I think adding a type parameter to `AttributeDict` would be consistent with future direction of this package.

mkitti: `Dataset` would be `struct Dataset{T,N} <: AbstractArray{T,N}` or `struct Dataset{T,N} <: AbstractDiskArray{T,N}`.

I wanted to explore if `Attribute` could become `Attribute{T}` and have a corresponding `struct Attributes{T2} <: AbstractDict{String, Attribute{<:T2}}`. However, you want to deprecate that interface, which is why I'm exploring the idea here.

I would be fine if we parameterized the struct but only implemented methods for `AttributeDict{Any}` in this pull request.

mkitti:
> I admit to not having much experience with C++, but this looks similar to the mid-level interface, combined with adding a type parameter or output array to `read_attribute`

Here's an excerpt from their code: https://github.com/BlueBrain/Brion/blob/2b42ecd9c251741c42970141f34d292aee53d991/brion/plugin/compartmentReportHDF5.cpp#L78-L98

```cpp
bool _verifyFile(const HighFive::File& file)
{
    try
    {
        uint32_t magic = 0;
        file.getAttribute("magic").read(magic);
        if (magic != _sonataMagic)
            return false;

        std::vector<uint32_t> version;
        file.getAttribute("version").read(version);
        if (version.size() != 2 || version[0] != _currentVersion[0] ||
            version[1] != _currentVersion[1])
            return false;
    }
    catch (HighFive::Exception& e)
    {
        return false;
    }
    return true;
}

```

simonbyrne:
Thanks that is helpful.

If we were to parametrise `Attribute`, would both `magic` and `version` be `Attribute{UInt32}`? or would `version` be `Attribute{Vector{UInt32}}`?

src/api/helpers.jl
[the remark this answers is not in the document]
simonbyrne: I've moved this change to a new PR #950.

mkitti:
> Thanks that is helpful.
> 
> If we were to parametrise `Attribute`, would both `magic` and `version` be `Attribute{UInt32}`? or would `version` be `Attribute{Vector{UInt32}}`?

```
julia> typeof(magic_attr)
Attribute{UInt32}

julia> typeof(version_attr)
Attribute{Vector{UInt32}}
```

`version` appears to be a `Vector{UInt32}` of with a length of 2.

mkitti:
The more I think about it, Attributes is really more like a `NamedTuple` or a `struct` than a `Dict`. We know the types of all of the attributes.

src/attributes.jl
simonbyrne: I still don't see the benefit of adding a parameter to `AttributeDict`: how would it help?

simonbyrne:
> The more I think about it, Attributes is really more like a `NamedTuple` or a `struct` than a `Dict`. We know the types of all of the attributes.

It's not static though: I can add or delete attributes

simonbyrne:
> `version` appears to be a `Vector{UInt32}` of with a length of 2.

I meant what would the attribute type parameter be?

mkitti:
`Attribute{Vector{UInt32}}` would correspond to "version". It's like a `RefValue`.

musm:
Could you move the deprecated methods to the `src/deprecation.jl` file?

simonbyrne:
> It's a shame we can't use `attributes` for this.

Agreed, but once it's deprecated it become easier.

src/attributes.jl
```diff
@@ -188,6 +193,84 @@ function h5readattr(filename, name::AbstractString)
...
+attrs(object)["name"] = value  # create/overwrite an attribute
+attr = attrs(object)["name"]   # read an attribute
+delete!(attrs(object), "name") # delete an attribute
+keys(attrs(object))            # list the attribute names
+```
+"""
+function attrs(parent::Object)
+    return AttributeDict(parent)
+end
+attrs(file::File) = attrs(open_group(file, "."))
```
mkitti, line 216:
```suggestion
AttributeDict(file::File) = AttributeDict(open_group(file, "."))

function attrs(parent)
    return AttributeDict(parent)
end
```

Should the constructor work directly on `File`?

I'm also tempted to suggest `const attrs = AttributeDict`.

src/attributes.jl
```diff
@@ -188,6 +193,84 @@ function h5readattr(filename, name::AbstractString)
...
+
+Base.haskey(attrdict::AttributeDict, path::AbstractString) = API.h5a_exists(checkvalid(attrdict.parent), path)
+Base.length(attrdict::AttributeDict) = Int(object_info(attrdict.parent).num_attrs)
```
mkitti, line 219: Just use https://portal.hdfgroup.org/display/HDF5/H5A_GET_NUM_ATTRS

https://github.com/mkitti/LibHDF5.jl/blob/712b6e306a15de37f748727b37676aca70ea0664/src/bind.jl#L1216

src/attributes.jl
```diff
@@ -178,8 +184,7 @@ function h5readattr(filename, name::AbstractString)
     file = h5open(filename,"r")
     try
         obj = file[name]
-        a = attributes(obj)
-        dat = Dict(x => read(a[x]) for x in keys(a))
+        dat = Dict(attrs(obj))
```
mkitti, line 187:
```suggestion
        dat = Dict(AttributeDict(obj))
```

Internally, we should just use the constructor.

test/attributes.jl
```diff
@@ -0,0 +1,48 @@
...
+    filename = tempname()
+    f = h5open(filename, "w")
+
+    @test attrs(f) isa HDF5.AttributeDict
+
+    attrs(f)["a"] = 1
+    @test haskey(attrs(f), "a")
+    @test attrs(f)["a"] == 1
+
+    attrs(f)["b"] = [2,3]
```
mkitti, line 13: Test `attrs` on groups and datasets as well.

src/attributes.jl
simonbyrne: > This function is deprecated in favor of the functions H5O_GET_INFO, H5O_GET_INFO_BY_NAME, and H5O_GET_INFO_BY_IDX.

src/attributes.jl
musm: Is there any difference?

mkitti: Currently,  there is one less level of indirection. For no difference we need `const attrs = AttributeDict`.

mkitti:
I've had the opportunity to dig into `read` and analyze type stability. In principle, if we specified the `Type` to be returned we should be able to achieve type stability.

However, there are fundamental issues down to the `_generic_read` function that I just refactored. In particular, we specify the memory type, but we do not specify whether the result will be an array or a scalar based on an argument type. Thus the type stability is problematic down to the core of the reading mechanism. Currently calling `Base.read(dataset, T)` could return either `T` or `Array{T}`. To change this behavior would involve a breaking change. It cannot be achieved by modifying the mid-level API alone.

The only non-breaking route to type stable reads that I can see is by adding new methods. In a sense, the new `copyto!` is type stable. If we implemented, `read!`, I think we could also make that type stable. `read!` looks very much like the C++ API I posted above, which suggests that perhaps we should revisit implementing `read!`. In contrast, the best we could do without a significant overhaul at the moment is equivalent to adding a type assertion.

That said, I believe it would be possible to eventually use the following syntax to do type stable reads of attributes. It would just need to be rebased upon a lower level API than `read_attribute`.
```julia
magic = AttributeDict{UInt32}(file)["magic"]
version = AttributeDict{Vector{UInt32}}(file)["version"]
```
The advantage of this syntax is that the second type parameter for `AbstractDict` is clearly the return type.

My conclusion at the moment is that it is probably possible to achieve type stability through this new access mechanism, but I'm sympathetic to the idea that this might be too much work to be achieved in this pull request. Scoping everything to `AttributeDict{Any}` with the current implementation would allow a revision to expand this mechanism to enhance type stability and safety.

The main thing on my mind at the moment is expanding test coverage for `Group`, `Dataset`, and `Datatype`.

musm:
My conclusion is that we should proceed with the changes in this PR and revisit the type stability issue in other PRs. I don't think it's a high priority for the new interface. We should probably instead aim to get the new API in and then focus on stability later.

Personally, I'm not confident how much of a performance benefit stability could have, although I understand it's desirable in a lot of contexts. 

> However, there are fundamental issues down to the _generic_read function that I just refactored. In particular, we specify the memory type, but we do not specify whether the result will be an array or a scalar based on an argument type. Thus the type stability is problematic down to the core of the reading mechanism. Currently calling Base.read(dataset, T) could return either T or Array{T}. To change this behavior would involve a breaking change. It cannot be achieved by modifying the mid-level API alone.

I think we should look into this more seriously, even if it means a breaking change. Especially if it simplifies type-stability with the AttributeDict interface and other API methods. It sounds more robust than trying to monkey patch the AttributeDict interface for type stability, which doesn't sound entirely trivial with the current code.

mkitti:
Within HDF5.jl, the performance impact is likely not great. However, the performance impact could be significant in downstream applications as this makes dynamic dispatch much more likely. Moreover, the issue is about correctness. Can people write scientific programs using HDF5.jl that will be robust to variations in HDF5 files? Furthermore, can we balance convenience and correctness?

The immediate question is if there is anything that we can do here that will make iteration on this design easier in the future?

1. Parameterize the type, but only implement the case for `Any`.
2. Consider if we want `attrs` to be a method or an alias for `AttributeDict`

```julia
const attrs = AttributeDict
AttributeDict(x) = AttributeDict{Any}(x)

magic = attrs(file)["magic"]
version = attrs(file)["version"]
```

Later we might then implement the following:
```julia
magic = attrs{UInt32}(file)["magic"]
version = attrs{Vector{UInt32}}(file)["version"]
```

simonbyrne:
> 1. Parameterize the type, but only implement the case for `Any`.
> 2. Consider if we want `attrs` to be a method or an alias for `AttributeDict`

I think both of those changes could easily be made at a later date without causing breakages.

For the record, I am opposed to 1, as I fail to see the benefits from the added complexity (I think type-stable access would be better done via `read_attribute`), and am ambivalent about 2.

mkitti:
Adding a type parameter after release could be breaking if someone directly accessed the type, so we should be clear that it is not part of the public API.

I do not see how qualifying the current implementation as `AttributeDict{Any}` adds significant complexity, but I'll propose this in a separate pull request.

Let's expand the testing to directly test groups, datasets, and datatypes, then merge.

mkitti (approved):
Merge when ready

src/show.jl
```diff
@@ -63,6 +63,16 @@ function Base.show(io::IO, attr::Attributes)
     print(io, "Attributes of ", attr.parent)
 end
 
+Base.show(io::IO, attrdict::AttributeDict) = summary(io, attrdict)
```
musm, line 66: How does Julia end up printing the rest of the attributes ? Is that in some sort of fallback AbstractDict printing?

simonbyrne: Yes, the REPL printing is this one:
https://github.com/JuliaLang/julia/blob/803f90db9195c5c72df90d8a424c7066f1a8f2ee/base/show.jl#L133

musm: Ah it's the 3 argument show method. Thanks, I was wondering why `@which` was sending me to a seemingly unrelated call.

PR #949, opened 2022-06-06
Expand DatasetAccessProperties coverage, add helpers

mkitti:
Expanded `DatasetAccessProperties` to cover

- `chunk_cache`: Chunk cache parameters as (nslots, nbytes, w0).
  Default: (521, 0x100000, 0.75)
- `efile_prefix`: Path prefix for reading external files.
  The default is the current working directory.
  - `:origin`: alias for `raw"\$ORIGIN"` will make the external file relative to
    the HDF5 file.
- `virtual_prefix`: Path prefix for reading virtual datasets.
- `virtual_printf_gap`: The maximum number of missing source files and/or
   datasets with the printf-style names when getting the extent of an unlimited
   virtual dataset
- `virtual_view`: Influences whether the view of the virtual dataset includes
  or excludes missing mapped elements
  - `:first_missing`: includes all data before the first missing mapped data
  - `:last_available`: includes all available mapped data

Pending tasks:
- [ ] Add tests

### DIFF
--- a/src/api/helpers.jl
+++ b/src/api/helpers.jl
@@ -392,6 +392,14 @@ function h5p_get_chunk(plist_id)
     return dims, ndims
 end
 
+function h5p_get_chunk_cache(dapl_id)
+    nslots = Ref{Csize_t}()
+    nbytes = Ref{Csize_t}()
+    w0 = Ref{Cdouble}()
+    h5p_get_chunk_cache(dapl_id, nslots, nbytes, w0)
+    return (nslots = nslots[], nbytes = nbytes[], w0 = w0[])
+end
+
 function h5p_get_create_intermediate_group(plist_id)
     cig = Ref{Cuint}()
     h5p_get_create_intermediate_group(plist_id, cig)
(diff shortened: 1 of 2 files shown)


mkitti:
The only missing dataset access property is `append_flush`:
https://portal.hdfgroup.org/display/HDF5/H5P_GET_APPEND_FLUSH
https://portal.hdfgroup.org/display/HDF5/H5P_SET_APPEND_FLUSH

This has a signature unlike the others.

mkitti:
```julia
julia> dapl = HDF5.init!(HDF5.DatasetAccessProperties())
HDF5.DatasetAccessProperties(
  chunk_cache     = (nslots = 0x0000000000000209, nbytes = 0x0000000000100000, w0 = 0.75),
  efile_prefix    = "",
  virtual_prefix  = "",
  virtual_printf_gap = 0x0000000000000000,
  virtual_view    = :last_available,
)
```

src/api/helpers.jl
```diff
@@ -411,6 +419,11 @@ function h5p_get_efile_prefix(plist)
     return String(buffer)
 end
 
+function h5p_set_efile_prefix(plist, sym::Symbol)
+    sym === :origin ? h5p_set_efile_prefix(plist, raw"$ORIGIN") :
+    error("The only valid symbol for h5p_set_efile_prefix is :origin.")
```
musm, line 424: `throw(ArgumentError(....` ?

mkitti, line 424: Is this what you meant?

```suggestion
    throw(ArgumentError("The only valid `Symbol` argument for `h5p_set_efile_prefix` is `:origin`. Got `$sym`."))
```

musm: Yes exactly, perhaps you missed my comment above?

mkitti: I saw your comment. It has an odd number of backticks.

musm: Ah, typo.

mkitti:
This should be good to go for now. I think `virtual_printf_gap` and `virtual_view` testing will need to be deferred until we actually implement virtual datasets.

PR #950, opened 2022-06-07
make iterators safe for error handling

simonbyrne:
This makes some minor changes to the iterator API functions:
- the callbacks can now return any integer type, or a bool
- errors from the callback are trapped and safely propagated back to the main process (doing this at the moment could leave the library in an invalid state: see https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/#Creating-C-Compatible-Julia-Function-Pointers and https://discourse.julialang.org/t/what-happens-when-you-throw-an-error-from-a-cfunction/71510?u=simonbyrne): this was partially done in https://github.com/JuliaIO/HDF5.jl/pull/948#discussion_r889711496, but I figured it would be better to do this in a new PR
- there was an opportunity for a small optimization in the error stack printing function

### DIFF
--- a/src/api/helpers.jl
+++ b/src/api/helpers.jl
@@ -52,35 +52,52 @@ end
 # emulating it with the less desirable form of creating closure handles directly in
 # `@cfunction` with `$f`.
 # This helper translates between the two preferred forms for each respective language.
-function h5a_iterate_helper(loc_id::hid_t, attr_name::Ptr{Cchar}, ainfo::Ptr{H5A_info_t}, @nospecialize(f::Any))::herr_t
-    return f(loc_id, attr_name, ainfo)
+function h5a_iterate_helper(loc_id::hid_t, attr_name::Ptr{Cchar}, ainfo::Ptr{H5A_info_t}, @nospecialize(data::Any))::herr_t
+    f, err_ref = data
+    try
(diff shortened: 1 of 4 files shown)


src/api/helpers.jl
```diff
@@ -52,35 +52,52 @@ end
...
 ```julia-repl
 julia> HDF5.API.h5a_iterate(obj, HDF5.API.H5_INDEX_NAME, HDF5.API.H5_ITER_INC) do loc, name, info
            println(unsafe_string(name))
-           return HDF5.API.herr_t(0)
+           return false
        end
 ```
 """
 function h5a_iterate(@nospecialize(f), obj_id, idx_type, order, idx = 0)
+    err_ref = Ref{Any}(nothing)
```
musm, line 89: Why can't the reference always be an integer and instead of the `isnothing` checks, the value `jlerr < 0` is instead done?

simonbyrne: `jlerr` is the error object that is caught in the helper function, not an integer.

musm: I see, thanks for the clarification.

simonbyrne:
Any thoughts?

musm:
Thanks!

PR #953, opened 2022-06-09
Specify AttributeDict{Any}

mkitti:
* Specifies that the current implementation is `AttributeDict{Any}`, allows for other potential future implementations without breakage
* Does not implement anything for a generic `AttributeDict{T} where T`
* Refer to `HDF5.API.H5Error` rather than `HDF5.H5Error`
* Implement conversion from `AttributeDict{Any}` to `Dict{Symbol,Any}`
* Implement conversion from `AttributeDict{Any}` to `NamedTuple`


mkitti:
I factored out some of the unrelated parts into #956 and #957.

The main thing about this pull request is to rename `AttributeDict` to `AttributeDict{Any}` to facilitate extension of the dictionary interface.

mkitti:
The better way to go would be to create an `AttributeTypeDict` that is an `AbstractDict{String, Type}`. This in turn could be used to build dynamic structure like a `NamedTuple` where constant propagation would allow for type-stable access. The idea is to isolate the type-stability to a single call that provides information for subsequent type-stable access.

Issue #945, opened 2022-05-29
Split up and rearrange files


simonbyrne:
The main HDF5.jl file is rather large, and very difficult to navigate, especially since the API is not yet fully documented. I think this would be easier to develop and maintain if we were to split this up into smaller files. I originally tried this in #833, but it was quite a big change, so instead I decided to tackle it smaller chunks.

The question is what should go where. @mkitti suggested I write it up as an issue so that it can be discussed in one place.

My overall idea is to split it out according to the how the C API is broken up
- [X] Properties (`h5p`): done in #847
- [x] File functions (`h5f`): `File` type and `h5open`: #970
- [x] Group functions (`h5g`): `Group` type and `setindex!`/`getindex` overloads: #971
- [x] Datatype (`h5t`): `Datatype` and conversions between Julia types and HDF5 datatypes (#975)
- [x] Dataspace (`h5s`): `Dataspace` and conversions between Julia indexing and dataspaces: See #961
- [x] Dataset (`h5d`):  `Dataset` and `read`/`write` functionality: see #955.
- [x] Attributes (`h5a`): See #943
- [x] References (`h5r`): `Reference` type and dereferencing: see #968
- [x] Objects (`h5o`): See #976

PR #955, opened 2022-06-10
Move code into datasets.jl, readwrite.jl, and types.jl

mkitti:
This is my attempt to move dataset code into datasets.jl for #945 

I ended up needing three new files:

* types.jl - high level struct and union definitions that need to be defined early
* datasets.jl - dataset specific code
* readwrite.jl - read and write code common to datasets and attributes


src/types.jl
```diff
@@ -0,0 +1,187 @@
...
+Base.length(::Type{FixedString{N,PAD}}) where {N,PAD} = N
+Base.length(str::FixedString) = length(typeof(str))
+pad(::Type{FixedString{N,PAD}}) where {N,PAD} = PAD
+pad(x::T) where {T<:FixedString} = pad(T)
+
+struct VariableArray{T}
+    len::Csize_t
+    p::Ptr{Cvoid}
+end
+Base.eltype(::Type{VariableArray{T}}) where T = T
```
simonbyrne, line 186: I think this part should be moved to a different file: these aren't types corresponding to HDF5 objects, but are instead convenience wrappers matching certain types of data (in fact, I don't think some of them should exist, but that's another story)

simonbyrne: perhaps move this back to `typeconversions.jl`?

simonbyrne (approved):
Overall I think this structure makes the most sense. Defining HDF5 types up front will make some organization pieces easier.

src/types.jl
mkitti: Moved back to typeconversions.jl in 565fcaa

musm:
LGTM will merge later today sans objections

mkitti:
Looks good from my side as well, so merging. Let's move on to the next things.

PR #957, opened 2022-06-10
AttributeDict to NamedTuple convenience methods

mkitti:
This adds a few methods to convert an `AttributeDict` to a `Dict{Symbol, Any}`, which can then also be easily converted to a `NamedTuple`.


mkitti:
There is something wrong with the macOS nightly and `rand`.

mkitti:
Closing this. I think they is probably a route to make a NamedTuple like object.

PR #958, opened 2022-06-10
Fix attribute docs for attrs and rename_attribute, export it.

mkitti:
`rename_attribute` was not exported so Documenter could not find it's docstring.

The docstring for `attrs` got dissociated from the method.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -12,7 +12,7 @@ import Mmap
 export
 @read, @write,
 h5open, h5read, h5write, h5rewrite, h5writeattr, h5readattr,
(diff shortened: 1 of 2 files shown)


musm:
Doesn't the doc string just need to be updated to `HDF5.rename_attribute`, unless I'm missing something?

mkitti:
Is there a reason not to export it? It seems weird to have a series of `*_attribute` functions documented and exported and then have `rename_attribute` documented but not exported.

musm:
My understanding was that this wouldn't really be  used  since the new dict interface handles renames for you automatically.

simonbyrne:
as long as we export the other mid-level functions, it probably makes sense to export `rename_attribute`

PR #961, opened 2022-06-13
move dataspace-related functions to new file

simonbyrne:
Part of #945


musm:
NFC change sans the additional documentation (which is greatly welcome)?

musm (approved):

docs/src/dataspaces.md
```diff
@@ -0,0 +1,7 @@
+# Dataspaces
+
+```@docs
+Dataspace
```
musm, line 4:
```suggestion
HDF5.Dataspace
```
This might need to be prefixed with `HDF5.Dataspace` , since it's not exported?

docs/src/dataspaces.md
```diff
@@ -0,0 +1,7 @@
...
+dataspace
+isnull
```
musm, line 6:
```suggestion
HDF5.isnull
```

simonbyrne:
Failure on MacOS nightly appears to be
https://github.com/JuliaLang/julia/issues/45669

simonbyrne:
It may make sense to wait on #955 though, as that is a bigger change and will be harder to rebase.

PR #962, opened 2022-06-13
Update Project.toml

jw3126:
Can you tag a release? It would be pretty nice to be compatible with compat 4.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "HDF5"
 uuid = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
-version = "0.16.9"
+version = "0.16.10"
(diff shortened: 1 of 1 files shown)


mkitti:
What is compat 4?

mkitti:
I think we at least need #956 and #958 before we release

jw3126:
> What is compat 4?

What I meant was `Compat.jl` version 4. I had a situation where I installed a package that needs `Compat.jl` version 4 and HDF5.jl in a single environment. The result was that a very old version HDF5.jl got installed.

musm:
@mkitti  all the recent changes constitute a patch release as they are all backwards compatible and no deprecation of the attribute interface has been incorporated yet.

PR #966, opened 2022-06-14
move type-unwrapping to unsafe_convert

simonbyrne:
Should be safer

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -47,7 +47,8 @@ mutable struct File <: H5DataStore
         f
     end
 end
-Base.cconvert(::Type{API.hid_t}, f::File) = f.id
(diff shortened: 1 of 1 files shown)


mkitti:
Do we have a case where it is problematic?

simonbyrne:
I don't have a reproducer, but this is generally the safe way to unwrap objects: the way `cconvert` and `unsafe_convert` work is that `cconvert` can do any conversions that allocate memory, but you should use `unsafe_convert` to unwrap to get a pointer: another way to think of it is that in a `ccall`, Julia may invoke the GC after `cconvert`, but not after `unsafe_convert`.

mkitti:
That's a good point. Have our finalizers been trying to close handles they should not be?

mkitti (approved):

Issue #960, opened 2022-06-12
StackOverflowError during initialization of module Drivers


david-macmahon:
When HDF5 v0.16.7 (or later) is configured to use system HDF5 libraries, I get a StackOverflowError during the initialization of the Drivers module.  It doesn't happen on v0.16.6 (or earlier) and it doesn't happen if I use the bundled HDF5 libraries.

This is the error:

```julia
julia> using HDF5
ERROR: InitError: StackOverflowError:
Stacktrace:
 [1] h5fd_core_init()
   @ HDF5.API ~/.julia/packages/HDF5/I9NLZ/src/api/functions.jl:4209
 [2] __init__()
   @ HDF5.Drivers ~/.julia/packages/HDF5/I9NLZ/src/drivers/drivers.jl:88
 [3] _include_from_serialized(path::String, depmods::Vector{Any})
   @ Base ./loading.jl:696
 [4] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String)
   @ Base ./loading.jl:782
 [5] _require(pkg::Base.PkgId)
   @ Base ./loading.jl:1020
 [6] require(uuidkey::Base.PkgId)
   @ Base ./loading.jl:936
 [7] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:923
during initialization of module Drivers
```

mkitti:
Which system?

david-macmahon:
This is on Linux (Ubuntu 16.04) if that’s what you mean. It happens with both Julia 1.6.4 and 1.7.1.

david-macmahon:
But I’m using HDF5 1.13 (I think) libraries that I built from source.

david-macmahon:
My system libraries work fine with HDF5.jl v0.16.6 or earlier

mkitti:
HDF5 1.13.x is a development version. We cannot support that at this time.

I would file a bug with HDF5 group at  https://github.com/HDFGroup/hdf5/issues indicating that calling `H5FDcore_init` causes a stack overflow.

We have initialized the core driver on `__init__` since 0.16.7 via https://github.com/JuliaIO/HDF5.jl/pull/928 .

mkitti:
I did some work and identified that driver initialization in HDF5 1.13 has changed. See
https://github.com/HDFGroup/hdf5/issues/1809

mkitti:
Technically we are using a non-public API because the public API is the H5FD_CORE macro.

[...]

Issue #959, opened 2022-06-11
Use of H5P_SET_FILE_LOCKING env var is unreliable


mkitti:
We should use https://portal.hdfgroup.org/display/HDF5/H5P_SET_FILE_LOCKING rather than setting the environment variable `HDF5_USE_FILE_LOCKING`. Environment variable communication with the HDF5 library can be inconsistent, especially on Windows.

https://github.com/JuliaIO/HDF5.jl/blob/3a805bfddedfdac742161bc115323fcaf0d4e77a/src/HDF5.jl#L1710-L1712

Overall this indicates the need for configurable default access property lists.

simonbyrne:
You can also use `wputenv`, which is what we use for RCall.jl (where I first came across the issue):
https://github.com/JuliaInterop/RCall.jl/blob/2da13513856f7d82cff9c5ce8c6cb99c21c3631e/src/setup.jl#L74-L75

mkitti:
This is still unreliable. You have to ensure you are calling into the same C library,  which might not be the case during interop.

mkitti:
See https://github.com/JuliaLang/julia/issues/44054

PR #967, opened 2022-06-15
Avoid calling non-public h5fd_xxx_init() functions

david-macmahon:
The h5fd_xxx_init() functions are not part of the HDF5 public API.  A
method of retrieving the hid_t value of drivers that uses
public/supported API calls was suggested in a comment on a related
github issue and is implemented in this commit.  For more details see:

https://github.com/HDFGroup/hdf5/issues/1809#issuecomment-1154397046

### DIFF
--- a/src/drivers/drivers.jl
+++ b/src/drivers/drivers.jl
@@ -96,7 +96,20 @@ function __init__()
     API.h5p_set_fapl_sec2(fapl)
     DRIVERS[API.h5p_get_driver(fapl)] = POSIX
     close(fapl)
-    @require MPI="da04e1cc-30fd-572f-bb4f-1f8673147195" include("mpio.jl")
+
+    # Check whether the loaded HDF5 libraries were compiled with parallel support.
+    loaded_libs = filter(contains(r"hdf5"i), Libc.Libdl.dllist())
+    for libname in loaded_libs
+        has_mpio = Libc.Libdl.dlopen(libname) do lib
+            Libc.Libdl.dlsym(lib, :H5Pset_fapl_mpio; throw_error=false) !== nothing
+        end
+        if has_mpio
+            HDF5.HAS_PARALLEL[] = true
+            break
+        end
+    end
+
+    @require MPI="da04e1cc-30fd-572f-bb4f-1f8673147195" (HDF5.has_parallel() && include("mpio.jl"))
 end
 
 end # module
--- a/src/drivers/drivers.jl
+++ b/src/drivers/drivers.jl
@@ -85,8 +85,12 @@ function set_driver!(p::Properties, ::POSIX)
 end
 
 function __init__()
-    DRIVERS[API.h5fd_core_init()] = Core
-    DRIVERS[API.h5fd_sec2_init()] = POSIX
+    fapl = HDF5.FileAccessProperties(fclose_degree=:default)
+    API.h5p_set_fapl_core(fapl, 4096, false)
+    DRIVERS[API.h5p_get_driver(fapl)] = Core
+    API.h5p_set_fapl_sec2(fapl)
+    DRIVERS[API.h5p_get_driver(fapl)] = POSIX
+    close(fapl)
     @require MPI="da04e1cc-30fd-572f-bb4f-1f8673147195" include("mpio.jl")
 end
 
--- a/src/drivers/drivers.jl
+++ b/src/drivers/drivers.jl
@@ -5,6 +5,7 @@ export POSIX
 import ..API
 import ..HDF5: HDF5, Properties, h5doc
 
+using Libdl: dllist, dlopen, dlsym
 using Requires: @require
 
 
@@ -98,10 +99,10 @@ function __init__()
     close(fapl)
 
     # Check whether the loaded HDF5 libraries were compiled with parallel support.
-    loaded_libs = filter(s->match(r"hdf5"i, s) !== nothing, Libc.Libdl.dllist())
+    loaded_libs = filter(s->match(r"hdf5"i, s) !== nothing, dllist())
     for libname in loaded_libs
-        has_mpio = Libc.Libdl.dlopen(libname) do lib
-            Libc.Libdl.dlsym(lib, :H5Pset_fapl_mpio; throw_error=false) !== nothing
+        has_mpio = dlopen(libname) do lib
+            dlsym(lib, :H5Pset_fapl_mpio; throw_error=false) !== nothing
         end
         if has_mpio
             HDF5.HAS_PARALLEL[] = true
--- a/src/drivers/drivers.jl
+++ b/src/drivers/drivers.jl
@@ -85,6 +85,11 @@ function set_driver!(p::Properties, ::POSIX)
 end
 
 function __init__()
+    # disable file locking as that can cause problems with mmap'ing
+    if !haskey(ENV, "HDF5_USE_FILE_LOCKING")
+        ENV["HDF5_USE_FILE_LOCKING"] = "FALSE"
+    end
+
     fapl = HDF5.init!(HDF5.FileAccessProperties())
     API.h5p_set_fapl_core(fapl, 4096, false)
     DRIVERS[API.h5p_get_driver(fapl)] = Core
--- a/src/drivers/drivers.jl
+++ b/src/drivers/drivers.jl
@@ -85,7 +85,7 @@ function set_driver!(p::Properties, ::POSIX)
 end
 
 function __init__()
-    fapl = HDF5.FileAccessProperties(fclose_degree=:default)
+    fapl = HDF5.init!(HDF5.FileAccessProperties())
     API.h5p_set_fapl_core(fapl, 4096, false)
     DRIVERS[API.h5p_get_driver(fapl)] = Core
     API.h5p_set_fapl_sec2(fapl)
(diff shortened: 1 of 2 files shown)


david-macmahon:
Fixes #960

david-macmahon:
OK, something weird is happening to the `mmap.jl` test.  Even if I revert this commit and just add the two lines marked with `# NEW` to the existing `__init__()` function, then the mmap test breaks.

```julia
fapl = HDF5.FileAccessProperties(fclose_degree=:default) # NEW
close(fapl) # NEW
DRIVERS[API.h5fd_core_init()] = Core # EXISTING
DRIVERS[API.h5fd_sec2_init()] = POSIX # EXISTING
```
I don't understand why that would break the `mmap.jl` test.

david-macmahon:
Even putting the `# NEW` lines at the end of `HDF5.Drivers.__init__()` breaks the `mmap.jl` test:

```julia
function __init__()
    DRIVERS[API.h5fd_core_init()] = Core
    DRIVERS[API.h5fd_sec2_init()] = POSIX
    @require MPI="da04e1cc-30fd-572f-bb4f-1f8673147195" include("mpio.jl")
    fapl = HDF5.init!(HDF5.FileAccessProperties()) # NEW
    close(fapl) # NEW
end
```
Is there some other part of the HDF5 libraries that need to be initialized before creating a file access property list?

mkitti:
Note that there is also the MPIO driver: https://github.com/JuliaIO/HDF5.jl/blob/cfa5a2401e168396ad05750c24a3446a43cf9b25/src/drivers/mpio.jl#L47

mkitti:
My suspicion is that you accidentally initialized the HDF5 library before we could disable file locking in `HDF5.__init__()`.
https://github.com/JuliaIO/HDF5.jl/blob/cfa5a2401e168396ad05750c24a3446a43cf9b25/src/HDF5.jl#L559-L561

This is why I pointed out the potential problem in https://github.com/JuliaIO/HDF5.jl/issues/959.

david-macmahon:
Thanks for the pointer to the MPIO driver.  And you are exactly right about the file locking messing with the `mmap` test.  I had just found that out, but you beat me to it!

mkitti:
I see two paths forward:
1. Create a new method to initialize `DRIVERS` and call that at the end of `HDF5.__init__()`.
2. Make this lazy, and set the key for `DRIVERS` when someone invokes `set_driver!`.

david-macmahon:
OK, yeah, I think the lazy approach is the way to go.  I tried adding similar `fapl` logic to the MPIO driver using this:

```julia
# Check whether the HDF5 libraries were compiled with parallel support.
try
    fapl = HDF5.init!(HDF5.FileAccessProperties())
    comm = info = reinterpret(H5MPIHandle, 0)
    API.h5p_set_fapl_mpio(fapl, comm, info)
    DRIVERS[API.h5p_get_driver(fapl)] = MPIO
    close(fapl)
    HDF5.HAS_PARALLEL[] = true
catch e
    @show e
end
```
but that resulted in an MPI error: "The MPI_Comm_dup() function was called before MPI_INIT was invoked."  It's obviously best to let the user call MPI_INIT, so the non-lazy approach seems fatally flawed in this case.

david-macmahon:
One issue with the lazy approach is that a `FileAccessProperties` object  can be created without calling `set_driver!`, but `h5p_get_driver` will still return a driver (i.e. SEC2/POSIX) for it.  So maybe the `DRIVERS` entries for the sec2 and core drivers could be set non-lazily, but then set the entry for mpio lazily in `set_driver!(fapl, mpio::MPIO)`.  The problem with lazy initializing the mpio entry is that `mpio.jl` currently uses a non-throwing call to `h5fd_mpio_init()` as the way to detect that `HAS_PARALLEL` should be set to `true` (i.e. only sets it to `true` if the function call does not throw an exception).  If we no longer call `h5fd_mpio_init()` then we will need some other way to determine whether the HDF5 libraries have parallel support.

mkitti:
Perhaps the only driver we need to initialize is driver is the default one. That may be platform dependent. This may need to be done in `HDF5.__init__`.

mkitti:
You probably should just do `using Libdl`.

src/drivers/drivers.jl
```diff
@@ -85,9 +85,31 @@ function set_driver!(p::Properties, ::POSIX)
...
+    loaded_libs = filter(s->match(r"hdf5"i, s) !== nothing, Libc.Libdl.dllist())
+    for libname in loaded_libs
+        has_mpio = Libc.Libdl.dlopen(libname) do lib
+            Libc.Libdl.dlsym(lib, :H5Pset_fapl_mpio; throw_error=false) !== nothing
+        end
+        if has_mpio
+            HDF5.HAS_PARALLEL[] = true
+            break
+        end
+    end
```
mkitti, line 110:
```suggestion
    HDF5.HAS_PARALLEL[] = Libdl.dlsym(API.HDF5_jll.libhdf5_handle, :H5Pset_fapl_mpio; throw_error=false) !== nothing
```

Add `using Libdl` above.

simonbyrne: I don't think this will work with a custom HDF5 library either

simonbyrne:
> Why are we not using the standard JLL override mechanism?

The original Overrides.toml approach (https://docs.binarybuilder.org/stable/jll/#Overriding-the-artifacts-in-JLL-packages) is way too restrictive to be useful (it requires that the libraries have the same .so version as the one in the jll)

The Preferences.jl approach (https://docs.binarybuilder.org/stable/jll/#Overriding-specific-products) should work though

src/drivers/drivers.jl
```diff
@@ -85,9 +86,24 @@ function set_driver!(p::Properties, ::POSIX)
...
-    DRIVERS[API.h5fd_sec2_init()] = POSIX
-    @require MPI="da04e1cc-30fd-572f-bb4f-1f8673147195" include("mpio.jl")
+    # disable file locking as that can cause problems with mmap'ing
+    if !haskey(ENV, "HDF5_USE_FILE_LOCKING")
+        ENV["HDF5_USE_FILE_LOCKING"] = "FALSE"
+    end
+
+    fapl = HDF5.init!(HDF5.FileAccessProperties())
+    API.h5p_set_fapl_core(fapl, 4096, false)
+    DRIVERS[API.h5p_get_driver(fapl)] = Core
```
simonbyrne, line 96: we could lazily initialize `Core` as well?

david-macmahon: This way seemed closer to the original, so it's a slightly smaller change, but I don't think there would be a problem lazily initializing Core as well if that's the consensus.

mkitti: Perhaps all the `set_driver!` methods should uniformly  register their hid_t in `DRIVERS`. From this `__init__`, we just call `set_driver!(fapl, POSIX())`.

mkitti:
I would also be willing to merge this as is and revise later.

mkitti (approved):

simonbyrne (approved):

src/drivers/drivers.jl
```diff
@@ -85,9 +86,24 @@ function set_driver!(p::Properties, ::POSIX)
 end
 
 function __init__()
-    DRIVERS[API.h5fd_core_init()] = Core
-    DRIVERS[API.h5fd_sec2_init()] = POSIX
-    @require MPI="da04e1cc-30fd-572f-bb4f-1f8673147195" include("mpio.jl")
+    # disable file locking as that can cause problems with mmap'ing
+    if !haskey(ENV, "HDF5_USE_FILE_LOCKING")
```
musm, line 90: What's the reason we need to add this check here as well? Is it related to `Drivers` being in its own module?

musm:
For future work, I think we should also try to get a build with HDF5 1.13  in our CI.

src/drivers/drivers.jl
mkitti: What we really need to do is address #959. Basically what happens is that the `Drivers` module is initialized before the `HDF5` module. However, if we create a FAPL during `Drivers` initialization, then the environment variable no longer has the desired effect. Then memory mapping breaks before the files are now locked.

musm: Thanks that makes sense.

mkitti:
I would like to merge this and setup another pull request for revisions.

mkitti:
> > 
> 
> The original Overrides.toml approach (https://docs.binarybuilder.org/stable/jll/#Overriding-the-artifacts-in-JLL-packages) is way too restrictive to be useful (it requires that the libraries have the same .so version as the one in the jll)
> 
> The Preferences.jl approach (https://docs.binarybuilder.org/stable/jll/#Overriding-specific-products) should work though

Perhaps we should deprecate deps.jl and recommend the Preferences.jl approach?

src/drivers/drivers.jl
```diff
@@ -85,9 +86,24 @@ function set_driver!(p::Properties, ::POSIX)
...
+        ENV["HDF5_USE_FILE_LOCKING"] = "FALSE"
+    end
+
+    fapl = HDF5.init!(HDF5.FileAccessProperties())
+    API.h5p_set_fapl_core(fapl, 4096, false)
```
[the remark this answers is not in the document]
david-macmahon: Nowhere in particular (other than typical Linux page size).  TBH, I suspect this could be set to 0 here because it's only used to setup fapl, which in turn is only used to get the ID of the Core driver.  This fapl is never used to actually create anything.  FWIW, I think this value defaults to 8192 elsewhere in HDF5.jl, but it's not clear where that value comes from.

PR #972, opened 2022-06-18
Make Core init lazy, revise FileAccessProperty interface, fix file locking

### LINKED ISSUE #959: Use of H5P_SET_FILE_LOCKING env var is unreliable

mkitti:
* Fix #959 by using `h5p_set_file_locking(false, true)` by default via `HDF5.init!(fapl::FileAccessProperties)`
* Add `h5p_get_file_locking` helper, integrate with property interface
* Make `HDF5.Drivers.Core()` registration in `DRIVERS` lazy
* Implement `do` syntax for `Properties`
* Moves some initialization to `API.__init__()` which always executes first.
* Exposes a `const Ref` `libhdf5handle`
* Implements `API._has_symbol`

Revises #967


src/api/api.jl
```diff
@@ -16,4 +16,24 @@ include("error.jl")
...
 include("helpers.jl")
 
+function __init__()
+    # HDF5.API.__init__() is run before HDF5.__init__()
+
+    # From deps.jl
+    check_deps()
+
+    # Ensure this is reinitialized on using
+    libhdf5handle[] = Libdl.dlopen(libhdf5)
```
musm, line 26: This is needed to ensure the handle doesn't change between precompilation ?

src/api/helpers.jl
```diff
@@ -285,6 +285,30 @@ function h5f_get_free_sections(file_id, type, sect_info::AbstractVector{H5F_sect
...
+   return _has_h5p_set_file_locking
+   #=
+   h5_version = h5_get_libversion()
+   if (h5_version >= v"1.10" && h5_version < v"1.10.7") ||
+      (h5_version >= v"1.12" && h5_version < v"1.12.1") ||
+      (h5_version < v"1.10")
+      return false
+   else
+      return true
+   end
```
musm, line 308: The commented out lines are an alternative way to check this, but you opted for the `_has_symbol` approach instead?

musm:
Are there any significant  runtime differences with the lazy approach?

mkitti:
> Are there any significant  runtime differences with the lazy approach? 

HDF5.jl should load faster.

I suppose if anyone were to create a loop over setting the driver, it might be a little slower. The workaround would be to call the low level API directly.

src/api/helpers.jl
mkitti: Yes. We could do it either by version or looking up the symbol. Symbols lookup seems more direct, but the version based mechanism may still be informative.

src/api/api.jl
musm: > In the future, we should just grab the handle from HDF5_jll.

How would that that actually reinitialize the library? It seems like you'd have to call `dlopen` here regardless

mkitti:
HDF_jll already uses `dlopen` for us via JLLWrappers.jl:
https://github.com/JuliaPackaging/JLLWrappers.jl/blob/bade7d4b7af3e1c0c590bfd383a3df5ed10852f8/src/products/library_generators.jl#L54

```
julia> HDF5.API.HDF5_jll.libhdf5_handle
Ptr{Nothing} @0x00000000009df580
```

musm:
Ah nvm,  I misunderstood what you meant...

PR #974, opened 2022-06-21
Remove confusing example of `create_dataset`

nhz2:
Based on the example:

https://github.com/JuliaIO/HDF5.jl/blob/d6ecf42d035784f78616a24b03cda7c037eda7c9/docs/src/index.md?plain=1#L89-L99

I thought `create_dataset` would write the data, but `create_dataset` doesn't actually write any data, it just creates the dataset.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -78,32 +78,41 @@ close(fid)
 
 Closing a file also closes any other open objects (e.g., datasets, groups) in that file. In general, you need to close an HDF5 file to "release" it for use by other applications.
 
-## Writing a group or dataset
+## Creating a group or dataset
 
-Groups can be created as follows:
+Groups can be created via the function [`create_group`](@ref)
 
 ```@repl main
 create_group(fid, "mygroup")
 ```
 
-We can write the `"mydataset"` by:
+We can write the `"mydataset"` by indexing into `fid`. This also happens to write data to the dataset.
 
 ```@repl main
-fid["mydataset"] = rand(10)
+fid["mydataset"] = rand()
 ```
 
-Writing to a dataset to a group is as simple as:
+Alternatively, we can call [`create_dataset`](@ref), which does not write data to the dataset. It merely creates the dataset.
+
+```@repl main
+create_dataset(fid, "myvector", Int, (10,))
+```
+
+Creating a dataset within a group is as simple as indexing into the group with the name of the dataset or calling [`create_dataset`](@ref) with the group as the first argument.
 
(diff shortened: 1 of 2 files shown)


mkitti:
How about we show how to use `write_dataset` ?

nhz2:
I think this is described in the "Reading and writing data" section.
https://github.com/JuliaIO/HDF5.jl/blob/d6ecf42d035784f78616a24b03cda7c037eda7c9/docs/src/index.md?plain=1#L153-L158

It is also described in the "Mid-level routines" section.

https://github.com/JuliaIO/HDF5.jl/blob/d6ecf42d035784f78616a24b03cda7c037eda7c9/docs/src/index.md?plain=1#L463-L468

musm:
Not sure how removing this helps clarity. It  seems like the paragraph above is clear on how to 'create' and also how to 'write' datasets. 

Perhaps instead you could add more explanatory text on where you think the prose is lacking?

mkitti:
He's right in that the second call to `create_dataset` does not actually write any data.

image

I think it would clearer if we called this section "Creating a group or dataset":

## Creating a group or dataset

Groups can be created via the function `create_group`.

```@repl main
create_group(fid, "mygroup")
```

We can create `"mydataset"` by indexing into `fid`. This also happens to write data to the dataset.

```@repl main
fid["mydataset"] = rand()
```

Alternatively, we can call `create_dataset`, which does _not_ write data to the dataset. It merely creates the dataset.

```@repl main
create_dataset(fid, "myvector", rand(10))
```

Creating a dataset within a group is as simple as indexing into the group with the name of the dataset or calling `create_dataset` with the group as the first argument.

```@repl main
g = fid["mygroup"]
g["mydataset"] = "Hello World!"
create_dataset(g, "myvector", Int, (10,))
```

The `do` syntax is also supported. The file, group, and dataset handles will automatically be closed after the `do` block terminates.

```@repl main
h5open("example2.h5", "w") do fid
    g = create_group(fid, "mygroup")
    create_dataset(g, "myvector", Int, (10,))
end
```

mkitti:
I changed all the instances to `create_dataset(fid, "myvector", Int, (10,))` now exemplifying the form giving the datatype and dataspace dimensions.

mkitti (approved):
This looks good to me. I suggest we consider deprecating the `create_dataset` form in _another_ pull request that takes an array instance but does not write. After deprecation, we could then consider whether to reinstate it along the lines of the h5py interface.

PR #976, opened 2022-06-26
move object-related functions to new file

simonbyrne:
Last item in #945.


mkitti:
I think you forgot to include objects.jl

src/HDF5.jl
```diff
@@ -115,45 +115,6 @@ function h5read(filename, name::AbstractString, indices::Tuple{Vararg{Union{Abst
 end
```
mkitti, line 117:
```suggestion
include("objects.jl")
```

mkitti:
In #979, I'm proposing to deprecate `object_info`.

PR #978, opened 2022-06-27
Add low level bindings for h5o*

mkitti:
This adds low level bindings for the H5O interface. There is more work to do with `h5o_get_info`. For now, `const H5O_info_t = H5O_info1_t`.

This should be non-breaking.


mkitti:
Closing this in favor the more advanced #979 , reopen if we want a slower approach.

PR #979, opened 2022-06-27
h5o_get_info update to h5o_get_info3, deprecate object_info

mkitti:
Possibly supersedes #978 .

Besides expanding coverage of H5O functions as in #978, this pull requests:
* Implements `HDF5.API.h5o_get_info1` as a legacy replacement for the former `HDF5.API.h5o_get_info`
* Uses versions bounds for API functions.
* Deprecates the exported function `object_info` with implementation based on `HDF5.API.h5o_get_info1`
* Implements `num_attrs` as a high level interface to access the `num_attrs` field of the `H5O_info{N}_t` struct, compatible with versions `N = 1` and `N = 2`.
* `Attributes` and `AttributesDict` now use `num_attrs` rather than `object_info` (cc: @simonbyrne )

`object_info` should either reimplemented using `HDF5.API.h5o_get_info` in the future or high level access for individual fields should be wrapped such as with `num_attrs`. It might be a bad idea to directly expose such as a low-level structure via the high-level API.


src/attributes.jl
```diff
@@ -180,6 +180,21 @@ function h5readattr(filename, name::AbstractString)
...
+Retrieve the number of attributes from an object.
+
+See [`API.h5o_get_info`](@ref).
+"""
+function num_attrs(obj)
+    info = @static if API._libhdf5_build_ver < v"1.12.0"
+        API.h5o_get_info(checkvalid(obj))
+    else
+        API.h5o_get_info(checkvalid(obj), API.H5O_INFO_NUM_ATTRS)
+    end
```
mkitti, line 195: For pre v1.12.0 HDF5, this uses `H5Oget_info1`. For post v1.12.0, this uses `H5Oget_info3`.

mkitti:
I deconflicted this branch with the recent merges.

src/api/types.jl
```diff
@@ -263,10 +349,19 @@ const H5L_TYPE_HARD    = 0
 const H5L_TYPE_SOFT    = 1
 const H5L_TYPE_EXTERNAL= 2
 
+# H5O_INFO constants
+const H5O_INFO_BASIC = Cuint(0x0001)
+const H5O_INFO_TIME = Cuint(0x0002)
+const H5O_INFO_NUM_ATTRS = Cuint(0x0004)
+const H5O_INFO_HDR = Cuint(0x0008)
+const H5O_INFO_META_SIZE = Cuint(0x0010)
+const H5O_INFO_ALL = H5O_INFO_BASIC | H5O_INFO_TIME | H5O_INFO_NUM_ATTRS | H5O_INFO_HDR | H5O_INFO_META_SIZE
```
musm, line 358: Shouldn't this just be 
```
H5O_INFO_BASIC | H5O_INFO_TIME | H5O_INFO_NUM_ATTRS
```

 as per 
https://github.com/HDFGroup/hdf5/blob/develop/src/H5Opublic.h#L89

seems like  you also folded in 
https://github.com/HDFGroup/hdf5/blob/develop/src/H5Opublic.h#L98

src/attributes.jl
```diff
@@ -180,6 +180,21 @@ function h5readattr(filename, name::AbstractString)
...
 
+"""
+    num_attrs()
+
+Retrieve the number of attributes from an object.
+
+See [`API.h5o_get_info`](@ref).
+"""
+function num_attrs(obj)
+    info = @static if API._libhdf5_build_ver < v"1.12.0"
```
musm, line 191: Out of curiosity does using `num_attrs` provide any speed benefits vs.  the existing check on `object_info`?

src/api/helpers.jl
```diff
@@ -427,9 +427,107 @@ end
...
+# Legacy h5o_get_info1 interface, for compat with pre-1.12.0
+# Used by deprecated object_info function
+"""
+    h5o_get_info1(object_id, [buf])
+
+Deprecated HDF5 function. Use [`h5o_get_info`](@ref) or [`h5o_get_native_info`](@ref) if possible.
+
+See `libhdf5` documentation for [`H5Oget_info1`](https://portal.hdfgroup.org/display/HDF5/H5O_GET_INFO1).
+"""
+function h5o_get_info1(object_id, buf)
```
musm, line 523: Shouldn't these have `depwarns` ?

src/api/types.jl
mkitti: It depends on which version of the library and function we are using:

`H5O_GET_INFO2` and `H5O_GET_INFO3` says `H5O_INFO_ALL` should be `H5O_INFO_BASIC | H5O_INFO_TIME | H5O_INFO_NUM_ATTRS | H5O_INFO_HDR | H5O_INFO_META_SIZE`

https://portal.hdfgroup.org/display/HDF5/H5O_GET_INFO2
https://portal.hdfgroup.org/display/HDF5/H5O_GET_INFO3

https://github.com/HDFGroup/hdf5/blob/1d90890a7b38834074169ce56720b7ea7f4b01ae/src/H5Opublic.h#L91

src/api/helpers.jl
musm: Nevermind, I see what you did here

src/attributes.jl
mkitti: Read through this forum thread:
https://forum.hdfgroup.org/t/h5oget-info-performance-with-large-number-of-chunks/2093/6

Basically `H5Oget_info` pulls a lot of information that we do not need which can be quite slow in certain circumstances. They resolved this by changing to the API so you can ask exactly for the data that you need, in this case the number of attributes.

musm: Makes sense

musm:
```
julia> @btime API.h5o_get_info1(h5f["lore"]).num_attrs
  5.840 μs (5 allocations: 416 bytes)
0x0000000000000001

julia> @btime API.h5o_get_info(h5f["lore"], API.H5O_INFO_NUM_ATTRS).num_attrs
  4.757 μs (5 allocations: 224 bytes)
0x0000000000000001

```
Defintely allocates a lot less even with small info structs.

Issue #939, opened 2022-05-17
`track_order` only used with `FileIO` & `OrderedDict`


lmh91:
It would be nice if the internal tracking property of the order would always be taken into account.
Especially in `display` and `keys`.

```julia
using HDF5
using OrderedCollections, FileIO
fn = "test_tracked_order.h5"

h5open(fn, "w"; track_order=true) do io
    fcpl = HDF5.get_create_properties(io)
    @info fcpl.track_order
    io["b"] = 1
    io["a"] = 2
    g = create_group(io, "G"; track_order=true)
    gcpl = HDF5.get_create_properties(io["G"])
    @info gcpl.track_order
    write(g, "z", 3)
    write(g, "f", 4)
end

load(fn; dict=OrderedDict()) # This uses the order
```

`load` returns an ordered dict:
```julia
OrderedDict{Any, Any} with 4 entries:
  "b"   => 1
  "a"   => 2
  "G/z" => 3
  "G/f" => 4
```

However, `display` and `keys` ignore the `track_order` keyword:
```julia
HDF5.h5open(fn, "r"; track_order=true) do f
    display(f) # This does not uses the order
    keys(f)    # This does not uses the order
end
```
Output: 
```julia
🗂️ HDF5.File: (read-only) test_tracked_order.h5
├─ 📂 G
│  ├─ 🔢 f
│  └─ 🔢 z
├─ 🔢 a
└─ 🔢 b
3-element Vector{String}:
 "G"
 "a"
 "b"
```

PR #982, opened 2022-07-05
Fix `track_order` usage in do block

t-bltg:
Follow up of https://github.com/JuliaIO/HDF5.jl/pull/912.

- fix https://github.com/JuliaIO/HDF5.jl/issues/939;
- fix https://github.com/JuliaIO/HDF5.jl/issues/985.

Remove `IDX_TYPE` and `ORDER` globals (non thread safe global states).

~~And a little cleanup / simplifications~~.

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -41,8 +41,8 @@ Filters, Drivers
 h5doc(name) = "[`$name`](https://portal.hdfgroup.org/display/HDF5/$(name))"
 
 include("api/api.jl")
-
 include("properties.jl")
+include("context.jl")
 include("types.jl")
 include("file.jl")
 include("objects.jl")
@@ -154,15 +154,11 @@ function Base.setindex!(parent::Union{File,Group}, val, path::Union{AbstractStri
     write(parent, path, val; pv...)
 end
 
-
 # end of high-level interface
-
 
 include("api_midlevel.jl")
 
-
 #API.h5s_get_simple_extent_ndims(space_id::API.hid_t) = API.h5s_get_simple_extent_ndims(space_id, C_NULL, C_NULL)
-
 
 # Functions that require special handling
 
@@ -172,8 +168,8 @@ const libversion = API.h5_get_libversion()
 get_access_properties(d::Dataset)   = DatasetAccessProperties(API.h5d_get_access_plist(d))
 get_access_properties(f::File)      = FileAccessProperties(API.h5f_get_access_plist(f))
 get_create_properties(d::Dataset)   = DatasetCreateProperties(API.h5d_get_create_plist(d))
-get_create_properties(g::Group)     = isvalid(g.gcpl) ? g.gcpl : GroupCreateProperties(API.h5g_get_create_plist(g))
-get_create_properties(f::File)      = isvalid(f.fcpl) ? f.fcpl : FileCreateProperties(API.h5f_get_create_plist(f))
+get_create_properties(g::Group)     = GroupCreateProperties(API.h5g_get_create_plist(g))
+get_create_properties(f::File)      = FileCreateProperties(API.h5f_get_create_plist(f))
 get_create_properties(a::Attribute) = AttributeCreateProperties(API.h5a_get_create_plist(a))
 
 
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -42,9 +42,6 @@ h5doc(name) = "[`$name`](https://portal.hdfgroup.org/display/HDF5/$(name))"
 
 include("api/api.jl")
 
-const IDX_TYPE = Ref(API.H5_INDEX_NAME)
-const ORDER = Ref(API.H5_ITER_INC)
-
 include("properties.jl")
 include("types.jl")
 include("file.jl")
@@ -175,8 +172,8 @@ const libversion = API.h5_get_libversion()
 get_access_properties(d::Dataset)   = DatasetAccessProperties(API.h5d_get_access_plist(d))
 get_access_properties(f::File)      = FileAccessProperties(API.h5f_get_access_plist(f))
 get_create_properties(d::Dataset)   = DatasetCreateProperties(API.h5d_get_create_plist(d))
-get_create_properties(g::Group)     = GroupCreateProperties(API.h5g_get_create_plist(g))
-get_create_properties(f::File)      = FileCreateProperties(API.h5f_get_create_plist(f))
+get_create_properties(g::Group)     = isvalid(g.gcpl) ? g.gcpl : GroupCreateProperties(API.h5g_get_create_plist(g))
+get_create_properties(f::File)      = isvalid(f.fcpl) ? f.fcpl : FileCreateProperties(API.h5f_get_create_plist(f))
 get_create_properties(a::Attribute) = AttributeCreateProperties(API.h5a_get_create_plist(a))
 
 
--- a/src/file.jl
+++ b/src/file.jl
@@ -42,7 +42,7 @@ function h5open(filename::AbstractString, mode::AbstractString, fapl::FileAccess
         end
         fid = API.h5f_open(filename, flag, fapl)
     end
-    File(fid, filename, fcpl)
+    return File(fid, filename)
 end
 
 
@@ -59,7 +59,7 @@ function h5open(filename::AbstractString, mode::AbstractString = "r";
         return h5open(filename, mode, fapl, fcpl; swmr=swmr)
     finally
         close(fapl)
-        # close(fcpl)  # FIXME: need to remain valid in read mode
+        close(fcpl)
     end
 end
 
@@ -75,12 +75,18 @@ For example with a `do` block:
     end
 
 """
-function h5open(f::Function, args...; pv...)
+function h5open(f::Function, args...; context = copy(CONTEXT), pv...)
     file = h5open(args...; pv...)
-    try
-        f(file)
-    finally
-        close(file)
+    task_local_storage(:hdf5_context, context) do
+        if (track_order = get(pv, :track_order, nothing)) !== nothing
+            context.file_create.track_order = context.group_create.track_order = track_order
+        end
+        try
+            f(file)  # the function body can access `context` via `get_context_property`
+        finally
+            close(file)
+            close(context)
+        end
     end
 end
 
--- a/src/file.jl
+++ b/src/file.jl
@@ -42,7 +42,7 @@ function h5open(filename::AbstractString, mode::AbstractString, fapl::FileAccess
         end
         fid = API.h5f_open(filename, flag, fapl)
     end
-    return File(fid, filename)
+    File(fid, filename, fcpl)
 end
 
 
@@ -56,28 +56,27 @@ function h5open(filename::AbstractString, mode::AbstractString = "r";
     try
         pv = setproperties!(fapl, fcpl; pv...)
         isempty(pv) || error("invalid keyword options $pv")
-        file = h5open(filename, mode, fapl, fcpl; swmr=swmr)
-        return file
+        return h5open(filename, mode, fapl, fcpl; swmr=swmr)
     finally
         close(fapl)
-        close(fcpl)
+        # close(fcpl)  # FIXME: need to remain valid in read mode
     end
 end
 
-
 """
-    function h5open(f::Function, args...; swmr=false, pv...)
+    function h5open(f::Function, args...; pv...)
 
 Apply the function f to the result of `h5open(args...; kwargs...)` and close the resulting
-`HDF5.File` upon completion. For example with a `do` block:
+`HDF5.File` upon completion.
+For example with a `do` block:
 
     h5open("foo.h5","w") do h5
         h5["foo"]=[1,2,3]
     end
 
 """
-function h5open(f::Function, args...; swmr=false, pv...)
-    file = h5open(args...; swmr=swmr, pv...)
+function h5open(f::Function, args...; pv...)
+    file = h5open(args...; pv...)
     try
         f(file)
     finally
--- a/src/objects.jl
+++ b/src/objects.jl
@@ -19,13 +19,30 @@ function Base.close(obj::Union{Group,Dataset})
     nothing
 end
 
+inherit_gcpl(parent) = GroupCreateProperties()
+inherit_gcpl(parent::Group) = isvalid(parent.gcpl) ? parent.gcpl : GroupCreateProperties()
+function inherit_gcpl(parent::File)
+    if isvalid(parent.fcpl)
+        # FIXME: propagate ∩(GroupCreateProperties, FileCreateProperties) properties
+        GroupCreateProperties(track_order = parent.fcpl.track_order)
+    else
+        GroupCreateProperties()
+    end
+end
+
 # Object (group, named datatype, or dataset) open
 function h5object(obj_id::API.hid_t, parent)
     obj_type = API.h5i_get_type(obj_id)
-    obj_type == API.H5I_GROUP ? Group(obj_id, file(parent)) :
-    obj_type == API.H5I_DATATYPE ? Datatype(obj_id, file(parent)) :
-    obj_type == API.H5I_DATASET ? Dataset(obj_id, file(parent)) :
-    error("Invalid object type for path ", path)
+    obj = if obj_type == API.H5I_GROUP
+        Group(obj_id, file(parent), inherit_gcpl(parent))
+    elseif obj_type == API.H5I_DATATYPE
+        Datatype(obj_id, file(parent))
+    elseif obj_type == API.H5I_DATASET
+        Dataset(obj_id, file(parent))
+    else
+        error("Invalid object type for path ", path)
+    end
+    obj
 end
 open_object(parent, path::AbstractString) = h5object(API.h5o_open(checkvalid(parent), path, API.H5P_DEFAULT), parent)
 function gettype(parent, path::AbstractString)
--- a/src/objects.jl
+++ b/src/objects.jl
@@ -19,32 +19,17 @@ function Base.close(obj::Union{Group,Dataset})
     nothing
 end
 
-inherit_gcpl(parent) = GroupCreateProperties()
-inherit_gcpl(parent::Group) = isvalid(parent.gcpl) ? parent.gcpl : GroupCreateProperties()
-function inherit_gcpl(parent::File)
-    if isvalid(parent.fcpl)
-        # FIXME: propagate ∩(GroupCreateProperties, FileCreateProperties) properties
-        GroupCreateProperties(track_order = parent.fcpl.track_order)
-    else
-        GroupCreateProperties()
-    end
-end
-
 # Object (group, named datatype, or dataset) open
 function h5object(obj_id::API.hid_t, parent)
     obj_type = API.h5i_get_type(obj_id)
-    obj = if obj_type == API.H5I_GROUP
-        Group(obj_id, file(parent), inherit_gcpl(parent))
-    elseif obj_type == API.H5I_DATATYPE
-        Datatype(obj_id, file(parent))
-    elseif obj_type == API.H5I_DATASET
-        Dataset(obj_id, file(parent))
-    else
-        error("Invalid object type for path ", path)
-    end
-    obj
+    obj_type == API.H5I_GROUP ? Group(obj_id, file(parent)) :
+    obj_type == API.H5I_DATATYPE ? Datatype(obj_id, file(parent)) :
+    obj_type == API.H5I_DATASET ? Dataset(obj_id, file(parent)) :
+    error("Invalid object type for path ", path)
 end
+
 open_object(parent, path::AbstractString) = h5object(API.h5o_open(checkvalid(parent), path, API.H5P_DEFAULT), parent)
+
 function gettype(parent, path::AbstractString)
     obj_id = API.h5o_open(checkvalid(parent), path, API.H5P_DEFAULT)
     obj_type = API.h5i_get_type(obj_id)
--- a/src/properties.jl
+++ b/src/properties.jl
@@ -272,7 +272,7 @@ end
 get_track_order(p::Properties) = API.h5p_get_link_creation_order(p) != 0 && API.h5p_get_attr_creation_order(p) != 0
 
 function set_track_order!(p::Properties, val::Bool)
-    crt_order_flags = val ? (HDF5.API.H5P_CRT_ORDER_TRACKED | HDF5.API.H5P_CRT_ORDER_INDEXED) : 0
+    crt_order_flags = val ? (API.H5P_CRT_ORDER_TRACKED | API.H5P_CRT_ORDER_INDEXED) : 0
     API.h5p_set_link_creation_order(p, crt_order_flags)
     API.h5p_set_attr_creation_order(p, crt_order_flags)
     nothing
--- a/src/types.jl
+++ b/src/types.jl
@@ -38,9 +38,10 @@ end
 mutable struct File <: H5DataStore
     id::API.hid_t
     filename::String
+    fcpl::FileCreateProperties
 
-    function File(id, filename, toclose::Bool=true)
-        f = new(id, filename)
+    function File(id, filename, fcpl::FileCreateProperties=FileCreateProperties(), toclose::Bool=true)
+        f = new(id, filename, fcpl)
         if toclose
             finalizer(close, f)
         end
@@ -53,9 +54,10 @@ Base.unsafe_convert(::Type{API.hid_t}, f::File) = f.id
 mutable struct Group <: H5DataStore
     id::API.hid_t
     file::File         # the parent file
+    gcpl::GroupCreateProperties
 
-    function Group(id, file)
-        g = new(id, file)
+    function Group(id, file, gcpl::GroupCreateProperties=GroupCreateProperties())
+        g = new(id, file, gcpl)
         finalizer(close, g)
         g
     end
@@ -138,3 +140,9 @@ const ScalarType = Union{BitsType,Reference}
 
 # Define an H5O Object type
 const Object = Union{Group,Dataset,Datatype}
+
+idx_type(obj::Union{File,Group}) = get_track_order(get_create_properties(obj)) ? API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
+idx_type(obj::Any) = API.H5_INDEX_NAME
+
+# TODO: implement alternative iteration order ?
+order(obj::Any) = API.H5_ITER_INC
--- a/src/types.jl
+++ b/src/types.jl
@@ -38,10 +38,9 @@ end
 mutable struct File <: H5DataStore
     id::API.hid_t
     filename::String
-    fcpl::FileCreateProperties
 
-    function File(id, filename, fcpl::FileCreateProperties=FileCreateProperties(), toclose::Bool=true)
-        f = new(id, filename, fcpl)
+    function File(id, filename, toclose::Bool=true)
+        f = new(id, filename)
         if toclose
             finalizer(close, f)
         end
@@ -54,10 +53,9 @@ Base.unsafe_convert(::Type{API.hid_t}, f::File) = f.id
 mutable struct Group <: H5DataStore
     id::API.hid_t
     file::File         # the parent file
-    gcpl::GroupCreateProperties
 
-    function Group(id, file, gcpl::GroupCreateProperties=GroupCreateProperties())
-        g = new(id, file, gcpl)
+    function Group(id, file)
+        g = new(id, file)
         finalizer(close, g)
         g
     end
@@ -141,8 +139,15 @@ const ScalarType = Union{BitsType,Reference}
 # Define an H5O Object type
 const Object = Union{Group,Dataset,Datatype}
 
-idx_type(obj::Union{File,Group}) = get_track_order(get_create_properties(obj)) ? API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
-idx_type(obj::Any) = API.H5_INDEX_NAME
+idx_type(obj::File) =
+    get_context_property(:file_create).track_order ||
+    get_create_properties(obj).track_order ? 
+    API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
+idx_type(obj::Group) =
+    get_context_property(:group_create).track_order ||
+    get_create_properties(obj).track_order ? 
+    API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
+idx_type(obj) = API.H5_INDEX_NAME
 
 # TODO: implement alternative iteration order ?
-order(obj::Any) = API.H5_ITER_INC
+order(obj) = API.H5_ITER_INC
--- a/test/fileio.jl
+++ b/test/fileio.jl
@@ -53,26 +53,27 @@ let fn = tempname() * ".h5"
 
   # issue #939
   h5open(fn, "r"; track_order=true) do io
-    @test HDF5.get_create_properties(io).track_order
+    @test HDF5.get_context_property(:file_create).track_order
     @test all(keys(io) .== ["b", "a", "G"])
-    @test HDF5.get_create_properties(io["G"]).track_order
+    @test HDF5.get_context_property(:group_create).track_order
+    @test HDF5.get_create_properties(io["G"]).track_order  # inferred from file, created with `track_order=true`
     @test all(keys(io["G"]) .== ["z", "f"])
   end
 
   h5open(fn, "r"; track_order=false) do io
-    @test !HDF5.get_create_properties(io).track_order
+    @test !HDF5.get_context_property(:file_create).track_order
     @test all(keys(io) .== ["G", "a", "b"])
-    @test !HDF5.get_create_properties(io["G"]).track_order
-    @test all(keys(io["G"]) .== ["f", "z"])
+    @test !HDF5.get_context_property(:group_create).track_order
+    @test HDF5.get_create_properties(io["G"]).track_order  # inferred from file
+    @test all(keys(io["G"]) .== ["z", "f"])
   end
 
   h5open(fn, "r") do io
     @test !HDF5.get_create_properties(io).track_order
     @test all(keys(io) .== ["G", "a", "b"])
-    @test HDF5.get_create_properties(io["G"]).track_order  # inferred from file, created with `track_order=true`
+    @test HDF5.get_create_properties(io["G"]).track_order  # inferred from file
     @test all(keys(io["G"]) .== ["z", "f"])
   end
-
 end
 
 let fn = tempname() * ".h5"
--- a/test/fileio.jl
+++ b/test/fileio.jl
@@ -50,6 +50,29 @@ let fn = tempname() * ".h5"
   dat = load(fn; dict=OrderedDict())  # `track_order` is inferred from `OrderedDict`
 
   @test all(keys(dat) .== ["b", "a", "G/z", "G/f"])
+
+  # issue #939
+  h5open(fn, "r"; track_order=true) do io
+    @test HDF5.get_create_properties(io).track_order
+    @test all(keys(io) .== ["b", "a", "G"])
+    @test HDF5.get_create_properties(io["G"]).track_order
+    @test all(keys(io["G"]) .== ["z", "f"])
+  end
+
+  h5open(fn, "r"; track_order=false) do io
+    @test !HDF5.get_create_properties(io).track_order
+    @test all(keys(io) .== ["G", "a", "b"])
+    @test !HDF5.get_create_properties(io["G"]).track_order
+    @test all(keys(io["G"]) .== ["f", "z"])
+  end
+
+  h5open(fn, "r") do io
+    @test !HDF5.get_create_properties(io).track_order
+    @test all(keys(io) .== ["G", "a", "b"])
+    @test HDF5.get_create_properties(io["G"]).track_order  # inferred from file, created with `track_order=true`
+    @test all(keys(io["G"]) .== ["z", "f"])
+  end
+
 end
 
 let fn = tempname() * ".h5"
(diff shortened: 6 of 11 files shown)


test/runtests.jl
```diff
@@ -12,51 +13,27 @@ Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zzstd")))
 # To debug HDF5.jl tests, uncomment the next line
 # ENV["JULIA_DEBUG"] = "Main"
 
-@testset "HDF5.jl" begin
-
-@debug "plain"
 include("plain.jl")
```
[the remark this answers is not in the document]
t-bltg: The same can be achieved with the removal of the `@testset "HDF5.jl"` block, so that specific `Test Summary: | Pass  Broken  Total` appear when using `include`. I believe that the top level `@testset` doesn't add anything and has the drawback of hiding the test sequence. With the removal you also track down hangs: https://github.com/JuliaIO/HDF5.jl/runs/7194199530?check_suite_focus=true#step:5:159.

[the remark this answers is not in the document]
mkitti: Yes, I understand. However, we only really wanted it enabled during debugging and for CI. The multiprocess setup of `Pkg.test` also creates some unusual interference. Let's discuss in another pull request, please.

src/file.jl
```diff
@@ -1,5 +1,5 @@
 """
-    h5open(filename::AbstractString, mode::AbstractString="r"; swmr=false, pv...)
+    h5open(filename::AbstractString, mode::AbstractString = "r"; swmr = false, pv...)
```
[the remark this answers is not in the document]
t-bltg: Already removed after https://github.com/JuliaIO/HDF5.jl/pull/983#issuecomment-1175011022.

src/file.jl
```diff
@@ -64,24 +64,37 @@ function h5open(filename::AbstractString, mode::AbstractString = "r";
     end
 end
 
+_change_iteration_order(track_order::Bool) = if track_order
```
mkitti, line 67: This formatting is a bit weird. Just use the long form function syntax here so the indentation helps indicate where the function begins and ends.

I'm also not clear why it was necessary to move the definition. The order of method definitions should not matter.

mkitti (changes requested):
Did the method definitions need to move?

src/file.jl
t-bltg: > Oh, I see that it was formatted like that before.

I'll change that to long def.

> Anyways, is this change actually necessary?

Yes, because `src/fileio.jl` is conditionally loaded by `Requires.jl`: https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/src/HDF5.jl#L208.

t-bltg:
> Did the method definitions need to move?

Yes, because `src/fileio.jl` is conditionally loaded by `Requires.jl`: https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/src/HDF5.jl#L208.

test/fileio.jl
```diff
@@ -50,6 +50,12 @@ let fn = tempname() * ".h5"
   dat = load(fn; dict=OrderedDict())  # `track_order` is inferred from `OrderedDict`
 
   @test all(keys(dat) .== ["b", "a", "G/z", "G/f"])
+
+  # issue #939
+  h5open(fn, "r"; track_order=true) do io
+    # display(io)
+    @test all(keys(io) .== ["b", "a", "G"])
+  end
```
mkitti, line 58: Test that the order actually reverts after the `do` block.

mkitti:
We may eventually want to move `IDX_TYPE` into `task_local_storage`, but that is outside the scope of this pull request.

test/fileio.jl
t-bltg: Will do.

t-bltg:
> We may eventually want to move IDX_TYPE into task_local_storage, but that is outside the scope of this pull request.

Not needed, I'm reworking the pr to remove unsafe globals `IDX_TYPE` and `ORDER`.

mkitti:
> Not needed, I'm reworking the pr to remove unsafe globals `IDX_TYPE` and `ORDER`.

Honestly, I think this is pretty close to mergeable. Perhaps this second refactor should be another pull request?

Here, I'm mainly waiting on someone else, perhaps @musm to review before merging.

mkitti (dismissed):
This is close to mergeable. The one thing that is missing is checking the reversion of `IDX_TYPE` after the `do` syntax.

mkitti:
`track_order` seems to conflate attribute order and link order. However, a dataset only has attribute order.

Thus you need to have a distinct implementation of the track_order accessors for dataset creation property lists.

This makes me wonder what happens when link track order and attribute track order differ.

src/types.jl
```diff
@@ -38,9 +38,10 @@ end
 mutable struct File <: H5DataStore
     id::API.hid_t
     filename::String
+    track_order::Bool
 
-    function File(id, filename, toclose::Bool=true)
-        f = new(id, filename)
+    function File(id, filename, fcpl::FileCreateProperties=FileCreateProperties(), toclose::Bool=true)
```
mkitti, line 43: Since we are getting a file id as an argument, shouldn't we get the `FileCreateProperty` from the file id via `H5Fget_create_plist`?

https://docs.hdfgroup.org/hdf5/develop/group___h5_f.html#ga2f823a9e929b00b06a6be80619a61778

Also why do we need to add a `track_order` property to the File struct?  Should it basically be a property of the root group?

t-bltg:
> Thus you need to have a distinct implementation of the track_order accessors for dataset creation property lists.

Yes, or support `track_order` only for `Group`s and `File`s.

I'm currently struggling with the property lists for `Group`s and `File`s **when reading** since `get_create_properties(...)` through `h5g_get_create_plist` or `h5f_get_create_plist` doesn't seem to return the correct values.
Currently, `fcpl` is used only upon writing a new file: https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/src/file.jl#L35
thus the `track_order` property of `fcpl` is lost when reading.

It doesn't seem to be possible to modify the property list after the call to `h56f_open` here: https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/src/file.jl#L43
This is why I've added a `fcpl` or `gcpl` named field to the `File` and `Group` structs and try to work with that now.

What remains is a failing test which occurs `h5object` is used for iterating the `File` objects:
https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/src/objects.jl#L25

src/types.jl
t-bltg: See the comments, below this commit was obsolete, I've updated the PR.

t-bltg:
The tests are now passing, although I'm not happy with the implementation which might be brittle.
An alternate view on this PR would help to see if we can make it bulletproof.

This would be breaking for `JLD.jl` (need to fix https://github.com/JuliaIO/JLD.jl/blob/14e66d98e1e31d1694ff7acf4139d5add72e4fa3/src/JLD.jl#L186).

mkitti:
If we broke the API for JLD, then we probably broke the API for someone else. This would not be mergeable until we decide to release v0.17 at least.

mkitti:
If we broke the API for JLD, then we probably broke the API for someone else. This would not be mergeable until we decide to release v0.17 at least.

mkitti:
My suggestion is to recover the commit where I said this was almost mergeable and submit that as another pull request. Merging this version is going to take some time.

t-bltg:
> If we broke the API for JLD, then we probably broke the API for someone else.

I can make this non-breakable by moving the positional argument after `toclose`, but that would be really odd since `toclose` is not a member of `File` and `fcpl` is ...

t-bltg:
> My suggestion is to recover the commit where I said this was almost mergeable and submit that as another pull request.

I honestly think that https://github.com/JuliaIO/HDF5.jl/pull/982/commits/1669dd3c708f64d44d256cbc14bd3597937e8648 is not viable either: I'd rather remove `IDX_TYPE` and `ORDER`.

> Merging this version is going to take some time.

I'd rather wait and polish than merge something wobbly.

mkitti (commented):
Thanks for attempting to address this.

The introduction of a mutable create property list for `File` and `Group` for use post-creation will require a further high-level design and planning review. The API modifications are breaking at multiple points and will thus need a minor release bump. My approval is withdrawn for now.

t-bltg:
> We'll have to discuss whether this removal alone is breaking.

Technically no, because `IDX_TYPE` and `ORDER` were never exported. I introduced them in https://github.com/JuliaIO/HDF5.jl/pull/912 in order to bypass fixed arguments of `h5..._iterate` as it was hard-coded previously, e.g.: https://github.com/JuliaIO/HDF5.jl/blob/6e8fc9242d12be7e2be5fb5407b9d272b9fedee9/src/show.jl#L211
This was an improvement at the time (over hard-coded arguments), but we might in the future need to change the indexing type and/or the order in a more robust and flexible manner.

> Additionally, I think the semantics of how track_order works is fundamentally changing here.

They need to, because https://github.com/JuliaIO/HDF5.jl/pull/912 was incomplete, as shown by https://github.com/JuliaIO/HDF5.jl/issues/939.

> There are other global variables that need to be addressed as well

Do you have an example ?

mkitti:
> Do you have an example ?

We basically need a structure that looks like this that contains property lists for everything. These will then serve as templates for the default property lists created by other functions. We can have a global one, and then perhaps provide a mechanism to provide a more local one.

```julia
struct HDF5Context
   attribute_access::HDF5.AttributeAccessProperties
   attribute_create::HDF5.AttributeCreateProperties
   dataset_access  ::HDF5.DatasetAccessProperties
   dataset_create  ::HDF5.DatasetCreateProperties
   dataset_transfer::HDF5.DatasetTransferProperties
   datatype_access ::HDF5.DatatypeAccessProperties
   datatype_create ::HDF5.DatatypeCreateProperties
   file_access     ::HDF5.FileAccessProperties
   file_create     ::HDF5.FileCreateProperties
   file_mount      ::HDF5.FileMountProperties
   group_access    ::HDF5.GroupAccessProperties
   group_create    ::HDF5.GroupCreateProperties
   link_access     ::HDF5.LinkAccessProperties
   link_create     ::HDF5.LinkCreateProperties
   object_copy     ::HDF5.ObjectCopyProperties
   object_create   ::HDF5.ObjectCreateProperties
   string_create   ::HDF5.StringCreateProperties
end

function HDF5Context()
   HDF5Context(
       HDF5.AttributeAccessProperties(),
       HDF5.AttributeCreateProperties(),
       HDF5.DatasetAccessProperties(),
       HDF5.DatasetCreateProperties(),
       HDF5.DatasetTransferProperties(),
       HDF5.DatatypeAccessProperties(),
       HDF5.DatatypeCreateProperties(),
       HDF5.FileAccessProperties(),
       HDF5.FileCreateProperties(),
       HDF5.FileMountProperties(),
       HDF5.GroupAccessProperties(),
       HDF5.GroupCreateProperties(),
       HDF5.LinkAccessProperties(),
       HDF5.LinkCreateProperties(),
       HDF5.ObjectCopyProperties(),
       HDF5.ObjectCreateProperties(),
       HDF5.StringCreateProperties()
   )
end
````

We could then use `task_local_storage` to store and retrieve this context. Functions would then use `H5Pcopy` to get their property lists. We could also just reference these default template property lists for `track_order`.

```julia
julia> context.file_create.track_order
false

julia> context.file_create.track_order = true
true

julia> task_local_storage(:hdf5_context, context) do
           println(task_local_storage(:hdf5_context).file_create.track_order)
       end
true
```

t-bltg:
> We basically need a structure that looks like this that contains property lists for everything. These will then serve as templates for the default property lists created by other functions. We can have a global one, and then perhaps provide a mechanism to provide a more local one.

Thanks for the example. It still feels a bit odd to use `task_local_storage`, but it seems to work and pass the tests.

mkitti (changes requested):
Thanks for implementing the context. Here's some feedback.

Ultimately, we'll need someone else to review a well.

src/HDF5.jl
```diff
@@ -41,11 +41,10 @@ Filters, Drivers
 h5doc(name) = "[`$name`](https://portal.hdfgroup.org/display/HDF5/$(name))"
 
 include("api/api.jl")
+include("properties.jl")
 
-const IDX_TYPE = Ref(API.H5_INDEX_NAME)
-const ORDER = Ref(API.H5_ITER_INC)
+const CONTEXT = HDF5Context()
```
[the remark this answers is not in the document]
t-bltg: Moved to `properties.jl`.

src/properties.jl
```diff
@@ -270,7 +270,7 @@ end
 
 get_track_order(p::Properties) = API.h5p_get_link_creation_order(p) != 0 && API.h5p_get_attr_creation_order(p) != 0
 
-function set_track_order(p::Properties, val::Bool)
+function set_track_order!(p::Properties, val::Bool)
```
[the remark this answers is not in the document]
t-bltg: Why ? These are unexported. I've renamed it to `set_track_order!` for consistency with e.g. `set_obj_track_times!`.

src/HDF5.jl
```diff
@@ -157,28 +156,28 @@ function Base.setindex!(parent::Union{File,Group}, val, path::Union{AbstractStri
...
+    local_context = get(task_local_storage(), :hdf5_context, CONTEXT)
+    property = getfield(local_context, name)
+    isvalid(property) ? property : nothing
+end
+get_access_properties(d::Dataset)   = something(get_global_property(:dataset_access), DatasetAccessProperties(API.h5d_get_access_plist(d)))
+get_access_properties(f::File)      = something(get_global_property(:file_access), FileAccessProperties(API.h5f_get_access_plist(f)))
+get_create_properties(d::Dataset)   = something(get_global_property(:dataset_create), DatasetCreateProperties(API.h5d_get_create_plist(d)))
+get_create_properties(g::Group)     = something(get_global_property(:group_create), GroupCreateProperties(API.h5g_get_create_plist(g)))
+get_create_properties(f::File)      = something(get_global_property(:file_create), FileCreateProperties(API.h5f_get_create_plist(f)))
+get_create_properties(a::Attribute) = something(get_global_property(:attribute_create), AttributeCreateProperties(API.h5a_get_create_plist(a)))
```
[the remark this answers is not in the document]
t-bltg: > I only want the global property if the object (Dataset, File, etc) is not valid.

Mmh no, this will make `track_order` fail since `API.h5f_get_create_plist(f)` returns a `fpcl` with incorrect `track_order`, although valid fpcl. A valid global context property should preempt object properties.

src/properties.jl
```diff
@@ -874,3 +874,45 @@ const UTF8_ATTRIBUTE_PROPERTIES = AttributeCreateProperties()
...
+   file_create     ::HDF5.FileCreateProperties
+   file_mount      ::HDF5.FileMountProperties
+   group_access    ::HDF5.GroupAccessProperties
+   group_create    ::HDF5.GroupCreateProperties
+   link_access     ::HDF5.LinkAccessProperties
+   link_create     ::HDF5.LinkCreateProperties
+   object_copy     ::HDF5.ObjectCopyProperties
+   object_create   ::HDF5.ObjectCreateProperties
+   string_create   ::HDF5.StringCreateProperties
+end
```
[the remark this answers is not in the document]
t-bltg: No, see https://github.com/JuliaIO/HDF5.jl/pull/982#discussion_r918227986.

src/file.jl
```diff
@@ -56,32 +56,39 @@ function h5open(filename::AbstractString, mode::AbstractString = "r";
...
+        end
+        try
+            f(file)
+        finally
+            close(file)
+            close(ctx.file_create)
+            close(ctx.group_create)
+        end
     end
 end
```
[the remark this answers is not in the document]
t-bltg: Ok fine, but I have no idea how to implement the `Properties` copy.

src/file.jl
```diff
@@ -42,7 +42,7 @@ function h5open(filename::AbstractString, mode::AbstractString, fapl::FileAccess
         end
         fid = API.h5f_open(filename, flag, fapl)
     end
-    return File(fid, filename)
+    File(fid, filename)
```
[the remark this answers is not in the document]
t-bltg: They are superfluous, but fine.

src/HDF5.jl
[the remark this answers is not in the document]
mkitti: No. I would just copy that line into gen/api_defs.jl, run `julia --project=. gen/gen_wrappers.jl`, and then build a high level interface to copy property lists. I will probably add a bunch of API defs in another pull request if you just would like to wait for that.

src/properties.jl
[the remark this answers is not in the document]
mkitti: `export` does not define what is valid API or not. Non-exported functions could be public API if documented. That's why a package such as [PublicAPI.jl](https://docs.juliahub.com/PublicAPI/Jfs06/0.1.0/) exists.

While this is not documented, I would lean towards using `@deprecate` when possible. I agree this should change, but I also think we should deprecate should since `set_track_order` has existed across several releases.

src/HDF5.jl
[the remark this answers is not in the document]
mkitti: > Mmh no, this will make `track_order` fail since `API.h5f_get_create_plist(f)` returns a `fpcl` with incorrect `track_order`, although valid opened `fpcl`. A valid global context property should preempt object properties.

Restore the old methods then and create a new family of methods. `get_access_from_context` or `get_context_create` or `get_context_access` perhaps?

mkitti:
I resolved a few line conflicts in 9585bd3.

src/properties.jl
```diff
@@ -874,3 +874,52 @@ const UTF8_ATTRIBUTE_PROPERTIES = AttributeCreateProperties()
 
 _link_properties(::AbstractString) = UTF8_LINK_PROPERTIES
 _attr_properties(::AbstractString) = UTF8_ATTRIBUTE_PROPERTIES
+
+struct HDF5Context
```
musm, line 878: HDF5 prefix below is unnecessary

src/properties.jl
```diff
@@ -874,3 +874,52 @@ const UTF8_ATTRIBUTE_PROPERTIES = AttributeCreateProperties()
...
+       HDF5.GroupCreateProperties(),
+       HDF5.LinkAccessProperties(),
+       HDF5.LinkCreateProperties(),
+       HDF5.ObjectCopyProperties(),
+       HDF5.ObjectCreateProperties(),
+       HDF5.StringCreateProperties(),
+   )
+end
+
+function Base.copy(x::HDF5Context)
```
musm, line 920: Is the `copy` method ever called in the current implementation, if so it seems like this wouldn't work without and PR  https://github.com/JuliaIO/HDF5.jl/pull/995 is needed for a correct implementation   (cc @mkitti )

src/file.jl
```diff
@@ -56,32 +56,39 @@ function h5open(filename::AbstractString, mode::AbstractString = "r";
...
-function h5open(f::Function, args...; swmr=false, pv...)
-    file = h5open(args...; swmr=swmr, pv...)
-    try
-        f(file)
-    finally
-        close(file)
+function h5open(f::Function, args...; context = copy(CONTEXT), pv...)
+    file = h5open(args...; pv...)
+    task_local_storage(:hdf5_context, context) do
+        ctx = task_local_storage(:hdf5_context)
```
musm, line 81: This is kinda confusing why do you need to use `ctx` instead of the passed in `context` ?

src/types.jl
```diff
@@ -138,3 +138,9 @@ const ScalarType = Union{BitsType,Reference}
 
 # Define an H5O Object type
 const Object = Union{Group,Dataset,Datatype}
+
+idx_type(obj::Union{File,Group}) = get_track_order(get_create_properties(obj)) ? API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
+idx_type(obj::Any) = API.H5_INDEX_NAME
```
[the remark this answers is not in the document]
t-bltg: No, not needed. I'm afk today so I will rework this pr tomorrow.

test/fileio.jl
```diff
@@ -36,20 +36,41 @@ end
 
 let fn = tempname() * ".h5"
   h5open(fn, "w"; track_order=true) do io
-    fcpl = HDF5.get_create_properties(io)
-    @test fcpl.track_order
+    @test HDF5.get_create_properties(io).track_order
```
musm, line 39: is this a NFC (non functional change)?

musm:
Generally looks good, still reviewing some of the PR. 

Could you give a brief summary of  how the task local changes are an improvement and operate?

One crude way to probe the value of the order would be to have global const refs, but I could easily imagine those values getting garbled with repeated calls in independent threads, even if things don't work 100% in multi-threaded contexts yet when using HDF5

mkitti:
The task local storage (TLS) is actually my idea, so I'll explain it. The (TLS) is basically a global namespace specific to each `Task`. If you start a new `Task`, you have a new `TLS`.

In our case, we need to temporarily change the global space for some fixed duration. To do so we can launch a new `Task` and then populate the TLS with the modified "global" context. Functions running within the `Task` will see the modified variables. These changes will be undone upon completion of the `Task`. 

Specifically in our case those local variables are property lists and their properties. In this PR, we are mainly thinking about index order. In the future, I think we should use these property lists as a template when creating new property lists. When we need a new property list, we copy it from the local context rather than creating one _de novo_.

This is better than using a global constant `Ref` to store state in that the changes are never seen outside the `Task`. Although multithreading does not currently apply here, this also happens to be compatible with that case.

I have a working draft here of what I think this should look like:

https://github.com/mkitti/HDF5.jl/blob/be385088adef7d1a83b994b0b140e75b530f7272/src/context.jl

test/fileio.jl
t-bltg: Rationale: consistency with the test written after. But I won't argue over a 2 lines change. Restored.

src/file.jl
t-bltg: If I understood `tls` correctly, this construction allow access of the context from withing a task (private). Anyway, this was a request of @mkitti to use tls so I will let him handle it.

src/properties.jl
t-bltg: Rebased against master for https://github.com/JuliaIO/HDF5.jl/pull/995.

src/HDF5.jl
```diff
@@ -157,27 +153,28 @@ function Base.setindex!(parent::Union{File,Group}, val, path::Union{AbstractStri
...
+    local_context = get(task_local_storage(), :hdf5_context, CONTEXT)
+    property = getfield(local_context, name)
+    isvalid(property) ? property : nothing
+end
+get_access_properties(d::Dataset)   = something(get_context_property(:dataset_access), DatasetAccessProperties(API.h5d_get_access_plist(d)))
+get_access_properties(f::File)      = something(get_context_property(:file_access), FileAccessProperties(API.h5f_get_access_plist(f)))
+get_create_properties(d::Dataset)   = something(get_context_property(:dataset_create), DatasetCreateProperties(API.h5d_get_create_plist(d)))
+get_create_properties(g::Group)     = something(get_context_property(:group_create), GroupCreateProperties(API.h5g_get_create_plist(g)))
+get_create_properties(f::File)      = something(get_context_property(:file_create), FileCreateProperties(API.h5f_get_create_plist(f)))
+get_create_properties(a::Attribute) = something(get_context_property(:attribute_create), AttributeCreateProperties(API.h5a_get_create_plist(a)))
```
mkitti, line 177:
```suggestion
get_access_properties(d::Dataset)   = DatasetAccessProperties(API.h5d_get_access_plist(d))
get_access_properties(f::File)      = FileAccessProperties(API.h5f_get_access_plist(f))
get_create_properties(d::Dataset)   = DatasetCreateProperties(API.h5d_get_create_plist(d))
get_create_properties(g::Group)     = GroupCreateProperties(API.h5g_get_create_plist(g))
get_create_properties(f::File)      = FileCreateProperties(API.h5f_get_create_plist(f))
get_create_properties(a::Attribute) = AttributeCreateProperties(API.h5a_get_create_plist(a))
```

Revert this. Just call `get_context_property` directly if we are interested in getting a property from the context. If we want to choose between object and context properties, that should be further downstream on a per-property basis.

src/file.jl
```diff
@@ -56,32 +56,38 @@ function h5open(filename::AbstractString, mode::AbstractString = "r";
...
+        if (track_order = get(pv, :track_order, nothing)) !== nothing
+            ctx.file_create.track_order = ctx.group_create.track_order = track_order
+        end
+        try
+            f(file)
+        finally
+            close(file)
+            close(ctx)
+        end
```
mkitti, line 90:
```suggestion
function h5open(f::Function, args...; context = copy(CONTEXT), pv...)
    file = h5open(args...; pv...)
    task_local_storage(:hdf5_context, context) do
        if (track_order = get(pv, :track_order, nothing)) !== nothing
            context.file_create.track_order = track_order
            context.group_create.track_order = track_order
        end
        try
            # The function body can access `context` via `get_context_property`
            f(file)
        finally
            close(file)
            close(context)
        end
```

src/types.jl
```diff
@@ -138,3 +138,9 @@ const ScalarType = Union{BitsType,Reference}
 
 # Define an H5O Object type
 const Object = Union{Group,Dataset,Datatype}
+
+idx_type(obj::Union{File,Group}) = get_track_order(get_create_properties(obj)) ? API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
```
mkitti, line 142:
```suggestion
idx_type(obj::File) = get_context_property(:file_create).track_order ? API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
idx_type(obj::Group) = get_context_property(:group_create).track_order ? API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
```

src/objects.jl
```diff
@@ -22,10 +22,16 @@ end
...
+    obj = if obj_type == API.H5I_GROUP
+        Group(obj_id, file(parent))
+    elseif obj_type == API.H5I_DATATYPE
+        Datatype(obj_id, file(parent))
+    elseif obj_type == API.H5I_DATASET
+        Dataset(obj_id, file(parent))
+    else
+        error("Invalid object type for path ", path)
+    end
+    obj
```
mkitti, line 34:
```suggestion
    obj_type == API.H5I_GROUP ? Group(obj_id, file(parent)) :
    obj_type == API.H5I_DATATYPE ? Datatype(obj_id, file(parent)) :
    obj_type == API.H5I_DATASET ? Dataset(obj_id, file(parent)) :
    error("Invalid object type for path ", path)
```

This is an unrelated stylistic change. If we are going to do this, we should do this globally rather than just right here. properties.jl uses the existing style a lot.

src/types.jl
mkitti: Perhaps we should always use `H5_INDEX_CRT_ORDER` if either the context or object property is enabled? Anyways, any fallback logic should be implemented here rather than `get_create_property` which has other uses.

```suggestion
idx_type(obj::File) =
    get_context_property(:file_create).track_order ||
    get_create_property(obj).track_order ? 
    API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
idx_type(obj::Group) =
    get_context_property(:group_create).track_order ||
    get_create_property(obj).track_order ? 
    API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
```

src/HDF5.jl
```diff
@@ -157,27 +153,28 @@ function Base.setindex!(parent::Union{File,Group}, val, path::Union{AbstractStri
...
-get_access_properties(d::Dataset)   = DatasetAccessProperties(API.h5d_get_access_plist(d))
-get_access_properties(f::File)      = FileAccessProperties(API.h5f_get_access_plist(f))
-get_create_properties(d::Dataset)   = DatasetCreateProperties(API.h5d_get_create_plist(d))
-get_create_properties(g::Group)     = GroupCreateProperties(API.h5g_get_create_plist(g))
-get_create_properties(f::File)      = FileCreateProperties(API.h5f_get_create_plist(f))
-get_create_properties(a::Attribute) = AttributeCreateProperties(API.h5a_get_create_plist(a))
+function get_context_property(name::Symbol)
+    local_context = get(task_local_storage(), :hdf5_context, CONTEXT)
+    property = getfield(local_context, name)
+    isvalid(property) ? property : nothing
```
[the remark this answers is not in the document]
mkitti: Perhaps we could do this in the future in another pull request?

```
get_create_properties(::Type{File}) = get_context_property(:file_create)
```

src/properties.jl
```diff
@@ -875,3 +875,53 @@ const UTF8_ATTRIBUTE_PROPERTIES = AttributeCreateProperties()
 
 _link_properties(::AbstractString) = UTF8_LINK_PROPERTIES
 _attr_properties(::AbstractString) = UTF8_ATTRIBUTE_PROPERTIES
+
+struct HDF5Context
```
mkitti, line 879: We should move the context stuff to it's own file, `context.jl`.

src/HDF5.jl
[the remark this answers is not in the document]
mkitti: `get_context_property` should move into `context.jl`.

test/fileio.jl
```diff
@@ -50,6 +50,29 @@ let fn = tempname() * ".h5"
   dat = load(fn; dict=OrderedDict())  # `track_order` is inferred from `OrderedDict`
 
   @test all(keys(dat) .== ["b", "a", "G/z", "G/f"])
+
+  # issue #939
+  h5open(fn, "r"; track_order=true) do io
+    @test HDF5.get_create_properties(io).track_order
```
mkitti, line 56:
```suggestion
    @test HDF5.get_context_property(:file_create).track_order
```

test/fileio.jl
```diff
@@ -50,6 +50,29 @@ let fn = tempname() * ".h5"
...
+    @test all(keys(io) .== ["b", "a", "G"])
+    @test HDF5.get_create_properties(io["G"]).track_order
```
mkitti, line 58:
```suggestion
    @test HDF5.get_context_property(:group_create).track_order
```

test/fileio.jl
```diff
@@ -50,6 +50,29 @@ let fn = tempname() * ".h5"
...
+    @test all(keys(io["G"]) .== ["z", "f"])
+  end
+
+  h5open(fn, "r"; track_order=false) do io
+    @test !HDF5.get_create_properties(io).track_order
```
mkitti, line 63:
```suggestion
    @test !HDF5.get_context_property(:file_create).track_order
```

test/fileio.jl
```diff
@@ -50,6 +50,29 @@ let fn = tempname() * ".h5"
...
+    @test all(keys(io) .== ["G", "a", "b"])
+    @test !HDF5.get_create_properties(io["G"]).track_order
```
mkitti, line 65:
```suggestion
    @test !HDF5.get_context_property(:group_create).track_order
```

test/fileio.jl
```diff
@@ -50,6 +50,29 @@ let fn = tempname() * ".h5"
...
+    @test all(keys(io["G"]) .== ["f", "z"])
+  end
+
+  h5open(fn, "r") do io
+    @test !HDF5.get_create_properties(io).track_order
```
mkitti, line 70:
```suggestion
    @test !HDF5.get_context_property(:file_create).track_order
```

test/fileio.jl
```diff
@@ -50,6 +50,29 @@ let fn = tempname() * ".h5"
...
+    @test all(keys(io) .== ["G", "a", "b"])
+    @test HDF5.get_create_properties(io["G"]).track_order  # inferred from file, created with `track_order=true`
```
mkitti, line 72:
```suggestion
    @test HDF5.get_context_property(:group_create).track_order  # inferred from file, created with `track_order=true`
```

test/fileio.jl
mkitti: This is clearer to me that the property is coming from the context rather than the object instance `io`.

mkitti (changes requested):
Merge master to get #993 if you have not already. #993 auto-initializes the property lists. Thus we do not need a fallback to the object's property list since the context will always have a valid property list on demand.

If someone wants to change the global index type or order, they can modify the global `CONTEXT`. Otherwise, they can use the `do` syntax for `h5open` to locally change the value.

src/objects.jl
t-bltg: Probably leftovers from developing. Restored.

src/properties.jl
t-bltg: Moved.

src/HDF5.jl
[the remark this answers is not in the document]
t-bltg: > After https://github.com/JuliaIO/HDF5.jl/pull/993 we do not need this. Property lists are now auto-initialized.

Did you try it ? Removing it fails the added tests.

> get_context_property should move into context.jl.

Moved.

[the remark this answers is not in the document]
mkitti: I also changed the tests. If you would like me to add the commit, I would be happy to do so.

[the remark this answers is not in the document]
t-bltg: Reworked.

t-bltg:
Looks like the failure on `MPI` built `hdf5` is spurious:

https://github.com/JuliaIO/HDF5.jl/runs/7544045615?check_suite_focus=true ✔
https://github.com/JuliaIO/HDF5.jl/runs/7544232871?check_suite_focus=true ✘

Even on failure, the outcome is a passing test.

mkitti:
```
  MethodError: no method matching h5o_get_info_by_name(::HDF5.File, ::String, ::Base.RefValue{HDF5.API.H5O_info1_t}, ::Int64)
```

That method error seems real though.

t-bltg:
Yes, but unrelated to this PR, since also present on master:
https://github.com/JuliaIO/HDF5.jl/runs/7429724952?check_suite_focus=true

mkitti:
I'm going to try to squash the test failure on this branch.

mkitti:
I fixed the tests here in [f5910a6](https://github.com/JuliaIO/HDF5.jl/pull/982/commits/f5910a6cf8c99b03737c4cdd47d9d39e7bec98ab) and in #1000

src/HDF5.jl
mkitti: This has not been resolved...

musm:
> The task local storage (TLS) is actually my idea, so I'll explain it. The (TLS) is basically a global namespace specific to each `Task`. If you start a new `Task`, you have a new `TLS`.
> 
> In our case, we need to temporarily change the global space for some fixed duration. To do so we can launch a new `Task` and then populate the TLS with the modified "global" context. Functions running within the `Task` will see the modified variables. These changes will be undone upon completion of the `Task`.
> 
> Specifically in our case those local variables are property lists and their properties. In this PR, we are mainly thinking about index order. In the future, I think we should use these property lists as a template when creating new property lists. When we need a new property list, we copy it from the local context rather than creating one _de novo_.
> 
> This is better than using a global constant `Ref` to store state in that the changes are never seen outside the `Task`. Although multithreading does not currently apply here, this also happens to be compatible with that case.
> 
> I have a working draft here of what I think this should look like:
> 
> https://github.com/mkitti/HDF5.jl/blob/be385088adef7d1a83b994b0b140e75b530f7272/src/context.jl

Do you plan on opening a PR after this is merged with  your working draft as I find  that those changes are important improvements?

musm:
@t-bltg  I'd recommend rebasing on master and cleaning up the commits. Then we can finally finalize merging in this PR. I'm not sure it would be best to squash all the commits or preserve the logical changes into several commits with more descriptive names. In either case a rebase would be helpful.

mkitti:
> Do you plan on opening a PR after this is merged with your working draft as I find that those changes are important improvements?

I would probably start another branch and send that in. I would also like to try implementing some of the applications. In particular, copying the context property lists to create default property lists for object creation.

For now, I've just documented this as an internal API that's under development.

mkitti:
I would just squash it.

mkitti (approved):

t-bltg:
> I fixed the tests here in [f5910a6](https://github.com/JuliaIO/HDF5.jl/pull/982/commits/f5910a6cf8c99b03737c4cdd47d9d39e7bec98ab) and in https://github.com/JuliaIO/HDF5.jl/pull/1000

Thanks for fixing.

> @t-bltg I'd recommend rebasing on master and cleaning up the commits.

Rebased.

src/types.jl
```diff
@@ -138,3 +138,16 @@ const ScalarType = Union{BitsType,Reference}
...
+    get_create_properties(obj).track_order ? 
+    API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
+idx_type(obj::Group) =
+    get_context_property(:group_create).track_order ||
+    get_create_properties(obj).track_order ? 
+    API.H5_INDEX_CRT_ORDER : API.H5_INDEX_NAME
+idx_type(obj) = API.H5_INDEX_NAME
+
+# TODO: implement alternative iteration order ?
+order(obj) = API.H5_ITER_INC
```
musm, line 153: I'd think a more descriptive name would be better. We might want to use `order` in the future for something else.

test/fileio.jl
```diff
@@ -50,6 +50,30 @@ let fn = tempname() * ".h5"
...
+  h5open(fn, "r"; track_order=false) do io
+    @test !HDF5.get_context_property(:file_create).track_order
+    @test all(keys(io) .== ["G", "a", "b"])
+    @test !HDF5.get_context_property(:group_create).track_order
+    @test HDF5.get_create_properties(io["G"]).track_order  # inferred from file
+    @test all(keys(io["G"]) .== ["z", "f"])
+  end
+
+  h5open(fn, "r") do io
+    @test !HDF5.get_create_properties(io).track_order
```
musm, line 72: If track order is false can't the check on line 73 fail since the keys are unordered? Wouldn't an equality check using `Set` be safer?

musm:
Thanks for sticking to this and sorry for the long review!

test/fileio.jl
t-bltg: In this case, `track_order` from `file_create` is inferred from file so this test is correct.

src/types.jl
t-bltg: Yes, left for another PR ;)

t-bltg:
Thanks all for helping finalizing this PR.

Issue #954, opened 2022-06-10
write on attribute doesn't check bounds


simonbyrne:
There is no checking of bounds if you attempt to `write` to an attribute with an array of the wrong size:

```
julia> using HDF5

julia> filename = tempname();

julia> f = h5open(filename, "w");

julia> a,_ = create_attribute(f, "a", zeros(4))
(HDF5.Attribute: a, HDF5.Datatype: H5T_IEEE_F64LE)

julia> read(a)
4-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0

julia> write(a, ones(2))

julia> read(a)
4-element Vector{Float64}:
 1.0
 1.0
 2.271916027e-314
 2.3075649345e-314

```

simonbyrne:
Same for datasets:
```
julia> d,_ = create_dataset(f,"dd",zeros(4))
(HDF5.Dataset: /dd (file: /var/folders/51/y84mbszs2vbd4mvl6ph4_57m0000gn/T/jl_DMPnFZ xfer_mode: 0), HDF5.Datatype: H5T_IEEE_F64LE)

julia> read(d)
4-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0

julia> write(d,ones(2))

julia> read(d)
4-element Vector{Float64}:
 1.0
 1.0
 0.0
 2.3075649345e-314

```

PR #983, opened 2022-07-05
Check buf size on attr or dset write

t-bltg:
Fix https://github.com/JuliaIO/HDF5.jl/issues/954 (only for `Attribute` and `Dataset` for now as the OP examples, I guess there could be more checks for other Types ?).

~~By the way, what do you say we indent all the `@testset` block in the tests ? They are barely readable when un-indented.~~

~~Or a simple solution would be to add a [formatter](https://github.com/domluna/JuliaFormatter.jl) check action to the github workflows (e.g. https://github.com/JuliaPlots/UnicodePlots.jl/blob/master/.github/workflows/format-check.yml), since the code style is inconsistent in `HDF5.jl`, and it's very irritating.~~

### DIFF
--- a/src/datasets.jl
+++ b/src/datasets.jl
@@ -335,14 +335,21 @@ write_dataset(dset::Dataset, memtype::Datatype, x, xfer::DatasetTransferProperti
     API.h5d_write(dset, memtype, API.H5S_ALL, API.H5S_ALL, xfer, x)
 
 # type-specific behaviors
-
+function _check_invalid(dataset::Dataset, buf::AbstractArray)
+    num_bytes_dset = sizeof(datatype(dataset)) * length(dataset)
+    num_bytes_buf = sizeof(eltype(buf)) * length(buf)
+    num_bytes_buf == num_bytes_dset || throw(ArgumentError(
+        "Invalid number of bytes: $(num_bytes_buf) != (num_bytes($(name(dataset))) = $num_bytes_dset)"
+    ))
+    stride(buf, 1) == 1 || throw(ArgumentError("Cannot read/write arrays with a different stride than `Array`"))
+end
 function read_dataset(dataset::Dataset, memtype::Datatype, buf::AbstractArray, xfer::DatasetTransferProperties=dataset.xfer)
-    stride(buf, 1) != 1 && throw(ArgumentError("Cannot read arrays with a different stride than `Array`"))
+    _check_invalid(dataset, buf)
     API.h5d_read(dataset, memtype, API.H5S_ALL, API.H5S_ALL, xfer, buf)
 end
 
 function write_dataset(dataset::Dataset, memtype::Datatype, buf::AbstractArray, xfer::DatasetTransferProperties=dataset.xfer)
-    stride(buf, 1) != 1 && throw(ArgumentError("Cannot write arrays with a different stride than `Array`"))
+    _check_invalid(dataset, buf)
     API.h5d_write(dataset, memtype, API.H5S_ALL, API.H5S_ALL, xfer, buf)
 end
 function write_dataset(dataset::Dataset, memtype::Datatype, str::Union{AbstractString,Nothing}, xfer::DatasetTransferProperties=dataset.xfer)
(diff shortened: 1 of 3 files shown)


t-bltg:
> I would appreciate if we did not mix stylistic changes and functional changes in the same pull request.

I knew I was going to get beaten for that, fixed.

src/datasets.jl
```diff
@@ -335,14 +335,21 @@ write_dataset(dset::Dataset, memtype::Datatype, x, xfer::DatasetTransferProperti
     API.h5d_write(dset, memtype, API.H5S_ALL, API.H5S_ALL, xfer, x)
 
 # type-specific behaviors
-
+function _check_invalid(dataset::Dataset, buf::AbstractArray)
+    num_bytes_dset = sizeof(datatype(dataset)) * length(dataset)
+    num_bytes_buf = sizeof(eltype(buf)) * length(buf)
```
mkitti, line 340: We should do an integer multiplication overflow check here.

See `MEM07-C` for `calloc`:
https://wiki.sei.cmu.edu/confluence/display/c/MEM07-C.+Ensure+that+the+arguments+to+calloc%28%29%2C+when+multiplied%2C+do+not+wrap

or 

`Base.Checked.checked_mul`

```julia
help?> Base.checked_mul
  Base.checked_mul(x, y)

  Calculates x*y, checking for overflow errors where applicable.

  The overflow protection may impose a perceptible performance penalty.
```

mkitti: Suppose someone crafts a special HDF5 file with
* `sizeof(datatype(dataset)) == typemax(Int)÷2`
* `length(dataset) == typemax(Int)-5`

```julia
julia> (typemax(Int)÷2)*(typemax(Int)-5)
6
```

Via integer overflow, one could craft a HDF5 to match the size of the buffer in some program, and then exploit that to do a buffer overflow attack. If we're going to check the buffer size, then we might as well do this to a high standard.

mkitti, line 340: https://docs.julialang.org/en/v1/base/math/#Base.Checked.checked_mul

```julia
julia> Base.checked_mul(typemax(Int)÷2, typemax(Int)-5)
ERROR: OverflowError: 4611686018427387903 * 9223372036854775802 overflowed for type Int64
```

```suggestion
    num_bytes_dset = Base.checked_mul(sizeof(datatype(dataset)), length(dataset))
    num_bytes_buf = Base.checked_mul(sizeof(eltype(buf)), length(buf))
```

t-bltg: I don't see where these are related to allocations: `num_bytes_dset` and `num_bytes_buf` are only used to validate arguments.

mkitti (approved):

src/datasets.jl
```diff
@@ -335,14 +335,21 @@ write_dataset(dset::Dataset, memtype::Datatype, x, xfer::DatasetTransferProperti
     API.h5d_write(dset, memtype, API.H5S_ALL, API.H5S_ALL, xfer, x)
 
 # type-specific behaviors
-
+function _check_invalid(dataset::Dataset, buf::AbstractArray)
+    num_bytes_dset = Base.checked_mul(sizeof(datatype(dataset)), length(dataset))
+    num_bytes_buf = Base.checked_mul(sizeof(eltype(buf)), length(buf))
+    num_bytes_buf == num_bytes_dset || throw(ArgumentError(
+        "Invalid number of bytes: $(num_bytes_buf) != (num_bytes($(name(dataset))) = $num_bytes_dset)"
```
musm, line 342: The only thing that's confusing in this error message is `(num_bytes($(name(dataset)))`, since `num_bytes` isn't the name of a valid function. I think you probably meant `sizeof` ?

musm:
LGTM sans review comment

src/datasets.jl
t-bltg: So I tried to use `sizeof` in the beginning, but it was failing on the HDF5 side in the tests (I didn't dig further though).
And for consistency, I also used `sizeof(eltype(...)) * length(...)` on the julia side even though it means `sizeof(buf)`.

Thus, it will be odd to name it `sizeof` even if we don't use `sizeof` under the hood, but I understand that `num_bytes` can be named alternatively even though it is IMO self explanatory.

Maybe @mkitti has an opinion on this ?

musm: If we opt for what was suggested the attribute error method should be tweaked to match.

t-bltg: I like this. Furthermore, we could argue that `num_bytes` is redundant with the previous sentence element `number of bytes:`.

musm: I guess we can leave the attribute error as is, since in that case the definitions agree ?

PR #986, opened 2022-07-07
Add bitshuffle

### LINKED ISSUE #981: BITSHUFFLE filter not supported out of the box for HDF5 reading
Commit #899 added support for using the BITSHUFFLE filter when writing, but does not register the filter for reading and so HDF5 files written with the bitshuffle filter are not in general readable without an external filter.

So in HDF5 the Blosc and bitshuffle filters have distinct IDs (32001 and 32008 respectively) but line 148 in H5ZBlosc.jl

```
filterid(::Type{BloscFilter}) = H5Z_FILTER_BLOSC
```
attaches the filter to id 32001 only.

It looks to me like adding an abstract type or two might be a quick fix for this.

jamesrhester:
This adds the bitshuffle filter for HDF5. Closes #981.

### DIFF
--- a/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
+++ b/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
@@ -257,11 +257,13 @@ The Bitshuffle filter can optionally include compression "lz4" or "zstd". For "z
 comp_level can be provided. This is ignored for "lz4" compression. If `blocksize`
 is zero the default bitshuffle blocksize is used.
 """
-BitshuffleFilter(;blocksize = 0, compressor="",comp_level=0) = begin
+function BitshuffleFilter(;blocksize = 0, compressor="",comp_level=0)
     lowercase(compressor) in ("lz4","zstd","") || throw(ArgumentError("Invalid bitshuffle compression $compression"))
     compcode = 0
-    if compressor == "lz4" compcode = BSHUF_H5_COMPRESS_LZ4
-    elseif compressor == "zstd" compcode = BSHUF_H5_COMPRESS_ZSTD
+    if compressor == "lz4"
+        compcode = BSHUF_H5_COMPRESS_LZ4
+    elseif compressor == "zstd"
+        compcode = BSHUF_H5_COMPRESS_ZSTD
     end
     BitshuffleFilter(BSHUF_VERSION_MAJOR,BSHUF_VERSION_MINOR,0,blocksize,compcode,comp_level)
 end
--- a/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
+++ b/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
@@ -1,7 +1,8 @@
 #==
 
-Code implementing the bitshuffle filter for HDF5. A rough translation
-of bshuf_h5filter.c from https://github.com/kiyo-masui/bitshuffle.
+Julia code wrapping the bitshuffle filter for HDF5. A rough translation of
+bshuf_h5filter.c by Kiyoshi Masui, see
+https://github.com/kiyo-masui/bitshuffle.
 
 ==#
 module H5Zbitshuffle
--- a/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
+++ b/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
@@ -1,10 +1,12 @@
 #==
-
 Julia code wrapping the bitshuffle filter for HDF5. A rough translation of
 bshuf_h5filter.c by Kiyoshi Masui, see
 https://github.com/kiyo-masui/bitshuffle.
-
 ==#
+"""
+The bitshuffle filter for HDF5. See https://portal.hdfgroup.org/display/support/Filters#Filters-32008
+and https://github.com/kiyo-masui/bitshuffle for details.
+"""
 module H5Zbitshuffle
 
 using bitshuffle_jll
--- a/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
+++ b/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
@@ -10,16 +10,15 @@ module H5Zbitshuffle
 using bitshuffle_jll
 
 using HDF5.API
-import HDF5.Filters: Filter, filterid, register_filter, filtername, filter_func, filter_cfunc
+import HDF5.Filters: Filter, filterid, register_filter, filtername, filter_func, filter_cfunc, set_local_func, set_local_cfunc
 
-export BSHUF_H5_COMPRESS_LZ4, BSHUF_H5_COMPRESS_ZSTD
+export BSHUF_H5_COMPRESS_LZ4, BSHUF_H5_COMPRESS_ZSTD, BitshuffleFilter, H5Z_filter_bitshuffle
 
 # From bshuf_h5filter.h
 
 const BSHUF_H5_COMPRESS_LZ4 = 2
 const BSHUF_H5_COMPRESS_ZSTD = 3
-const BSHUF_H5_FILTER = 32008
-const H5Z_FILTER_BITSHUFFLE = API.H5Z_filter_t(BSHUF_H5_FILTER)
+const H5Z_FILTER_BITSHUFFLE = API.H5Z_filter_t(32008)
 
 const BSHUF_VERSION_MAJOR = 0
 const BSHUF_VERSION_MINOR = 4
@@ -32,28 +31,38 @@ const bitshuffle_name = "HDF5 bitshuffle filter; see https://github.com/kiyo-mas
 function bitshuffle_set_local(dcpl::API.hid_t, htype::API.hid_t, space::API.hid_t)
 
     # Sanity check of provided values and set element size
-    
+
     bs_flags = Ref{Cuint}()
     bs_values = Vector{Cuint}(undef,8)
     bs_nelements = Ref{Csize_t}(length(bs_values))
     
-    r = API.h5p_get_filter_by_id(dcpl, H5Z_FILTER_BITSHUFFLE, bs_flags, bs_nelements,
-                                 bs_values, 0, NULL, NULL)
-    if r<0 return -1 end
+    API.h5p_get_filter_by_id(dcpl, H5Z_FILTER_BITSHUFFLE, bs_flags, bs_nelements,
+                                 bs_values, 0, C_NULL, C_NULL)
+
+    @debug "Initial filter info" bs_flags bs_values bs_nelements
     
     flags = bs_flags[]
-    nelements = max(bs_nelements[],3)
 
     # set values
     
     bs_values[1] = BSHUF_VERSION_MAJOR
     bs_values[2] = BSHUF_VERSION_MINOR
-    bs_values[3] = API.h5t_get_size(htype)
+    
+    elem_size = API.h5t_get_size(htype)
 
+    @debug "Element size for $htype reported as $elem_size"
+
+    if elem_size <= 0
+        return API.herr_t(-1)
+    end
+
+    bs_values[3] = elem_size
+    nelements = bs_nelements[]
+    
     # check user-supplied values
     
     if nelements > 3
-        if bs_values[4] % 8 || bs_values[4] < 0 return API.herr_t(-1) end
+        if bs_values[4] % 8 !=0 || bs_values[4] < 0 return API.herr_t(-1) end
     end
 
     if nelements > 4
@@ -62,8 +71,9 @@ function bitshuffle_set_local(dcpl::API.hid_t, htype::API.hid_t, space::API.hid_
         end
     end
 
-    r = API.h5p_modify_filter(dcpl, H5Z_FILTER_BITSHUFFLE, bs_flags, bs_nelements, bs_values)
-    if r < 0 return API.herr_t(-1) end
+    @debug "Final values" bs_values
+
+    API.h5p_modify_filter(dcpl, H5Z_FILTER_BITSHUFFLE, bs_flags[], nelements, bs_values)
     
     return API.herr_t(1)
 end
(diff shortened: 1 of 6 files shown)


mkitti:
Great. Could you add some tests?
https://github.com/JuliaIO/HDF5.jl/blob/master/test/filter.jl

jamesrhester:
Will do.

mkitti:
I think you are missing an implementation of `bshuf_h5_set_local` function.
https://github.com/kiyo-masui/bitshuffle/blob/master/src/bshuf_h5filter.c#L29

jamesrhester:
Yes, working on it now. Having to write tests meant I discovered that I'd ignored the writing part of the filter...

mkitti:
See https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/filters/H5Zblosc/src/H5Zblosc.jl#L19 if you need an example of that.

Also, we could just compile the C code for the plugin, probably as a separate Yggrassil recipe. We could then just use the `H5PL` interface to tell HDF5 where the plugin is.

jamesrhester:
It turns out that I put the lower compat limit of `bitshuffle_jll` as 1.6, have to wait for that to be fixed.

jamesrhester:
OK, I'm informed (see https://github.com/JuliaPackaging/Yggdrasil/pull/5111) that the lower compatibility limit for `bitshuffle_jll` should be 1.6. So I don't know if I should be trying to pass the Julia 1.3 checks?

filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
```diff
@@ -0,0 +1,280 @@
+#==
```
mkitti, line 1: Make this a docstring so that `?H5Zbitshuffle` works.

jamesrhester:
So adding the following line to the `filters.jl` test causes problems:

```
@static if VERSION >= v"1.6" using H5Zbitshuffle end
```

Running with julia 1.6 the relevant part of the error is:
```
┌ Error: Non-fatal H5 bitshuffle plugin error: 
│   e = UndefVarError: H5Z_FLAG_REVERSE not defined
└ @ H5Zbitshuffle ~/programs/HDF5.jl/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl:228
22-element Vector{Base.StackTraces.StackFrame}:
 H5Z_filter_bitshuffle(flags::UInt32, cd_nelmts::UInt64, cd_values::Ptr{UInt32}, nbytes::UInt64, buf_size::Ptr{UInt64}, buf::Ptr{Ptr{Nothing}}) at H5Zbitshuffle.jl:199
 h5d_write(dataset_id::HDF5.Dataset, mem_type_id::HDF5.Datatype, mem_space_id::Int64, file_space_id::Int64, xfer_plist_id::HDF5.DatasetTransferProperties, buf::Matrix{Float64}) at functions.jl:617
```
I have committed anyway to see how CI copes. I have no idea how to fix it after banging my head against a wall for an hour.

filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
jamesrhester: Did this in the latest commit.

filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
```diff
@@ -0,0 +1,282 @@
...
+is zero the default bitshuffle blocksize is used.
+"""
+BitshuffleFilter(;blocksize = 0, compressor="",comp_level=0) = begin
+    lowercase(compressor) in ("lz4","zstd","") || throw(ArgumentError("Invalid bitshuffle compression $compression"))
+    compcode = 0
+    if compressor == "lz4" compcode = BSHUF_H5_COMPRESS_LZ4
+    elseif compressor == "zstd" compcode = BSHUF_H5_COMPRESS_ZSTD
+    end
+    BitshuffleFilter(BSHUF_VERSION_MAJOR,BSHUF_VERSION_MINOR,0,blocksize,compcode,comp_level)
+end
```
mkitti, line 267: The formatting here is strange. Could we just use the normal `function` syntax here?

```suggestion
function BitshuffleFilter(;blocksize = 0, compressor="",comp_level=0)
    lowercase(compressor) in ("lz4","zstd","") || throw(ArgumentError("Invalid bitshuffle compression $compression"))
    compcode = 0
    if compressor == "lz4"
        compcode = BSHUF_H5_COMPRESS_LZ4
    elseif compressor == "zstd"
        compcode = BSHUF_H5_COMPRESS_ZSTD
    end
    BitshuffleFilter(BSHUF_VERSION_MAJOR,BSHUF_VERSION_MINOR,0,blocksize,compcode,comp_level)
end
```

mkitti:
```suggestion
function BitshuffleFilter(;blocksize = 0, compressor=:none, comp_level=0)
    if isa(compressor, String)
       compressor = Symbol(lowercase(compressor))
    end
    compressor in (:lz4, :zstd, :none) || throw(ArgumentError("Invalid bitshuffle compression $compression"))
    compcode = 0
    if compressor == :lz4
        compcode = BSHUF_H5_COMPRESS_LZ4
    elseif compressor == :zstd
        compcode = BSHUF_H5_COMPRESS_ZSTD
    end
    BitshuffleFilter(BSHUF_VERSION_MAJOR,BSHUF_VERSION_MINOR,0,blocksize,compcode,comp_level)
end```

File: filters/H5Zbitshuffle/src/H5Zbitshuffle.jl [hunk 3]
```diff
@@ -0,0 +1,284 @@
...
+
+"""
+    BitshuffleFilter(blocksize=0,compression="",comp_level=0)
+
+The Bitshuffle filter can optionally include compression "lz4" or "zstd". For "zstd"
+comp_level can be provided. This is ignored for "lz4" compression. If `blocksize`
+is zero the default bitshuffle blocksize is used.
+"""
+function BitshuffleFilter(;blocksize = 0, compressor="",comp_level=0)
```
mkitti, line 260: Thanks for the formatting. I wonder if using symbols `:lz4, :zstd, or :none` might be cleaner. It also might be good to explicitly check for `:none` so that typos do not fail silently. If someone typed "lz4d" or `:cstd` as a compressor, a useful error should be emitted.

test/filter.jl
```diff
@@ -65,8 +67,35 @@ ds = create_dataset(
     f, "blosc_bitshuffle", datatype(data), dataspace(data),
     chunk=(100,100), filters=BloscFilter(shuffle=H5Zblosc.BITSHUFFLE)
 )
+    
 write(ds, data)
 
+extra_bitshuffle() = begin
```
mkitti, line 73: Just use `function` please.

test/runtests.jl
```diff
@@ -6,6 +6,7 @@ Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zblosc")))
 Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zbzip2")))
 Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zlz4")))
 Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zzstd")))
+Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zbitshuffle")))
```
mkitti, line 9:
```suggestion
@static if VERSION >= v"1.6"
    Pkg.develop(PackageSpec(path=joinpath(filter_path, "H5Zbitshuffle")))
end
```

filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
jamesrhester: OK makes sense, will update to use symbols.

test/filter.jl
jamesrhester: Fixed in latest commit

filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
jamesrhester: Changed to symbols in latest commit. `:none` is included in the list of allowed symbols that are checked and the error message should be sufficiently informative I think.

mkitti:
I'm happy to merge this if you are. Thanks again.

jamesrhester:
Yes, I'm happy to merge. Thanks for your review and help.

mkitti (approved):

mkitti:
I'm pausing for a few days to give others the chance to review. I'm also going to do some further testing with plugins from other sources and via different APIs.

mkitti:
Can the license correctly be categorized as a MIT license?

jamesrhester:
Yes, it was meant to be a MIT license copied from the original bitshuffle distribution. Thanks for merging this.

mkitti:
@jamesrhester , I'm seeing a sporadic CI failure on 64-bit Windows:
https://github.com/JuliaIO/HDF5.jl/runs/7292646756?check_suite_focus=true

We may need to review

jamesrhester:
Looks like nightly is OK, but Julia 1.7 has a problem. I'll see if I can spin up a Windows 64 VM here and run the test repeatedly.

jamesrhester:
I'm suspecting it could be because of a lack of `@GC.preserve` at the end of the filter function with the `unsafe_load` and `unsafe_store!` calls. The corresponding H5Zblosc filter doesn't have these, however, so once I've got a test VM running I'll try and duplicate the error and then see if the `@GC.preserve` fixes it.

mkitti:
The original code allocates 11 `Cuints` for values, but you only allocate 8 (similar to `tmp_values`) in `bitshuffle_set_local`. You also don't seem to move the values to a higher position. I think you may be overwriting values stored in `bs_values` by `API.h5p_get_filter_by_id`. Perhaps I am missing something?

https://github.com/kiyo-masui/bitshuffle/blob/fdfcd404ac8dcb828857a90c559d36d8ac4c2968/src/bshuf_h5filter.c#L47-L50

mkitti:
I see now that you added three extra values to the struct similar to how we implemented the Blosc filter.

mkitti:
Hmm, you never initialize `bs_values` to `0`.

jamesrhester:
The way this works, as I understand it, is that the `set_local` routine provides an opportunity to fill and check values in the `BitshuffleFilter` struct, which is what `bs_values` is filled with after `H5API.h5p_get_filter_by_id`. It would also be OK to completely skip this step and set everything at BitshuffleFilter initialisation, except that the element size isn't known at the time that the filter is constructed. So in the Julia version I have set almost every value at initialisation, except for element size, but repeat the setting of MAJOR, MINOR values.
Still sorting out access to a Windows VM.

mkitti (commented):
The other things I noticed were:

1. The values are zero initialized.
2. We probably should integer division. `/` in Julia will produce a Float64.

filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
```diff
@@ -0,0 +1,284 @@
...
+
+            # Use compression
+            
+            if(flags & API.H5Z_FLAG_REVERSE) != 0 # unshuffle and decompress
+
+                # First 8 bytes is number of uncompressed bytes
+                nbytes_uncomp = ccall((:bshuf_read_uint64_BE,libbitshuffle),Cuint,(Ptr{Cvoid},),in_buf)
+                # Next 4 bytes are the block size
+                
+                block_size = ccall((:bshuf_read_uint32_BE,libbitshuffle),Cuint,(Ptr{Cvoid},),in_buf+8)/elem_size
```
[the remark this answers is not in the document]
jamesrhester: Makes sense.

mkitti:
One issue I encountered that relates to `GC.@preserve` and decompression is detailed here:
https://github.com/JuliaLang/julia/issues/43402

This was due a specific change on the development branch of Julia though.

What happens if `bs_values` is getting freed after `bitshuffle_set_local` but before `H5Z_filter_bitshuffle` finishes?

I have a Windows box, and so far I am not able to reproduce the error reliably.

mkitti:
I'm going to try to focus testing here:
https://github.com/JuliaIO/HDF5.jl/pull/988

jamesrhester:
> One issue I encountered that relates to `GC.@preserve` and decompression is detailed here: [JuliaLang/julia#43402](https://github.com/JuliaLang/julia/issues/43402)
> 
> This was due a specific change on the development branch of Julia though.

I read through that but couldn't see how it applies here. Weird bug though.

> What happens if `bs_values` is getting freed after `bitshuffle_set_local` but before `H5Z_filter_bitshuffle` finishes?

Well, as `bs_values` is of type `Ptr{Cuint}` it should not get freed by Julia if I understand the whole `Ptr`, `Ref` thing.

mkitti:
> Well, as `bs_values` is of type `Ptr{Cuint}` it should not get freed by Julia if I understand the whole `Ptr`, `Ref` thing.

The problem is that you allocate it via `bs_values = Vector{Cuint}(undef,8)`. When it gets passed to `ccall`, the pointer behind the `Vector` is passed to HDF5. After `bitshuffle_set_local` completes, Julia thinks it can free the pointer behind `bs_values`.

The question is if `H5Pmodify_filter` just hangs on to the pointer or if it makes a copy.

mkitti:
`H5Pmodify_filter` does seem to copy the data:
https://github.com/HDFGroup/hdf5/blob/c064d3481b582653c1e0d0043a17527fd73e8c4d/src/H5Z.c#L1095-L1096

mkitti:
Aha, I found something. 

The C function signature is:

```C
int64_t bshuf_decompress_lz4(const void* in, void* out, const size_t size,
        const size_t elem_size, size_t block_size) {
```

The `ccall` however is

```julia
                    err = ccall((:bshuf_decompress_lz4,libbitshuffle),Cint,
                                (Ptr{Cvoid},Ptr{Cvoid},Cuint,Cuint,Cuint),
                                in_buf,out_buf,size,elem_size,block_size)
```

This should be 

```julia
                    err = ccall((:bshuf_decompress_lz4,libbitshuffle),Cint,
                                (Ptr{Cvoid},Ptr{Cvoid},Csize_t,Csize_t,Csize_t),
                                in_buf,out_buf,size,elem_size,block_size)
```



```julia
julia> Csize_t
UInt64

julia> Cuint
UInt32
```

PR #988, opened 2022-07-13
Focus testing on H5Zbitshuffle

mkitti:


mkitti:
@jamesrhester these changes seemed to have improved the test passing consistency.
1. Use integer division, `÷` instead of right division `/` operator which gives floating point results for integer arguments.
2. Match C function signature and types.

jamesrhester:
Thanks for working on that, sorry I couldn't contribute more.

mkitti:
Thanks for contributing the code once again!

PR #989, opened 2022-07-13
Add missing H5P functions

mkitti:
In particular, I am interested in `h5p_copy`. Also fix alphabetization.


musm:
Seems like you forgot to actually run the generator file to create the definitions.

mkitti:
Ah yes, I forgot to commit it.

PR #992, opened 2022-07-19
Ignore src/api/functions.jl for code coverage

mkitti:
Ignore `src/api/functions.jl` since those are just bindings for the C API.

See https://docs.codecov.com/docs/ignoring-paths


musm:
Makes sense

PR #996, opened 2022-07-28
test HDF5_jll 1.12.2

visr:
Locally it works fine for me, but I wanted to trigger CI here as well for other platforms.

https://github.com/JuliaPackaging/Yggdrasil/pull/5248
https://github.com/JuliaBinaryWrappers/HDF5_jll.jl/releases/tag/HDF5-v1.12.2%2B0

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,6 @@ Convenience functions for attributes attached to datasets are also provided:
 ```julia
 A = Vector{Int}(1:10)
 h5write("bar.h5", "foo", A)
-h5writeattr("bar.h5", "foo", Dict("c"=>"value for metadata parameter c","d"=>"metadata d"))
+h5writeattr("bar.h5", "foo", Dict("c"=>"value for metadata parameter c", "d"=>"metadata d"))
 h5readattr("bar.h5", "foo")
 ```


visr:
The macOS failure should be fixed when https://github.com/JuliaPackaging/Yggdrasil/pull/5249 is out.

Windows 32 bit has a single test failure:
```
AbstractString: Test Failed at D:\a\HDF5.jl\HDF5.jl\test\plain.jl:1237
  Expression: ext_prop.size == 10 * 20 * sizeof(Int)
   Evaluated: 0x0000000000000000 == 800
```

I thought it might be a Int32/Int thing, but that would not explain why the left hand side is 0.

visr:
Ok so HDF5_jll 1.12.2 is now yanked from the registry to avoid breaking this package for some users. Summary in https://github.com/JuliaPackaging/Yggdrasil/issues/4511#issuecomment-1198134988. Hope that we can somehow get that release distributed so we can finally get unstuck with NetCDF builds.

Feel free to merge this if you like space, or close.

musm:
I'll leave this open until the binaries get updated with the correct curl distribution? Sounds like that's the remaining problem.

visr:
Indeed that's the problem. I'm just not sure how to get those binaries at this point.

visr:
@musm and others, what do you think about the proposal in https://github.com/JuliaPackaging/Yggdrasil/issues/4511#issuecomment-1198615732?

visr:
Could you elaborate? I think the problem is that for libnetcdf there is no libhdf5 version that works on all platforms, including newer platforms. Registry edits cannot be platform specific.

mkitti:
Is there a platform that does not work with HDF5 v1.12.0? I thought the problem was particular to HDF5 v1.12.1.

visr:
I believe so, macOS M1 and other platforms on 1.8, see https://github.com/Alexander-Barth/NCDatasets.jl/issues/182#issuecomment-1178659255.

mkitti:
Hmm, let me read more about this.

visr:
When I compare 1.12.1 and 1.12.0 in https://github.com/JuliaBinaryWrappers/HDF5_jll.jl/releases there is one extra asset, `HDF5.v1.12.1.aarch64-apple-darwin.tar.gz`.

visr:
I triggered CI with the new build, and as can be expected macOS passes again since the build is the same as before.

And, also as expected, Windows 32 bits still fails just like I mentioned above in https://github.com/JuliaIO/HDF5.jl/pull/996#issuecomment-1197947881.

Does anyone have any ideas there? Or should we just mark the test as broken?

mkitti:
I don't see the Win32 x86 test failure as a huge blocker especially if the fix involves just downgrading HDF5_jll

visr:
Yeah it's only one failure, which seems to suggest that overall it works, though I don't understand the importance of this particular test, what the consequences are of it being broken.

mkitti:
I'll investigate this over the weekend.

PR #998, opened 2022-07-30
Debug 32-bit windows

mkitti:
Our API definition for `h5p_set_external` was wrong. `size` was `Csize_t` but should have been `hsize_t`. On 32-bit Windows, `Csize_t` is 32-bits while `hsize_t` is 64-bits.

Changing that did not fix the issue, however.

With the latest binaries it seems that the size is now shifted by 32 bits. Previously the upper 32 bits had garbage in it.

I suspect the issue is related to https://github.com/HDFGroup/hdf5/pull/1821

```julia
    @static if Sys.iswindows() && sizeof(Int) == 4
        lower = 0xffffffff & sz[]
        upper = 0xffffffff & (sz[] >> 32)
        sz[] = lower == 0 ? upper : lower
    end
```

Fix #996

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,6 +9,33 @@ on:
   pull_request:
 
 jobs:
+  HDF5-system-libs:
+    name: julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} (system libhdf5 + mpich)
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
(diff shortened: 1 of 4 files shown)


visr:
Wow that turned out to be quite a deep hole, nice work finding the issue!

musm:
Good hunting. Yikes, on the hdf5 lib side... Would've expected their testing to pick this up.

musm:
Ops there's still a debug statement in there..

PR #1000, opened 2022-07-31
Fix failing tests by revising H5O_get_info* API on library version

mkitti:
Cherry picked from f5910a6cf8c99b03737c4cdd47d9d39e7bec98ab


musm:
Thanks for handling this. Would it be better to merge this in first? If that's your preference I'd only suggest to change the commit name so it's more descriptive since this is enabling additional version dependent API.

musm:
I take that back. Honestly, the title is pretty clear, just a minor edit.

PR #1002, opened 2022-08-06
Move code to highlevel.jl

mkitti:
* Moves high-level interface and property manipulation to highlevel.jl


src/propertymanipulation.jl
```diff
@@ -0,0 +1,7 @@
+### Property manipulation ###
```
musm, line 1: Can't these live in the highlevel file? I don't think it's helpful to have these in a separate file.

mkitti: Sure, why not?

src/highlevel.jl
```diff
@@ -0,0 +1,110 @@
...
+        pv = pairs(Base.structdiff((; pv...), chunked_props))
+    else
+        if need_chunks && !have_chunks
+            pv = pairs((; chunk=chunk, pv...))
+        end
+    end
+    write(parent, path, val; pv...)
+end
+
+### Property manipulation ###
```
musm, line 104: On second thought why not at the end of https://github.com/JuliaIO/HDF5.jl/blob/master/src/properties.jl ? Seems more appropriate especially since that's where the `*Properties` types are defined?

mkitti: I tried that. It does not work. We would need to move the property types up into types.

musm: ahh

PR #984, opened 2022-07-05
Add formatting check action and apply an initial format

t-bltg:
This PR is a proposition in order to add consistency across the source and test files (after https://github.com/JuliaIO/HDF5.jl/pull/983#issuecomment-1175011022).

The tests are hard to read because the `@testset` blocks are not indented: https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/test/plain.jl#L438
requiring (spurious) manual addition of comment such as `end # testset "Test h5d_fill`.

The source files mixes both `function ... end` and one liner function declarations like :
https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/src/datasets.jl#L320
, and https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/src/datasets.jl#L332
making this hard to follow (inconsistent).

The declarations like:
https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/src/groups.jl#L115
are hard to read and it would be IMO better if they could be split into multiple lines such as https://github.com/JuliaIO/HDF5.jl/blob/ccdb3e90a5910bee954d066de7b3f775f9067bc3/src/readwrite.jl#L247
(basically break on a specific margin / line width).

The indentation is inconsistent e.g. https://github.com/JuliaIO/HDF5.jl/blob/d6ecf42d035784f78616a24b03cda7c037eda7c9/src/api/helpers.jl#L299 which uses 3 spaces instead of 4 (https://docs.julialang.org/en/v1/manual/style-guide/).

I'll stop there with the examples, my point being that I don't care which style convention is used (this is very opinionated), **as long as it is consistent across all source files**.

Currently in this repo, no style guidelines are provided for developers, inducing wasted time on style discussions (e.g. https://github.com/JuliaIO/HDF5.jl/pull/983#issuecomment-1175011022 or https://github.com/JuliaIO/HDF5.jl/pull/917#discussion_r830190912).

If needed, formatting auto-generated files can be omitted  by adding `#! format: off` to a file header (e.g. `gen/gen_wrappers.jl` -> `src/api/functions.jl`) ?

I propose that we only apply formatting to `src` and `test` directories (for the end users), but this could be extended to other directories like `gen` (to be discussed ?).

[...]

PR #1003, opened 2022-08-06
cleanup test files

t-bltg:
After https://github.com/JuliaIO/HDF5.jl/pull/984.
Per https://github.com/JuliaIO/HDF5.jl/pull/984#issuecomment-1204353657.

### DIFF
--- a/test/chunkstorage.jl
+++ b/test/chunkstorage.jl
@@ -169,4 +169,4 @@ using Test
     end == reshape(1:24, 4, 6)
 
     rm(fn)
-end # testset "Raw Chunk I/O"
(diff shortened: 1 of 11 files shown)


mkitti:
I prefer these comments to stay.

musm:
I appreciate the effort. Although I don't have a problem with this,  this just seems like churn without  benefit.

mkitti:
Thank you again for this. Having the utility of having the comments there seems to still outweigh the aesthetics at the moment.

For these to go away, an more effective approach would be to refactor the test sets into smaller sets.

Issue #990, opened 2022-07-14
Finalizers cause exception_access_violation while using HDF5 from a thread.


mkitti:
When running long jobs, I'm occasionally running into this traces like this. What I think is happening is that the finalizer is running to close out a property while another thread is trying to access the HDF5 library. This can be be mitigated by running `GC.gc()` when between jobs where threads are used.

Note I'm doing this while using HDF5.jl on a single spawned thread.

```julia
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x7ffb692e3cac -- H5FL_reg_calloc at C:\Users\simview\.julia\artifacts\a285947f59f2fa90c9ac36239548c5af147d0da5\bin\libhdf5-0.dll (unknown line)
in expression starting at REPL[1]:1
H5FL_reg_calloc at C:\Users\simview\.julia\artifacts\a285947f59f2fa90c9ac36239548c5af147d0da5\bin\libhdf5-0.dll (unknown line)
H5CX_push at C:\Users\simview\.julia\artifacts\a285947f59f2fa90c9ac36239548c5af147d0da5\bin\libhdf5-0.dll (unknown line)
H5Iis_valid at C:\Users\simview\.julia\artifacts\a285947f59f2fa90c9ac36239548c5af147d0da5\bin\libhdf5-0.dll (unknown line)
h5i_is_valid at C:\Users\simview\.julia\packages\HDF5\7zvRl\src\api\functions.jl:1354
isvalid at C:\Users\simview\.julia\packages\HDF5\7zvRl\src\properties.jl:19 [inlined]
close at C:\Users\simview\.julia\packages\HDF5\7zvRl\src\properties.jl:11
jl_apply at /cygdrive/c/buildbot/worker/package_win64/build/src\julia.h:1788 [inlined]
run_finalizer at /cygdrive/c/buildbot/worker/package_win64/build/src\gc.c:278
jl_gc_run_finalizers_in_list at /cygdrive/c/buildbot/worker/package_win64/build/src\gc.c:365
run_finalizers at /cygdrive/c/buildbot/worker/package_win64/build/src\gc.c:394 [inlined]
run_finalizers at /cygdrive/c/buildbot/worker/package_win64/build/src\gc.c:372
jl_gc_collect at /cygdrive/c/buildbot/worker/package_win64/build/src\gc.c:3267
[...]

PR #1006, opened 2022-08-21
Cleanup group code

### LINKED ISSUE #829: `create_group` probably should throw a custom exception
Right now there's a call to `error`:

https://github.com/JuliaIO/HDF5.jl/blob/ef228d57be54d2082a6a7581895fd4becbfe3714/src/HDF5.jl#L723

In my code, I'd like to overwrite the data in a group _if it already exists_, but currently that would make me catch a `ErrorException`:

```julia
try
    create_group(fid, grp_name)
catch e
   if e isa ErrorException
      @warn "Replacing data!"
   else
      throw(e)
   end
end
```

However, this doesn't tell me anything about where or why the `ErrorException` occurred: maybe the file was corrupt, maybe it got randomly deleted on disk - I can't tell.

It would be nice to have a _type_ of exception that would explain what caused it and let me decide how to handle it. Something like this:

```julia
try
    create_group(fid, grp_name)
catch e
   if e isa GroupExistsException
      @warn "Replacing data!"
   else
      throw(e)
   end
end
```

simonbyrne:
- Clarifies and adds docs
- Simplifies interface to `create_group`, avoiding mutation of default properties
- Closes properties after completion (helps #990)
- Removes custom exception for `create_group` (fixes #829)


src/groups.jl
```diff
@@ -1,43 +1,43 @@
 """
-    create_group(parent, path[, lcpl, gcpl]; properties...)
+    create_group(parent::Union{File,Group}, path::AbstratString; properties...)
```
mkitti, line 2:
```suggestion
    create_group(parent::Union{File,Group}, path::AbstractString; properties...)
```

mkitti (changes requested):
Could we separate the breaking and non-breaking components of this? 

For the breaking portion, we will probably want to use the `HDF5Context` structure to initialize the property lists rather than continuing to use the globals. We also should still support but deprecate the old methods.

simonbyrne:
What parts do you consider breaking?

simonbyrne:
I added a deprecation warning if they're both used.

musm:
> For the breaking portion, we will probably want to use the HDF5Context structure to initialize the property lists rather than continuing to use the globals. We also should still support but deprecate the old methods.

Let's save the HDF5Context changes for a separate PR

musm (approved):

mkitti (approved):
With the `depwarn`, this works for me.

Issue #552, opened 2019-05-21
reading and writing from generic `IO` objects


ExpandingMan:
It is important to be able to read and write from general `IO` objects rather than just files.  This is really important in case you need to stash files over a network, for example with AWS S3 rather than to the file system.

I don't know how cooperative the HDF5 library is going to be with this.  Skimming through the code, it does not look like it will be easy.

ggggggggg:
Can you be more specific, eg do you want to do something like this?

```julia
s3_bucket_with_hdf5_file_in_it = S3.openbucket()
h5 = HDF5.h5open(s3_bucket_with_hdf5_file_in_it, "rw")
h5["a"]=4
x = h5["b"]
```

ExpandingMan:
In that case it would involve writing to a `IOBuffer` object, taking the `Vector{UInt8}` buffer and sending it to AWS via HTTP.  Something like
```julia
io = IOBuffer()
get_data_from_s3!(s3, io)
h5 = h5open(io)
```
I suppose.

baumgold:
Hi @ExpandingMan.  Did you ever manage to find a solution to being able to read HDF5 files from an S3 object store?  The [HDF5/S3 connector](https://www.hdfgroup.org/solutions/cloud-amazon-s3-storage-hdf5-connector/) may be useful here, but I'm not sure if this has since been solved in a different way.  Thanks!

mkitti:
The canonical way would be to use HDF5 virtual file drivers. https://docs.hdfgroup.org/hdf5/v1_12/_v_f_l.html

baumgold:
@mkitti - that’s what I suspected. Any idea if this is available/integrated with HDF5.jl? My understanding is virtual file drivers need to be selected at HDF5 compile-time. I presume we’ll need some changes from HDF5_jll to get this support?

mkitti:
We started adding support for drivers here:
https://github.com/JuliaIO/HDF5.jl/blob/master/src/drivers/drivers.jl

We may be able to use the Core driver to read an I/O stream completely into memory and use that.

ExpandingMan:
I haven't had many occasions to use HDF5, but when I did I was certainly resorting to temp files, which is certainly not ideal.  In linux it's very easy to do this all in-memory (you can use `/dev/shm` or another in-memory directory) there is probably still a lot of overhead to that, so it's in no way an ideal solution.

gszep:
[...]

Issue #818, opened 2021-02-16
Writing fixed width strings is tricky


david-macmahon:
While exploring Parallel HDF5 via HDF5.jl I discovered that parallel HDF5 does not support variable length strings (maybe any variable type?).  To store fixed length strings, I made my own Datatype for HDF5 strings of length N (68 in my case) and then converted my `Vector{String}` to `SVector{68, UInt8}` by doing something like:
```julia
julia> dataset[1:2] = @. rpad(["abc","def"], 68) |> collect |> SVector{68, UInt8};

julia> dataset[1:2]
2-element Array{String,1}:
 "abc                                                                 "
 "def                                                                 "
```
That works, but it is a bit tricky/cumbersome and introduces a dependency on StaticArrays.  Is there an easier way to do this using existing HDF5.jl functionality?  Maybe involving `HDF5.FixedString`???

musm:
Can you simply use?
```julia
c = "rannndomstring"
ntuple(i -> i <= ncodeunits(c) ? codeunit(c, i) : '\0', 68)
```

david-macmahon:
That's better as it removes the dependency on StaticArrays (though have to change `'\0'` to `UInt8(0)`), but I was hoping for an existing function that does this.  I discovered that I could create the Datatype by calling `datatype(" "^68)` so that helps too.  I guess this can be considered a feature request and/or documentation update request.  It's certainly not a show stopper.  Thanks!

PR #1010, opened 2022-09-11
add support for read-only s3 virtual driver

gszep:
Related to #552; We would like the [equivalent functionality as h5py](https://github.com/h5py/h5py/pull/1755)
```julia
h5open("s3://..."; driver=Drivers.ROS3(aws_region="us-west-2", key="", id="")) do file
    file
end
```


mkitti (changes requested):
Thanks for the pull request.

We need to pass a pointer to `H5Pset_fapl_ros3`. Perhaps we should box the struct in a Ref or concretely as a `RefValue`.

gszep:
@mkitti thank you for the helpful review! Do you have a recommendation for a unit test for this? We need a public HDF5 file hosted on S3.. It appears that some of the builds do not pass raising
```could not find function H5Pset_fapl_ros3 in library libhdf5```
suggesting that these build to not have `libhdf5` build with `ros3`

**Comment by mkitti** [comment]
You might need to register this lazily like the mpio driver rather than during `__init__`. Not all of the HDF5 binaries we bundle have this driver enabled.

**Comment by mkitti** [comment]
The modifications to functions.jl directly are not going to work. They will be removed the next time someone calls the API generator.

I will provide a suggestion t as an alternate fix.

File: src/drivers/drivers.jl [hunk 1]
```diff
@@ -79,10 +79,39 @@ function set_driver!(fapl::Properties, ::POSIX)
...
+
+cconvert(s::AbstractString, N::Integer) = Tuple(i ≤ length(s) ? s[i] : '\0' for i in 1:N)
+
+function get_driver(fapl::Properties, ::Type{ROS3})
+    ROS3()
+end
+
+function set_driver!(fapl::Properties, driver::ROS3)
+    HDF5.init!(fapl)
+    API.h5p_set_fapl_ros3(fapl, driver.fa)
```
mkitti, line 105:
```suggestion
    API.h5p_set_fapl_ros3(fapl, Ref(driver.fa))
```

src/drivers/drivers.jl
```diff
@@ -79,10 +79,39 @@ function set_driver!(fapl::Properties, ::POSIX)
...
+    1, true, cconvert(region, 33), cconvert(id, 129), cconvert(key, 129)
+)
+ROS3() = (ROS3 ∘ API.H5FD_ros3_fapl_t)(
+    1, false, cconvert("", 33), cconvert("", 129), cconvert("", 129)
+)
+
+cconvert(s::AbstractString, N::Integer) = Tuple(i ≤ length(s) ? s[i] : '\0' for i in 1:N)
+
+function get_driver(fapl::Properties, ::Type{ROS3})
+    ROS3()
```
mkitti, line 100:
```suggestion
    r_fa = Ref{H5FD_ros3_fapl_t}()
    H5Pget_fapl_ros3(fapl, r_fa)
    return ROS3(r_fa[])
```

mkitti (changes requested):
Here are some suggestions.

Please regenerate functions.jl.

src/drivers/drivers.jl
```diff
@@ -79,10 +79,39 @@ function set_driver!(fapl::Properties, ::POSIX)
...
+    fa::API.H5FD_ros3_fapl_t
+end
+ROS3(region::T, id::T, key::T) where {T<:AbstractString} = (ROS3 ∘ API.H5FD_ros3_fapl_t)(
+    1, true, cconvert(region, 33), cconvert(id, 129), cconvert(key, 129)
+)
+ROS3() = (ROS3 ∘ API.H5FD_ros3_fapl_t)(
+    1, false, cconvert("", 33), cconvert("", 129), cconvert("", 129)
+)
+
+cconvert(s::AbstractString, N::Integer) = Tuple(i ≤ length(s) ? s[i] : '\0' for i in 1:N)
```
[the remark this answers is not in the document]
gszep: removing the need for this by using `Cstring` type

src/drivers/drivers.jl
```diff
@@ -79,10 +79,39 @@ function set_driver!(fapl::Properties, ::POSIX)
...
+"""
+struct ROS3 <: Driver
+    fa::API.H5FD_ros3_fapl_t
+end
+ROS3(region::T, id::T, key::T) where {T<:AbstractString} = (ROS3 ∘ API.H5FD_ros3_fapl_t)(
+    1, true, cconvert(region, 33), cconvert(id, 129), cconvert(key, 129)
+)
+ROS3() = (ROS3 ∘ API.H5FD_ros3_fapl_t)(
+    1, false, cconvert("", 33), cconvert("", 129), cconvert("", 129)
+)
```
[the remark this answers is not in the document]
gszep: this is required for the case with no authentication info

src/api/types.jl
```diff
@@ -222,6 +222,14 @@ struct H5FD_hdfs_fapl_t
     stream_buffer_size::Int32
 end
 
+struct H5FD_ros3_fapl_t
+    version::Int32
+    authenticate::hbool_t
+    aws_region::NTuple{33,Cchar}
+    secret_id::NTuple{129,Cchar}
+    secret_key::NTuple{129,Cchar}
```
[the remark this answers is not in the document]
gszep: agreed. using `Cstring` type instead

src/drivers/drivers.jl
```diff
@@ -79,10 +79,39 @@ function set_driver!(fapl::Properties, ::POSIX)
...
 
+"""
+    ROS3()
+
+This is the read-only virtual driver that enables access to HDF5 objects stored in AWS S3
+"""
+struct ROS3 <: Driver
+    fa::API.H5FD_ros3_fapl_t
+end
+ROS3(region::T, id::T, key::T) where {T<:AbstractString} = (ROS3 ∘ API.H5FD_ros3_fapl_t)(
```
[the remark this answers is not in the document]
gszep: agreed. applied changes

src/file.jl
```diff
@@ -42,9 +42,6 @@ function h5open(
         flag = swmr ? API.H5F_ACC_TRUNC | API.H5F_ACC_SWMR_WRITE : API.H5F_ACC_TRUNC
         fid = API.h5f_create(filename, flag, fcpl, fapl)
     else
-        ishdf5(filename) || error(
-            "unable to determine if $filename is accessible in the HDF5 format (file may not exist)"
-        )
```
mkitti, line 47: It would be great if you could add another condition to the `if` statement rather than deleting this. If you cannot just comment it out, and I can look into it later.

test/plain.jl
```diff
@@ -1546,7 +1546,6 @@ end
     @test !HDF5.ishdf5(fn1) # check that a non-hdf5 file retuns false
     @test !HDF5.ishdf5(fn2) # checks that a file that does not exist returns false
 
-    @test_throws ErrorException h5write(fn1, "x", 1) # non hdf5 file throws
```
mkitti, line 1549: Same as the other comment. Comment this out or consider checking to see if the file is actually writable.

test/ros3.jl
```diff
@@ -0,0 +1,10 @@
+using HDF5
+using Test
+
+@testset "ros3" begin
+    @test HDF5.has_ros3()
+
+    h5open("http://s3.us-east-2.amazonaws.com/hdf5ros3/GMODO-SVM01.h5"; driver=HDF5.Drivers.ROS3()) do f
+        @test keys(f) == ["All_Data", "Data_Products"]
+    end
```
mkitti, line 9: Nice. Can you add a test to see if you can actually read data correctly?

docs/src/api_bindings.md
```diff
@@ -477,6 +477,7 @@ h5pl_size
 - [`h5p_get_fapl_mpio32`](@ref h5p_get_fapl_mpio32)
 - [`h5p_get_fapl_mpio64`](@ref h5p_get_fapl_mpio64)
 - [`h5p_get_fapl_multi`](@ref h5p_get_fapl_multi)
+- [`h5p_get_fapl_ros`](@ref h5p_get_fapl_ros)
```
mkitti, line 480: Hmm... I think we may need to the patch the generator.

mkitti (changes requested):
Just a few changes. If you could just comment out the failing tests and mark it for later that would be fine. It would be even better if you could narrow the situation to remove the test / error case.

docs/src/api_bindings.md
gszep: agreed. the name should be `h5p_get_fapl_ros3`

src/drivers/ros3.jl
```diff
@@ -7,11 +7,24 @@ struct ROS3 <: Driver
...
+    version::Integer,
+    auth::Bool,
+    region::AbstractString,
+    id::AbstractString,
+    key::AbstractString
+)
+    return (ROS3 ∘ API.H5FD_ros3_fapl_t)(
+        version,
+        auth,
+        Base.unsafe_convert(Cstring, region),
```
musm, line 20: Can this be a regular convert or does it have to be an `unsafe_convert`  ?

mkitti: It needs to be an `unsafe_convert`:

```
julia> convert(Cstring, "hello")
ERROR: MethodError: Cannot `convert` an object of type String to an object of type Cstring
Closest candidates are:
  convert(::Type{Cstring}, ::Union{Ptr{Int8}, Ptr{Nothing}, Ptr{UInt8}}) at ~/src/julia-1.7.1/share/julia/base/c.jl:167
  convert(::Type{T}, ::T) where T at ~/src/julia-1.7.1/share/julia/base/essentials.jl:218
Stacktrace:
 [1] top-level scope
   @ REPL[55]:1

julia> Base.unsafe_convert(Cstring, "hello")
Cstring(0x00007fc5c5e06838)
```

docs/src/api_bindings.md
mkitti: I added a bind exception for ros3 in cc6625e

mkitti:
I think this is about ready now. Is it?

mkitti:
@simonbyrne , could you take a look at this pull request?

mkitti:
@gszep is there any reason this is still in draft form?

mkitti:
I think we should move forward with this.

src/drivers/ros3.jl
simonbyrne: Probably the best option is to add them as extra fields to the ROS3 struct?

src/drivers/ros3.jl
```diff
@@ -0,0 +1,45 @@
+"""
+    ROS3()
```
simonbyrne, line 2: Update the docstring signature

src/drivers/ros3.jl
mkitti: Couldn't we just do `unsafe_convert(Cstring, String(region))`?

simonbyrne: > Couldn't we just do `unsafe_convert(Cstring, String(region))`?

Yes, it's equivalent:
https://github.com/JuliaLang/julia/blob/99225ab1d11925b8e904b0b12f1ee8b0f6946660/base/c.jl#L196-L198
but we have the API, so you should use it.

Either way, you still need to keep a reference to the resulting `String`.

musm: @simonbyrne  is this what you mean? 
```julia
function ROS3(
    version::Integer,
    auth::Bool,
    region::AbstractString,
    id::AbstractString,
    key::AbstractString
)

    regionbuf = Base.cconvert(Cstring, region)
    idbuf = Base.cconvert(Cstring, id)
    keybuf = Base.cconvert(Cstring, key)

    GC.@preserve regionbuf idbuf keybuf begin
         regionstr = Base.unsafe_convert(Cstring, regionbuf)
         idstr = Base.unsafe_convert(Cstring, idbuf)
         keystr = Base.unsafe_convert(Cstring, keybuf)
         ROS3(API.H5FD_ros3_fapl_t(version, auth, regionstr, idstr, keystr))
    end

end
```

mkitti: That still is not going to work. We need to preserve `regionbuf`, `idbuf`, and `keybuf`.

mkitti: I think we need this:

```julia
struct ROS3 <: Driver
    version::Int32
    authenticate::hbool_t
    aws_region::String
    secret_id::String
    secret_key::String
end

function Base.unsafe_convert(API.H5FD_ros3_fapl_t, driver::ROS3)
    API.H5FD_ros3_fapl_t(
        driver.version,
        driver.authenticate,
        Base.unsafe_convert(Cstring, driver.aws_region),
        Base.unsafe_convert(Cstring, driver.secret_id),
        Base.unsafe_convert(Cstring, driver.secret_key),
    )
end
```

`ccall` then should then make sure that ROS3 instance and its member Strings are preserved.

musm: I think I missed the original point of this all. Please correct me if I'm still wrong in my thinking  
So the reason we need this is because we  use the ROS3 instance in  further calls  (specifically `h5p_set_fapl_ros3` `h5p_get_fapl_ros3`) and those need to convert the outward driver struct `ROS3` to the internal libhdf5  struct `H5FD_ros3_fapl_t` and right now there's a mismatch in that `ROS3` doesn't agree with the layout of  `H5FD_ros3_fapl_t`?

mkitti: We need to convert the Julia struct to the C struct. Part of the issue is that the Julia struct and its Strings could be garbage collected. Meanwhile, the Cstring is an unmanaged pointer to a null terminated character array.

Anyways, it turns out that the C struct definition here is wrong. The correct definition is

https://github.com/HDFGroup/hdf5/blob/100b22e6c23c44a082fd69b8c05a63c7492083f7/src/H5FDros3.h#L82-L88
```c
typedef struct H5FD_ros3_fapl_t {
    int32_t version;
    hbool_t authenticate;
    char    aws_region[H5FD_ROS3_MAX_REGION_LEN + 1];
    char    secret_id[H5FD_ROS3_MAX_SECRET_ID_LEN + 1];
    char    secret_key[H5FD_ROS3_MAX_SECRET_KEY_LEN + 1];
} H5FD_ros3_fapl_t;
```

In Julia, that would be
```julia
struct H5FD_ros3_fapl_t
    version::Int32
    authenticate::Bool
    aws_region::NTuple{H5FD_ROS3_MAX_REGION_LEN + 1,Cchar}
    secret_id ::NTuple{H5FD_ROS3_MAX_SECRET_ID_LEN + 1, Cchar}
    secret_key::NTuple{H5FD_ROS3_MAX_SECRET_KEY_LEN + 1, Cchar};
end
```

I've started to rewrite this. Also, I think we need to check our definition of `hbool_t`. Currently it is `UInt32`, but I think it probably should be `UInt8`.

musm: That's probably not the first time there has been a mismatch in definitions. Do you want to push your changes on this branch?

mkitti: Yes, I will later today. I just started it on a different computer than I am on right now. I didn't quite get the tests to pass yet on the branch I was working on.

simonbyrne:
Do we have a way to test this?

mkitti:
There is a test already, but without authentication.

There seems to be an example here:
https://www.hdfgroup.org/solutions/cloud-amazon-s3-storage-hdf5-connector/

src/drivers/ros3.jl
mkitti: I added additonal signatures below this line.

src/drivers/ros3.jl
```diff
@@ -4,28 +4,50 @@
...
+            "length(secret_key), $(length(secret_key)), must be less than or equal to $(API.H5FD_ROS3_MAX_SECRET_KEY_LEN)"
+        new(version, authenticate, aws_region, secret_id, secret_key)
+    end
 end
 
+ROS3() = ROS3(1, false, "", "", "")
 ROS3(region::AbstractString, id::AbstractString, key::AbstractString) =
     ROS3(1, true, region, id, key)
-ROS3() = ROS3(1, false, "", "", "")
+_ntuple_to_string(x) = unsafe_string(Ptr{Cchar}(pointer_from_objref(Ref(x))), length(x))
```
musm: This doesn't convert the ntuple to a string back for me
```julia
_ntuple_to_string(x) = unsafe_string(Ptr{Cchar}(pointer_from_objref(Ref(x))), length(x))
x = "Dfsddfs"
a = ntuple(i -> i <= length(x) ? Cchar(x[i]) : Cchar(0x0), 10 + 1)
julia> _ntuple_to_string(a)
"\xc8\x18\xb2\n\0\0\0\0X\xd0\xc3"
```
Shouldn't this actually look like something like the following
```julia
function _ntuple_to_string(buf)
    len = findfirst(iszero, buf) -1
    unsafe_string(Base.unsafe_convert(Ptr{Int8}, collect(buf[1:len])))
end
```

src/drivers/ros3.jl
musm: I quickly saw this comment on the hbool issue https://github.com/aldanor/hdf5-rust/issues/28
but perhaps it doesn't affect us since we dont' use msvc

src/drivers/ros3.jl
mkitti: unsafe_string already searches for a null itself.

```
julia> function _ntuple_to_string(x)
           r = Ref(x)
           GC.@preserve r unsafe_string(Ptr{Cchar}(pointer_from_objref(r)))
       end
_ntuple_to_string (generic function with 1 method)

julia> _ntuple_to_string(a)
"Dfsddfs"
```

mkitti: We also probably should use `ncodeunits` and `codeunits` like you suggested before:
https://github.com/JuliaIO/HDF5.jl/issues/818#issuecomment-780071965

simonbyrne: Why not use `FixedString`, which is already in the package?
https://github.com/JuliaIO/HDF5.jl/blob/0120cb2d32fc8d57727b3dcb17860c63c00a0cd3/src/typeconversions.jl#L99-L105

mkitti: FixedString is not an AbstractString.

I started working on a version of InlineStrings.jl called NStrings.jl over the weekend that basically is a full blown general implementation of this and is an AbstractString.

I'll publish that repo soon.

simonbyrne: I was actually about to suggest moving it to a separate repo! Nice!

mkitti: The initial version of NTuple based Strings is here:
https://github.com/mkitti/NStrings.jl

mkitti: The current NTuple based Strings are now here and registered:

https://github.com/mkitti/StaticStrings.jl

mkitti:
Let's merge just merge this. I'll follow up with a StaticStrings.jl pull request.

mkitti (approved):

Issue #859, opened 2021-08-04
Empty (extensible) dataset cannot be read back


tamasgal:
I just stumbled upon the issue that `HDF5.jl` chokes on datasets which were initialised to 0-size but never resized or populated later.
A workaround is of course to not create datasets which are empty but currently I set-up my data processing in a homogenous way, so that each HDF5 file has the same structure and sometimes a dataset is empty.

Maybe I am doing something wrong and this is not the way to initialise datasets which might be empty, but that's what I found in the HDF5 docs and also in examples.

Here is a MWE, just uncomment the two lines and the error disappears:

```julia
using HDF5

fname = tempname()
groupname = "a_group"
dsetname = "foo"

h5open(fname, "w") do fid
    create_group(fid, groupname)
    create_dataset(fid[groupname], dsetname, datatype(Float32), ((0,), (-1,)), chunk=(100,))
    
    # Uncomment these otherwise the datasets are unreadable
    #dset = fid[groupname][dsetname]
    #HDF5.set_extent_dims(dset, (100,))
end
    
h5open(fname, "r") do fid
    fid[groupname][dsetname][:]
end
```

Here is the error which indicates that the binary parsing loses track:

```
UInt64[0x0000000000000000]
UInt64[0x0000000000000001]
UInt64[0x0000000000000000]
(0,)
index out of range

Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] hyperslab(dspace::HDF5.Dataspace, I::Base.Slice{Base.OneTo{Int64}})
    @ HDF5 ~/.julia/packages/HDF5/0iEnL/src/HDF5.jl:1691
  [3] generic_read(obj::HDF5.Dataset, filetype::HDF5.Datatype, #unused#::Type{Float32}, I::Function)
    @ HDF5 ~/.julia/packages/HDF5/0iEnL/src/HDF5.jl:1252
  [4] getindex(dset::HDF5.Dataset, I::Function)
    @ HDF5 ~/.julia/packages/HDF5/0iEnL/src/HDF5.jl:1188
  [5] (::var"#77#78")(fid::HDF5.File)
    @ Main ./In[28]:17
  [6] h5open(::var"#77#78", ::String, ::Vararg{String, N} where N; swmr::Bool, pv::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
[...]

PR #1011, opened 2022-09-12
add BlockRange object to represent a dimension of a hyperslab

simonbyrne:
This defines a `BlockRange` object which losslessly represents a single dimension of a hyperslab selection. It roughly corresponds to a [`MultiBlockSlice` object in h5py](https://docs.h5py.org/en/stable/high/dataset.html#multi-block-selection).

Additionally it makes a few more tweaks:
- converting an `AbstractUnitRange` to a `BlockRange` will now use `count=1, block=length(r)` instead of `stride=1, count=length(r), block=1`. This should be equivalent, but there is some suggestion (e.g. [here](https://portal.hdfgroup.org/display/knowledge/Hints+to+using+Parallel+HDF5?preview=%2F48809760%2F48809759%2Fparallelhdf5hints.pdf#)) that it might be more efficient.
- `HDF5.get_regular_hyperslab` now returns a tuple of `BlockRange` objects: it is an breaking change, but the function isn't exported: if you would prefer not to break it, we can perhaps call it something else (suggestions?)
- `count` can now be `-1` (unlimited). This is used by the virtual datasets feature: previously this would give conversion errors.
- added `select_hyperslab!` as a mutating operation on a `Dataspace`, which now supports an `op` argument.
- `hyperslab` is now simplified considerably.

Let me know if you're happy with the design, and I can add a couple of extra unit tests.

### DIFF
--- a/src/dataspaces.jl
+++ b/src/dataspaces.jl
@@ -177,36 +177,6 @@ BlockRange(br::BlockRange) = br
 Base.to_index(d::Dataset, br::BlockRange) = br
 Base.length(br::BlockRange) = Int(br.count * br.block)
 
-function Base.show(io::IO, br::BlockRange)
-    start = Int(br.start0 + 1)
-    # choose the simplest possible representation
-    if br.count == 1
-        if br.block == 1
-            # integer
-            print(io, start)
-        else
-            # UnitRange
-            print(io, range(start; length=Int(br.block)))
-        end
-    elseif br.block == 1 && br.count != API.H5S_UNLIMITED
-        # StepRange
-        print(io, range(start; step=Int(br.stride), length=Int(br.count)))
-    else
-        # BlockRange(int; ...)
-        print(io, BlockRange, "(start=", start)
-        if br.stride != 1
-            print(io, ", stride=", Int(br.stride))
-        end
-        if br.count != 1
-            print(io, ", count=", br.count == API.API.H5S_UNLIMITED ? -1 : Int(br.count))
-        end
-        if br.block != 1
-            print(io, ", block=", Int(br.block))
-        end
-        print(io, ")")
-    end
-end
-
 function Base.range(br::BlockRange)
     start = Int(br.start0 + 1)
     if br.count == 1
@@ -230,7 +200,7 @@ integers, ranges or [`blockrange`](@ref) objects.
 - `op` determines how the new selection is to be combined with the already
   selected dataspace:
   - `:select` (default): replace the existing selection with the new selection.
-  - `:or`: adds the new selection to the existing selection. 
+  - `:or`: adds the new selection to the existing selection.
      Aliases: `|`, `∪`, `union`.
   - `:and`: retains only the overlapping portions of the new and existing
     selection. Aliases: `&`, `∩`, `intersect`.
@@ -244,9 +214,7 @@ integers, ranges or [`blockrange`](@ref) objects.
 
 """
 function select_hyperslab!(
-    dspace::Dataspace,
-    op::Union{Symbol,typeof.((&, |, ⊻, ∪, ∩, setdiff))...},
-    idxs::Tuple
+    dspace::Dataspace, op::Union{Symbol,typeof.((&, |, ⊻, ∪, ∩, setdiff))...}, idxs::Tuple
 )
     N = ndims(dspace)
     length(idxs) == N || error("Number of indices does not match dimension of Dataspace")
@@ -259,11 +227,11 @@ function select_hyperslab!(
 
     _op = if op == :select
         API.H5S_SELECT_SET
-    elseif (op == :or || op === | || op === ∪)
+    elseif (op == :or || op === (|) || op === (∪))
         API.H5S_SELECT_OR
-    elseif (op == :and || op === & || op === ∩)
+    elseif (op == :and || op === (&) || op === (∩))
         API.H5S_SELECT_AND
-    elseif (op == :xor || op === ⊻)
+    elseif (op == :xor || op === (⊻))
         API.H5S_SELECT_XOR
     elseif op == :notb || op === setdiff
         API.H5S_SELECT_NOTB
--- a/src/dataspaces.jl
+++ b/src/dataspaces.jl
@@ -121,51 +121,148 @@ function isnull(dspace::Dataspace)
     return API.h5s_get_simple_extent_type(checkvalid(dspace)) == API.H5S_NULL
 end
 
+"""
+    HDF5.get_regular_hyperslab(dspace)::Tuple
+
+Get the hyperslab selection from `dspace`. Returns a tuple of [`BlockRange`](@ref) objects.
+"""
 function get_regular_hyperslab(dspace::Dataspace)
-    start, stride, count, block = API.h5s_get_regular_hyperslab(dspace)
-    N = length(start)
-    @inline rev(v) = ntuple(i -> @inbounds(Int(v[N - i + 1])), N)
-    return rev(start), rev(stride), rev(count), rev(block)
+    start0, stride, count, block = API.h5s_get_regular_hyperslab(dspace)
+    N = length(start0)
+    ntuple(N) do i
+        ri = N - i + 1
+        @inbounds BlockRange(start0[ri], stride[ri], count[ri], block[ri])
+    end
 end
 
-function hyperslab(dspace::Dataspace, I::Union{AbstractRange{Int},Int}...)
-    dims = size(dspace)
-    n_dims = length(dims)
-    if length(I) != n_dims
-        error(
-            "Wrong number of indices supplied, supplied length $(length(I)) but expected $(n_dims)."
-        )
+struct BlockRange
+    start0::API.hsize_t
+    stride::API.hsize_t
+    count::API.hsize_t
+    block::API.hsize_t
+end
+
+"""
+    BlockRange(;start::Integer, stride::Integer=1, count::Integer=1, block::Integer=1)
+
+A `BlockRange` represents a selection along a single dimension of a HDF5
+hyperslab. It is similar to a Julia `range` object, with some extra features for
+selecting multiple contiguous blocks.
+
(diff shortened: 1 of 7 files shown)


simonbyrne:
I also found [this discussion](https://forum.hdfgroup.org/t/hyperslabs-count-vs-blocks/4977?u=simonbyrne) on `block` vs `count`: it says they should be equivalent, but using `block` returns a single value via [`H5Sget_select_hyper_nblocks`](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5S.html#Dataspace-SelectHyperNBlocks)

mkitti:
Should BlockRange be an AbstractRange?

mkitti:
Perhaps we could offer a `convert(AbstractRange, bl::BlockRange)` that may or may not work?

simonbyrne:
> Perhaps we could offer a `convert(AbstractRange, bl::BlockRange)` that may or may not work?

Yeah, that's a good idea.

simonbyrne:
added conversions and tests

simonbyrne:
This fixes #859 (added a test)

src/dataspaces.jl
```diff
@@ -121,51 +121,165 @@ function isnull(dspace::Dataspace)
...
+        # UnitRange
+        return range(start; length=Int(br.block))
+    elseif br.block == 1 && br.count != API.H5S_UNLIMITED
+        # StepRange
+        return range(start; step=Int(br.stride), length=Int(br.count))
+    else
+        error("$br cannot be converted to a Julia range")
+    end
+end
+Base.convert(::Type{AbstractRange}, br::BlockRange) = range(br)
```
mkitti, line 222:
```suggestion
Base.convert(::Type{T <: AbstractRange}, br::BlockRange)::T where T = range(br)
```

This would allow `Base.convert(UnitRange, ...)` and `Base.convert(StepRange, ...)` as well. Could we also add a few conversion tests?

simonbyrne: i'm not sure that will work generally?

simonbyrne: or i'm not sure what you had in mind.

mkitti: Currently `convert(AbstractRange, BlockRange(0, 2, 2, 1))` works. I would also like `convert(StepRange, BlockRange(0, 2, 2, 1))` or `convert(UnitRange, BlockRange(0, 1, 1, 10))` to work.

This would allow something like the following to work.
```
julia> ranges = OrdinalRange[]
OrdinalRange[]

julia> push!(ranges, BlockRange(0, 2, 2, 1))
1-element Vector{OrdinalRange}:
 1:2:3

julia> push!(ranges, BlockRange(0, 1, 1, 10))
2-element Vector{OrdinalRange}:
 1:2:3
 1:10
```

Meanwhile, this should fail:

```julia
julia> convert(UnitRange, BlockRange(0, 2, 2, 1))
ERROR: TypeError: in typeassert, expected UnitRange, got a value of type StepRange{Int64, Int64}
Stacktrace:
 [1] convert(#unused#::Type{UnitRange}, br::BlockRange)
   @ Main .\REPL[22]:1
 [2] top-level scope
   @ REPL[51]:1
```

mkitti:
Looks like we just need to run the formatter. Would you like me to do that for you?

simonbyrne:
Yea please!

src/dataspaces.jl
```diff
@@ -121,51 +121,165 @@ function isnull(dspace::Dataspace)
...
+  selected dataspace:
+  - `:select` (default): replace the existing selection with the new selection.
+  - `:or`: adds the new selection to the existing selection
+  - `:and`: retains only the overlapping portions of the new and existing
+    selection.
+  - `:xor`: retains only the elements that are members of the new selection or
+    the existing selection, excluding elements that are members of both
+    selections.
+  - `:notb`: retains only elements of the existing selection that are not in the
+    new selection.
```
mkitti, line 240:
```suggestion
  - `:or`: adds the new selection to the existing selection. 
     Aliases: `|`, `∪`, `union`.
  - `:and`: retains only the overlapping portions of the new and existing
    selection. Aliases: `&`, `∩`, `intersect`.
  - `:xor`: retains only the elements that are members of the new selection or
    the existing selection, excluding elements that are members of both
    selections. Aliases: `⊻`, `xor`
  - `:notb`: retains only elements of the existing selection that are not in the
    new selection. Alias: `setdiff`.
```

src/dataspaces.jl
```diff
@@ -121,51 +121,165 @@ function isnull(dspace::Dataspace)
...
+    _count = API.hsize_t[blockranges[N - i + 1].count for i in 1:N]
+    _block = API.hsize_t[blockranges[N - i + 1].block for i in 1:N]
+
+    _op = if op == :select
+        API.H5S_SELECT_SET
+    elseif (op == :or || op === |)
+        API.H5S_SELECT_OR
+    elseif (op == :and || op === &)
+        API.H5S_SELECT_AND
+    elseif (op == :xor || op === xor)
```
mkitti, line 263:
```suggestion
    elseif (op == :xor || op === ⊻)
```

src/dataspaces.jl
```diff
@@ -121,51 +121,165 @@ function isnull(dspace::Dataspace)
...
+        API.H5S_SELECT_XOR
+    elseif op == :notb
```
mkitti, line 265:
```suggestion
    elseif op == :notb || op === setdiff
```

src/dataspaces.jl
```diff
@@ -121,51 +121,165 @@ function isnull(dspace::Dataspace)
...
+    start = Int(br.start0 + 1)
+    # choose the simplest possible representation
+    if br.count == 1
+        if br.block == 1
+            # integer
+            print(io, start)
         else
-            error("index must be range or integer")
+            # UnitRange
+            print(io, range(start; length=Int(br.block)))
```
mkitti, line 189:
```suggestion
            print(io, "BlockRange(")
            print(io, range(start; length=Int(br.block)))
            print(io, ")")
```

src/dataspaces.jl
```diff
@@ -121,51 +121,165 @@ function isnull(dspace::Dataspace)
...
+        end
+    elseif br.block == 1 && br.count != API.H5S_UNLIMITED
+        # StepRange
+        print(io, range(start; step=Int(br.stride), length=Int(br.count)))
```
mkitti, line 193:
```suggestion
        print(io, "BlockRange(")
        print(io, range(start; step=Int(br.stride), length=Int(br.count)))
        print(io, ")")
```

src/dataspaces.jl
```diff
@@ -121,51 +121,165 @@ function isnull(dspace::Dataspace)
...
+end
+BlockRange(start::Integer; stride=1, count=1, block=1) =
+    BlockRange(; start=start, stride=stride, count=count, block=block)
+BlockRange(r::AbstractUnitRange; stride=max(length(r), 1), count=1) =
+    BlockRange(; start=first(r), stride=stride, count=count, block=length(r))
+BlockRange(r::OrdinalRange) = BlockRange(; start=first(r), stride=step(r), count=length(r))
+BlockRange(br::BlockRange) = br
+
+Base.to_index(d::Dataset, br::BlockRange) = br
+Base.length(br::BlockRange) = Int(br.count * br.block)
```
mkitti, line 178:
```suggestion
Base.length(br::BlockRange) = Int(br.count * br.block)
Base.first(br::BlockRange) = br.start0 + 1
Base.stride(br::BlockRange) = br.stride
```

Will `length` be negative if `count` is negative?

mkitti (changes requested):
A few more items:
* While the simplified showing of the BlockRange as UnitRange or StepRange is nice, I think we should still make it clear that it is a BlockRange. I suggest `BlockRange(1:5)` and `BlockRange(2:3:8)` as the display in those cases.
* Could we implement `first`, `stride`, and maybe even `last` for `BlockRange`?
* Allowing for the operators to actually be Julia operators is a nice touch. I added suggestions to expand this.
    * `:or` could also be "union" or `∪` (`\cup`)
    * `:and` could also be "intersect" or `∩` (`\cap`)
    * `:xor` also has a symbol `⊻` (`\xor`)
    * `:notb` could also be `setdiff`

src/dataspaces.jl
simonbyrne: [`Base.stride`](https://docs.julialang.org/en/v1/base/arrays/#Base.stride) has a different meaning. I assume you mean [`Base.step`](https://docs.julialang.org/en/v1/base/collections/#Base.step)? The problem is that it doesn't make sense in the case where `block > 1` (since stride refers to the spacing between blocks, not spacing between elements)

`count` can't be negative (they're unsigned integers). It can be `API.H5S_UNLIMITED`, but this will throw a conversion error.

src/dataspaces.jl
simonbyrne: This will change the printing of Dataspaces, but perhaps we could make that a `compact` option?

simonbyrne: I've added this.

src/dataspaces.jl
simonbyrne: We could define `Base.first`, but I'm not really sure how useful it would be.

mkitti: `Base.first` is the most useful since it requires some calculation to translate to 1-based indexing. I did mean `stride`. I'm not sure if actually carries any direct meaning here since this is not an `AbstractArray`. Perhaps it should not be `Base.stride` though.

The overall question here is if you want to add a functional interface as a layer of abstraction or should users just rely on accessing the properties directly. It looks to me like there are multiple kinds of `BlockRange`s here which may become distinct types in some future refactoring. We already seem to effectively have `UnitBlockRange` and `StepBlockRange` that are mappable to `UnitRange` and `StepRange`.

mkitti (approved):

simonbyrne:
Would be good to merge this so I can refine #1012.

Issue #819, opened 2021-02-16
No Intuitive Way to Write Compound Data Types


nicrummel:
While it is incredibly easy to read HDF5 files and access the data within (no matter the complexity of the datatypes and structure of the files). After hours of looking at documentation and reading the source code, I still do not see a straight forward way to **write** compound data types to a file. 

A simple task to demonstrate where the api is failing me is the the following:

> Take two hdf5 datasets with exactly the same structure, and 'concatenate' the two maintaining the original format.

```
dset1 = | Col1 (Int32) | Col2 (String) | # This syntax is just supposed to visualize the compound datatype
dset2 = | Col1 (Int32) | Col2 (String) | # Rather than to show actual output of Julia REPL
```
I can simply read in the data with 
```
using HDF5
filename1, filename2, datasetName, newFilename = # Strings
h5_file1, h5_file2 = h5open(filename1), h5open(filename2) # load files 
h5_dict1,  h5_dict2 = read(h5_file1), read(h5_file2) # Turn HDF5.Files into Dictionaries
dset1, dset2 = h5_dict1[datasetName ], h5_dict2[datasetName] # Returns a Vector{NamedTuple} 
                                                             # with names specifying the Col
# Try to concatenate datasets
h5open(newFilename,"w") do wf
    create_dataset(wf, 
                   datasetName , 
                   datatype(h5_file1[datasetName]), # obtain the compound datatype for dataset
                   (length(dset1) + length(dset2),)) # pre-allocate space for the combined datasets 
    for (i, datum) in enumerate(dset1) 
        wf[datasetName][i] = datum # this causes an error
    end
    for (i, datum) in enumerate(dset2) 
        wf[datasetName][i + length(dset1)] = datum # this causes an error
    end
end

```
Here is the error message

```
ERROR: MethodError: no method matching setindex!(::HDF5.Dataset, ::NamedTuple{(Col1, Col2), (Int32, String)}, Int64}
```

I believe that there is a way to write the compound datasets, but I cannot find it anywhere. Could someone enlighten me?

musm:
Thanks for the issue.

Does this example help at all?
```julia


using HDF5

struct Foo
    a::Int32
    b::String
end

struct Foo_hdf5
    a::Int32
    b::Cstring
end

function Base.unsafe_convert(::Type{Foo_hdf5}, x::Foo)
    Foo_hdf5(x.a, Base.unsafe_convert(Cstring, x.b))
end

[...]

PR #1013, opened 2022-09-27
Add built-in support to write compound data types

tamasgal:
This PR contains a utility function (EDIT: actually after some discussions it's now realised via adding a method to `datatype()`) which I carry around my projects and I discovered that there is already an open issue which is related too, see [#819](https://github.com/JuliaIO/HDF5.jl/issues/819)

It's by all means not fully fledged, but it works for `struct`s with primitive field types very well. It's not reversible, meaning that `read` will yield the usual `NamedTuple`. One way to make it reversible would be to attach metadata to the dataset with e.g. the type name and when read back, check if the type is already defined in the namespace and if not, create it dynamically. Kind of what I do in https://github.com/JuliaHEP/UnROOT.jl where generating types during runtime is mandatory.

I provided an example in the function docstring which demonstrates that it's fairly intuitive to use and covers the basic needs. Here is it:

```julia
julia> struct Foo
           x::Int32
           y::Float64
       end

julia> foos = [[Foo(1, 2) Foo(3, 4) Foo(5, 6)]; [Foo(7, 8) Foo(9, 10) Foo(11, 12)]]
2×3 Matrix{Foo}:
 Foo(1, 2.0)  Foo(3, 4.0)   Foo(5, 6.0)
 Foo(7, 8.0)  Foo(9, 10.0)  Foo(11, 12.0)

julia> h5open("foo.h5", "w") do h5f
           write_dataset(h5f, "the/foo", foos)
       end

julia> thefoo = h5open("foo.h5", "r") do file
           read(file, "the/foo")
       end
2×3 Matrix{NamedTuple{(:x, :y), Tuple{Int32, Float64}}}:
 (x = 1, y = 2.0)  (x = 3, y = 4.0)   (x = 5, y = 6.0)
 (x = 7, y = 8.0)  (x = 9, y = 10.0)  (x = 11, y = 12.0)
```

The data can be reinterpreted to the original data type using `reinterpret()`:

```julia
julia> reinterpret(Foo, thefoo)
2×3 reinterpret(Foo, ::Matrix{NamedTuple{(:x, :y), Tuple{Int32, Float64}}}):
 Foo(1, 2.0)  Foo(3, 4.0)   Foo(5, 6.0)
 Foo(7, 8.0)  Foo(9, 10.0)  Foo(11, 12.0)
```


src/datasets.jl
```diff
@@ -482,6 +482,57 @@ write_dataset(
...
+
+```
+julia> reinterpret(Foo, thefoo)
+2×3 reinterpret(Foo, ::Matrix{NamedTuple{(:x, :y), Tuple{Int32, Float64}}}):
+ Foo(1, 2.0)  Foo(3, 4.0)   Foo(5, 6.0)
+ Foo(7, 8.0)  Foo(9, 10.0)  Foo(11, 12.0)
+```
+
+"""
+function write_compound(f, name, data::AbstractArray{T}) where {T}
```
mkitti, line 526: Can you add some more type information to this?

Also, does the compound type need to be in an AbstractArray?

tamasgal: Yes sure, I'll restrict to `AbstractString`. I thought it makes more sense to write arrays of compound structures. I'll have a look!

tamasgal: I am a bit confused. Is it a common use-case to store a single struct in a dataset?

mkitti: It can be, especially if we are writing the data type to an attribute. For example, I encountered one case where we are encapsulating a foreign header as a 1024 byte compound type:
https://github.com/mkitti/FIBSEMHeaders.jl/blob/eded59eac531fe4610edc71453c1ace7fc1039f9/src/FIBSEMHeaders.jl#L26

It seems this assumes we are writing compound datasets to a dataset.

src/HDF5.jl
```diff
@@ -28,6 +28,7 @@ export @read,
     open_dataset,
     read_dataset,
     write_dataset,
+    write_compound,
```
mkitti, line 31:
```suggestion
    write_compound_dataset,
```

mkitti (changes requested):
At minimum I think this should be `write_compound_dataset` since one could also write a compound type to an attribute.

My preference would be to extend `HDF5.datatype` to work on arbitrary structs and named tuples so that we can just use the normal `write_dataset` interface. After enabling that, then we should reevaluate if a special convenience method is still warranted.

src/datasets.jl
tamasgal: OK yes, you are right, I will have a look ;)

tamasgal:
Let me comment at the end since I changed a few things.

- `write_compound_dataset()` is now removed
- `datatype()` has now a new method with `datatype(A::AbstractArray{T}) where {T}`

so that `write_dataset()` now automatically works with an array of (flat) structs.

@mkitti I am still not sure about the type freedom, so let me know what you think!

src/typeconversions.jl
```diff
@@ -60,6 +60,15 @@ function datatype(str::VLen{C}) where {C<:CharType}
     Datatype(API.h5t_vlen_create(type_id))
 end
 
+# Compound types
+function datatype(A::AbstractArray{T}) where {T}
```
mkitti, line 64: Since you're just creating the compound type rather than array type, the argument should probably just be the type itself. You may then want to manually dispatch based on whether `isstruct` is `true`.

tamasgal: Make sense... However, I hit the wall somewhere because `Type Array does not have a definite size.`, which I do not understand yet ;)

mkitti: You need something from StaticArrays.jl for the size to be in the type.

Otherwise, you can just take an instance and then dispatch on `isstruct`.

mkitti:
@tamasgal, in case you don't follow what I'm talking about in the comments, I mean the following:

```julia
julia> import HDF5: API, Datatype

julia> HDF5.datatype(x::T) where T = HDF5.datatype(x, Val(isstructtype(T)))

julia> function HDF5.datatype(x::T, isstruct::Val{true}) where T
           dtype = API.h5t_create(API.H5T_COMPOUND, sizeof(T))
           for (idx, fn) in enumerate(fieldnames(T))
               API.h5t_insert(dtype, fn, fieldoffset(T, idx), datatype(fieldtype(T, idx)))
           end
           Datatype(dtype)
       end

julia> datatype(MyStruct(3,5))
HDF5.Datatype: H5T_COMPOUND {
      H5T_STD_I64LE "x" : 0;
      H5T_STD_I64LE "y" : 8;
   }

julia> datatype((x=5, y=3.f0))
HDF5.Datatype: H5T_COMPOUND {
      H5T_STD_I64LE "x" : 0;
      H5T_IEEE_F32LE "y" : 8;
   }

```

tamasgal:
Ah yes, I was struggling with that. Did not came up with the `Val`-solution, thanks, I will try ;)

mkitti:
People may want to override this method still for their specific types. Your implementation relies on Julia's memory aligned version, which may take up more space than needed.

```julia
julia> struct MyStruct32
           x::Int32
           y::Int32
       end

julia> datatype(MyStruct32(3,5))
HDF5.Datatype: H5T_COMPOUND {
      H5T_STD_I32LE "x" : 0;
      H5T_STD_I32LE "y" : 4;
   }

julia> struct MyStructMixed
           x::Int32
           y::Int
       end

julia> datatype(MyStructMixed(3,5))
HDF5.Datatype: H5T_COMPOUND {
      H5T_STD_I32LE "x" : 0;
      H5T_STD_I64LE "y" : 8;
   }

julia> sizeof(MyStruct32)
8

julia> sizeof(MyStructMixed)
16
```

As for the `NamedTuple` thing, perhaps people should write conversions methods to their types from a `NamedTuple`? It might be as simple as using `Base.@kwdef`:

```julia
julia> Base.@kwdef struct MyStructKW
           x::Int32
           y::Int
       end
MyStructKW

julia> MyStructKW(;(y = 1, x = 2)...)
MyStructKW(2, 1)
```

tamasgal:
You are right about the memory padding, I guess I will just use the actual sizes as offset then.

Not sure if I understand regarding `NamedTuple`, at least in the tests they work fine. Or did I miss anything? `;)`

mkitti:
I'm talking about your comment:

> It's not reversible, meaning that read will yield the usual NamedTuple

If they wrote a `convert` method they could just use a type assertion to get what they want.

mkitti:
> You are right about the memory padding, I guess I will just use the actual sizes as offset then.

We would have to test the conversion closely I think. This may make memory mapping difficult for example.

tamasgal:
Ah yes, understood `:)`. I guess that should go into the docs then.

Btw. regarding the `Type Array does not have a definite size.`: I think it would make sense to simply use the length of the passed `AbstractArray` at the time of writing the dataset. I mean, using `StaticArrays` is of course a viable option but it might be user-friendly to support also the "obvious" use-case. Or do you think we should force fixed size sets?

tamasgal:
I am still struggling to find out where I hit the `Type Array does not have a definite size.` error since when I define `datatype()` for `AbstractArrays`, it does not complain and uses the actual size of the array (read: everything works as expected). If I only define the struct (the very same function body for the method, just a different dispatch type), the error bubbles up from somewhere.

mkitti:
What is the stack trace?

tamasgal:
So with this it works (but of course not for writing single structs):

```julia
datatype(x::AbstractArray{T}) where {T} = HDF5.datatype(x, Val(isstructtype(T)))
function datatype(x::AbstractArray{T}, isstruct::Val{true}) where {T}
    dtype = API.h5t_create(API.H5T_COMPOUND, sizeof(T))
    for (idx, fn) in enumerate(fieldnames(T))
        API.h5t_insert(dtype, fn, fieldoffset(T, idx), datatype(fieldtype(T, idx)))
    end
    Datatype(dtype)
end
```

and if I change the signatures to:

```julia
datatype(x::T) where {T} = HDF5.datatype(x, Val(isstructtype(T)))
function datatype(x::T, isstruct::Val{true}) where {T}
```

and call with the same tests, like

```julia
    bars = [
        [Bar(1, 1.1, true) Bar(2, 2.1, false) Bar(3, 3.1, true)]
        [Bar(4, 4.1, false) Bar(5, 5.1, true) Bar(6, 6.1, false)]
    ]

    namedtuples = [(a=1, b=2.3), (a=4, b=5.6)]

    fn = tempname()
    h5open(fn, "w") do h5f
       write_dataset(h5f, "the/bars", bars)
       write_dataset(h5f, "the/namedtuples", namedtuples)
    end
```

I get `Type Array does not have a definite size.` and I don't even find that line in the source code. It's most likely in `HDF5_jll`. I checked the `dataspace()` methods but I have not stepped through with the debugger.

```
  Got exception outside of a @test
  Type Array does not have a definite size.
  Stacktrace:
    [1] sizeof(x::Type)
      @ Base ./essentials.jl:473
    [2] datatype(x::Matrix{Bar}, isstruct::Val{true})
      @ HDF5 ~/Dev/HDF5.jl/src/typeconversions.jl:66
    [3] datatype(x::Matrix{Bar})
      @ HDF5 ~/Dev/HDF5.jl/src/typeconversions.jl:64
    [4] create_dataset(parent::HDF5.File, name::String, data::Matrix{Bar}; pv::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ HDF5 ~/Dev/HDF5.jl/src/datasets.jl:268
    [5] create_dataset(parent::HDF5.File, name::String, data::Matrix{Bar})
      @ HDF5 ~/Dev/HDF5.jl/src/datasets.jl:265
    [6] write_dataset(parent::HDF5.File, name::String, data::Matrix{Bar}; pv::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ HDF5 ~/Dev/HDF5.jl/src/datasets.jl:282
    [7] write_dataset(parent::HDF5.File, name::String, data::Matrix{Bar})
      @ HDF5 ~/Dev/HDF5.jl/src/datasets.jl:279
    [8] (::var"#254#257"{Vector{NamedTuple{(:a, :b), Tuple{Int64, Float64}}}, Matrix{Bar}})(h5f::HDF5.File)
      @ Main ~/Dev/HDF5.jl/test/compound.jl:167
    [9] (::HDF5.var"#17#18"{HDF5.HDF5Context, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, var"#254#257"{Vector{NamedTuple{(:a, :b), Tuple{Int64, Float64}}}, Matrix{Bar}}, HDF5.File})()
      @ HDF5 ~/Dev/HDF5.jl/src/file.jl:96
   [10] task_local_storage(body::HDF5.var"#17#18"{HDF5.HDF5Context, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, var"#254#257"{Vector{NamedTuple{(:a, :b), Tuple{Int64, Float64}}}, Matrix{Bar}}, HDF5.File}, key::Symbol, val::HDF5.HDF5Context)
      @ Base ./task.jl:292
   [11] h5open(::var"#254#257"{Vector{NamedTuple{(:a, :b), Tuple{Int64, Float64}}}, Matrix{Bar}}, ::String, ::Vararg{String}; context::HDF5.HDF5Context, pv::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ HDF5 ~/Dev/HDF5.jl/src/file.jl:91
   [12] h5open(::Function, ::String, ::String)
      @ HDF5 ~/Dev/HDF5.jl/src/file.jl:89
   [13] macro expansion
      @ ~/Dev/HDF5.jl/test/compound.jl:165 [inlined]
   [14] macro expansion
      @ ~/.julia/juliaup/julia-1.8.1+0.aarch64/share/julia/stdlib/v1.8/Test/src/Test.jl:1357 [inlined]
   [15] top-level scope
      @ ~/Dev/HDF5.jl/test/compound.jl:157
   [16] include(fname::String)
      @ Base.MainInclude ./client.jl:476
   [17] macro expansion
      @ ~/Dev/HDF5.jl/test/runtests.jl:24 [inlined]
   [18] macro expansion
      @ ~/.julia/juliaup/julia-1.8.1+0.aarch64/share/julia/stdlib/v1.8/Test/src/Test.jl:1357 [inlined]
   [19] top-level scope
      @ ~/Dev/HDF5.jl/test/runtests.jl:19
   [20] include(fname::String)
      @ Base.MainInclude ./client.jl:476
   [21] top-level scope
      @ none:6
   [22] eval
      @ ./boot.jl:368 [inlined]
   [23] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:276
   [24] _start()
      @ Base ./client.jl:52
```

tamasgal:
I also checked with `StaticArrays` but then it hangs and uses 100% CPU (I waited for 5 minutes, it's stuck somewhere):

```julia
    bars = @SMatrix [
        [Bar(1, 1.1, true) Bar(2, 2.1, false) Bar(3, 3.1, true)]
        [Bar(4, 4.1, false) Bar(5, 5.1, true) Bar(6, 6.1, false)]
    ]

    namedtuples = @SVector [(a=1, b=2.3), (a=4, b=5.6)]
```

mkitti:
I'm a bit confused about what you're trying to do. You want to create a dataset where each element of the dataset is an array of structs? That is the dataset is an array of arrays of structs?

src/typeconversions.jl
```diff
@@ -60,6 +60,16 @@ function datatype(str::VLen{C}) where {C<:CharType}
...
 
+# Compound types
+datatype(x::AbstractArray{T}) where {T} = HDF5.datatype(x, Val(isstructtype(T)))
+function datatype(x::AbstractArray{T}, isstruct::Val{true}) where {T}
+    dtype = API.h5t_create(API.H5T_COMPOUND, sizeof(T))
+    for (idx, fn) in enumerate(fieldnames(T))
+        API.h5t_insert(dtype, fn, fieldoffset(T, idx), datatype(fieldtype(T, idx)))
+    end
+    Datatype(dtype)
+end
```
mkitti, line 71:
```suggestion

datatype(::T) where T = Datatype(hdf5_type_id(T), false)
datatype(::Type{T}) where T = Datatype(hdf5_type_id(T), false)
datatype(x::AbstractArray{T}) where T = Datatype(hdf5_type_id(T), false)


# Move closer to where hdf5_type_id is defined
hdf5_type_id(::Type{T}) where T = hdf5_type_id(T, Val(isstructtype(T)))
function hdf5_type_id(::Type{T}, isstruct::Val{true}) where T
    dtype = API.h5t_create(API.H5T_COMPOUND, sizeof(T))
    for (idx, fn) in enumerate(fieldnames(T))
        API.h5t_insert(dtype, fn, fieldoffset(T, idx), hdf5_type_id(fieldtype(T, idx)))
    end
    return dtype
end
# 
hdf5_type_id(::Type{T}, isstruct::Val{false}) where T = MethodError(hdf5_type_id, (T, isstruct))
```

How about this. Some of this logic belongs at the `hdf5_type_id` level whereas `datatype` is more of a frontend for wrapping the hdf5 type id as `Datatype`

mkitti:
This looks like it works:

```julia
julia> import HDF5: datatype, hdf5_type_id, API, Datatype

julia> datatype(::T) where T = Datatype(hdf5_type_id(T), false)
datatype (generic function with 16 methods)

julia> datatype(::Type{T}) where T = Datatype(hdf5_type_id(T), false)
datatype (generic function with 16 methods)

julia> datatype(x::AbstractArray{T}) where T = Datatype(hdf5_type_id(T), false)
datatype (generic function with 16 methods)

julia> # Move closer to where hdf5_type_id is defined
       hdf5_type_id(::Type{T}) where T = hdf5_type_id(T, Val(isstructtype(T)))
hdf5_type_id (generic function with 16 methods)

julia> function hdf5_type_id(::Type{T}, isstruct::Val{true}) where T
           dtype = API.h5t_create(API.H5T_COMPOUND, sizeof(T))
           for (idx, fn) in enumerate(fieldnames(T))
               API.h5t_insert(dtype, fn, fieldoffset(T, idx), hdf5_type_id(fieldtype(T, idx)))
           end
           return dtype
       end
hdf5_type_id (generic function with 16 methods)

julia> # Perhaps we need a custom error type here
       hdf5_type_id(::Type{T}, isstruct::Val{false}) where T = throw(MethodError(hdf5_type_id, (T, isstruct)))
hdf5_type_id (generic function with 16 methods)

julia>     bars = @SMatrix [
               [Bar(1, 1.1, true) Bar(2, 2.1, false) Bar(3, 3.1, true)]
               [Bar(4, 4.1, false) Bar(5, 5.1, true) Bar(6, 6.1, false)]
           ]

           namedtuples = @SVector [(a=1, b=2.3), (a=4, b=5.6)]

julia> datatype(bars)
HDF5.Datatype: H5T_COMPOUND {
      H5T_STD_I32LE "a" : 0;
      H5T_IEEE_F64LE "b" : 8;
      undefined integer "c" : 16;
   }

julia> datatype(namedtuples)
HDF5.Datatype: H5T_COMPOUND {
      H5T_STD_I64LE "a" : 0;
      H5T_IEEE_F64LE "b" : 8;
   }
```

src/typeconversions.jl
tamasgal: OK now I start to grasp. This `hdf5_type_id` was the key. I tried something similar but always ended up in infinite recursion...

src/HDF5.jl
```diff
@@ -28,6 +28,7 @@ export @read,
     open_dataset,
     read_dataset,
     write_dataset,
+    write_compound_dataset,
```
mkitti, line 31:
```suggestion
```

Does this function still exist?

mkitti (changes requested):
We can deal with the memory compaction issue later. Let's cleanup the code and merge it.

test/compound.jl
```diff
@@ -146,3 +146,54 @@ end
...
+    fn = tempname()
+    h5open(fn, "w") do h5f
+        # write_dataset(h5f, "the/bar", Bar(1, 2.3, false))
+        write_dataset(h5f, "the/bars", bars)
+        write_dataset(h5f, "the/namedtuples", namedtuples)
+    end
+
+    # thebar = h5open(fn, "r") do h5f
+    #     read(h5f, "the/bar")
+    # end
```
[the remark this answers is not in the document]
tamasgal: That's a left-over ;)

mkitti (approved):
We may need a bit of reorganization after the merge. It also may be necessary to evaluate if some type signatures are now redundant.

mkitti:
@musm, would you like to take a look?

mkitti:
I will merge in approximately 4 days unless I hear otherwise.

musm:
Opps sorry, I completely forgot about this. I'll look at this tomorrow!

mkitti:
Heh, ok. Me too.

nhz2:
Writing a dataset with an array of mutable structs will fail silently, and just write garbage data.

For example:
```julia
using HDF5

mutable struct Bar
    a::Int32
    b::Float64
    c::Bool
end

bars = fill(Bar(1, 1.1, true), 10)

fn = tempname()
h5open(fn, "w") do h5f
    write_dataset(h5f, "the/bars", bars)
end

thebars = h5open(fn, "r") do h5f
    read(h5f, "the/bars")
end

thebars
```
Gives:
```julia
10-element Vector{NamedTuple{(:a, :b, :c), Tuple{Int32, Float64, Bool}}}:
 (a = 1884163280, b = 6.9470829104969e-310, c = 0)
 (a = 1884163280, b = 6.9470829104969e-310, c = 0)
 (a = 1884163280, b = 6.9470829104969e-310, c = 0)
 (a = 1884163280, b = 6.9471988325945e-310, c = 0)
 ⋮
 (a = 167768072, b = 6.9472103088853e-310, c = 1)
 (a = -1, b = 0.0, c = 0)
 (a = 66742336, b = 3.1828515e-316, c = 0)
```

tamasgal:
Thanks, good catch! I added test and will fix asap.

src/typeconversions.jl
```diff
@@ -60,6 +60,25 @@ function datatype(str::VLen{C}) where {C<:CharType}
     Datatype(API.h5t_vlen_create(type_id))
 end
 
+# Compound types
+
+datatype(::T) where {T} = Datatype(hdf5_type_id(T), false)
```
simonbyrne, line 65: why `toclose = false`?

tamasgal: To be honest, I am not sure why this is needed, I took it over from @mkitti

mkitti: I'm just following similar methods here:
https://github.com/JuliaIO/HDF5.jl/blob/987d9c89e43abe68dfeced963314ed09ad4c1803/src/typeconversions.jl#L197-L199

mkitti:
We may need to dispatch on `isbitstype`.

nhz2:
Yeah, I think it makes sense to error somewhere if the struct type isn't `isbitstype`. Also if you update to Compat 3.40 you can use `ismutabletype`, but I'm not sure if that would work, because an immutable struct may have mutable struct type fields.

src/typeconversions.jl
mkitti: We're also defining new `hdf5_type_id` below.

Using the finalizer can get messy due to multithreading issues. What we should consider, perhaps in another pull request, is a `do` method where we automatically close the `Datatype` in the current thread.

simonbyrne: > We're also defining new `hdf5_type_id` below.

What I meant is that e,g, `hdf5_type_id(::Type{Int64})` is just accessing a constant, it doesn't create a new handle. The new method in this PR (https://github.com/JuliaIO/HDF5.jl/pull/1013/files#diff-bf557f1b887a640c4fe55cee7279c9f621001ff726a876b9f7bfbd7f8a7d0b44R70) calls `h5t_create`, creating a new handle. They need to be cleaned up, or you'll have a memory leak.

I'm not sure I understand the threading issue: we have finalizers on lots of other objects, why would Datatypes be special in this case?

mkitti: The finalizers are problematic elsewhere as well. But we also have a `do` interface in many of those cases which helps.

Quoting Jeff Benzanson in https://github.com/JuliaLang/julia/issues/11207 :

> Finalizers are inefficient and unpredictable. And with the new GC, it might take much longer to get around to freeing an object, therefore tying up its resources longer. Ideally releasing external resources should not be tied to how memory management works.

The problem is particularly acute here since the underlying HDF5 C library is not thread-safe. https://github.com/JuliaLang/julia/pull/45272 improves the situation by making the finalizer calls eager and deterministic, making it likely that the finalizer calls `H5T_close` using the same thread.

The better solution here is to either explicitly call `close` which should invoke `HDF5.API.h5t_close` or we provide a `do` syntax to assist with this.

simonbyrne: I don't disagree, and adding some sort of `do` syntax pattern is a good idea, but that doesn't preclude having finalizers as well for when that approach isn't feasible.

Within the scope of this PR, these objects should be finalized, or otherwise you're going to get memory leaks.

mkitti:
I have solution for mutable types. Basically, we need to convert the mutable type to a `NamedTuple`. Fortunately, we already have `HDF5.get_mem_compatible_jl_type` which already describes that `NamedTuple`. I will push soon.

tamasgal:
Thanks for jumping in @mkitti ! `:)`

How should we proceed with the finalisation problem?

mkitti:
> Thanks for jumping in @mkitti ! `:)`
> 
> How should we proceed with the finalisation problem?

For now, I've changed the second argument to `true` in [8eacf5f](https://github.com/JuliaIO/HDF5.jl/pull/1013/commits/8eacf5f31505c6a84965bd17f4d531857ee7b400) with some reservation.

Perhaps we should probably `close` any compound datatypes that we create eagerly. Perhaps we should even close any compound datatypes passed into these methods. The larger issue of threading and overall use of finalization should be addressed, but perhaps elsewhere.

simonbyrne:
One thing we recently added in MPI.jl (which has a similar construction for custom datatypes) is an internal cache, which may be helpful when reading/writing multiple times:
https://github.com/JuliaParallel/MPI.jl/blob/eaa2599f4f79611852df21d7c86cd81da8b9f156/src/datatypes.jl#L132-L153

tamasgal:
I am a bit confused which actions need to be taken in order to merge this. Can I do anything else to make progress?

mkitti:
I'm leaning towards the merge side, but I think @musm needs to review.

I think we should consider @simonbyrne suggestion of a cache, but that could be another pull request.

src/typeconversions.jl
```diff
@@ -60,6 +60,26 @@ function datatype(str::VLen{C}) where {C<:CharType}
     Datatype(API.h5t_vlen_create(type_id))
 end
 
+# Compound types
+
+# These will use finalizers. Close them eagerly to avoid issues.
+datatype(::T) where {T} = Datatype(hdf5_type_id(T), true)
+datatype(::Type{T}) where {T} = Datatype(hdf5_type_id(T), true)
```
musm, line 67: Isn't this definition redundant with the method right above it?

src/dataspaces.jl
```diff
@@ -46,6 +46,12 @@ The default `Dataspace` used for representing a Julia object `data`:
  - arrays: a simple `Dataspace`
  - `nothing` or an `EmptyArray`: a null dataspace
 """
```
musm, line 48: Should we update the docs here to include the fact that the new default is   a scalar `Dataspace` for struct types?

src/typeconversions.jl
mkitti: No. The first works on instances of a type. The second works on the type itself.

musm: We don't seem to need this definition (on L67)

musm:
LGTM sans minor comment 

Agree cache sounds interesting but probably for seperate PR

musm:
Sounds good!

src/dataspaces.jl
```diff
@@ -46,6 +46,12 @@ The default `Dataspace` used for representing a Julia object `data`:
  - arrays: a simple `Dataspace`
  - `nothing` or an `EmptyArray`: a null dataspace
```
musm, line 47:
```suggestion
 - `struct` types: a scalar `Dataspace`
 - `nothing` or an `EmptyArray`: a null dataspace
```

alhirzel:
Great work! I am wondering if it is possible to do a release soon so that this can be consumed by general users without `]dev`ing the package?

musm:
done

PR #1012, opened 2022-09-15
Add virtual dataset support

simonbyrne:
Requires #1011 first.

PR #1018, opened 2022-10-18
enable tests for virtual datasets

simonbyrne:
From https://github.com/JuliaIO/HDF5.jl/pull/1012#discussion_r997583755, I had forgot to enable the tests.

It turns out the docs at https://portal.hdfgroup.org/display/HDF5/H5P_SET_VIRTUAL are wrong, the `%b` specifier does not take a dimension argument (at least none of the examples do, and it breaks if you attempt to add one).


mkitti:
I think I'm slightly ahead of you in #1019

PR #1022, opened 2022-11-01
API Lock Preferences

mkitti:
This builds on the `sb/lock` branch by using Preferences.jl to configure the lock usage at compile time.

The preference, `use_api_lock` can be set to either `true` or `false`.

The preference is set via the `API.set_use_lock_pref!` function and can be quered by `API.get_use_lock_pref()`.

This mainly meant to help assess the `sb/lock` branch.

Other features such as the RawAPI module have been withdrawn.


simonbyrne:
This adds a bunch of complexity, and an additional runtime lookup at each call. Personally i don't think it's worth it (especially since I haven't been able to see any noticeable overhead with the locking).

I think if you want to make it optional, just make it an all-or-nothing: `use_api_lock` is `true`/`false` with
```julia
const use_api_lock = @load_preference("use_api_lock", true)
```
and make the API function wrappers:
```julia
    use_api_lock && lock(liblock)
    var"#status#" = try
            # ccall
        finally
            use_api_lock && unlock(liblock)
        end
    var"#status#" < 0 && @h5error ...
    return nothing
```

mkitti:
After testing, I am getting this error. I suspect this is because the finalizer is trying to run after we call `h5_close()`. https://github.com/JuliaIO/HDF5.jl/blob/af62ebf5b87a347edab6f903c2f7dced3ed03e25/test/runtests.jl#L77

```
error in running finalizer: HDF5.API.H5Error(msg="Error closing file", id=1008806316530991104)
macro expansion at /home/mkitti/.julia/dev/HDF5/src/api/error.jl:18 [inlined]
h5f_close at /home/mkitti/.julia/dev/HDF5/src/api/functions.jl:1054
close at /home/mkitti/.julia/dev/HDF5/src/file.jl:120 [inlined]
lock_and_close at /home/mkitti/.julia/dev/HDF5/src/api/lock.jl:58
unknown function (ip: 0x7f8eb5b0b122)
_jl_invoke at /cache/build/default-amdci5-0/julialang/julia-release-1-dot-8/src/gf.c:2367 [inlined]
ijl_apply_generic at /cache/build/default-amdci5-0/julialang/julia-release-1-dot-8/src/gf.c:2549
jl_apply at /cache/build/default-amdci5-0/julialang/julia-release-1-dot-8/src/julia.h:1838 [inlined]
run_finalizer at /cache/build/default-amdci5-0/julialang/julia-release-1-dot-8/src/gc.c:280
jl_gc_run_finalizers_in_list at /cache/build/default-amdci5-0/julialang/julia-release-1-dot-8/src/gc.c:367
run_finalizers at /cache/build/default-amdci5-0/julialang/julia-release-1-dot-8/src/gc.c:410
ijl_atexit_hook at /cache/build/default-amdci5-0/julialang/julia-release-1-dot-8/src/init.c:236
jl_repl_entrypoint at /cache/build/default-amdci5-0/julialang/julia-release-1-dot-8/src/jlapi.c:720
main at /cache/build/default-amdci5-0/julialang/julia-release-1-dot-8/cli/loader_exe.c:59
unknown function (ip: 0x7f8f0cec8d8f)
__libc_start_main at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
unknown function (ip: 0x401098)
```

simonbyrne:
Is there a reason why the tests call `HDF5.API.h5_close()`?

test/virtual_dataset.jl
```diff
@@ -5,36 +5,36 @@ using Test, HDF5
 
     filename = joinpath(dirname, "main.hdf5")
 
-    f = h5open(filename, "w")
```
mkitti, line 8: The issue was here. We forgot to close this handle explicitly.

mkitti:
@simonbyrne I simplified this along the lines of your https://github.com/JuliaIO/HDF5.jl/pull/1022#issuecomment-1306547077 . I think this will allow us to more easily diagnose any potential issues or regressions that might emerge from this.

I also removed the RawAPI module. Something I would like to pay particular attention to is the use of the `iterate` functions where the C library is calling back into Julia code. Do we have sufficient guarantees on what thread the calls back will be run on?

src/api/lock.jl
```diff
@@ -0,0 +1,81 @@
...
+
+@static if VERSION >= v"1.6"
+    """
+        HDF5.API.use_api_lock::Bool
+
+    Julia compile time preference of whether to use a `ReentrantLock`. By default
+    the HDF5 C library is not thread-safe. Concurrent calls to the HDF5 library
+    from multiple threads will likely fail.
+    """
+    const use_api_lock = Preferences.@load_preference("use_api_lock", default = true)
```
simonbyrne, line 11:
```suggestion
    const use_api_lock::Bool = Preferences.@load_preference("use_api_lock", default = true)
```
then the error below is not necessary

src/api/lock.jl
```diff
@@ -0,0 +1,81 @@
...
+    # For diagnostics
+    # We do not need to lock twice when use_api_lock is on
+    const use_lock_and_close =
+        Preferences.@load_preference("use_lock_and_close", default = true) && !use_api_lock
```
simonbyrne, line 15: `use_lock_on_close`?

src/api/lock.jl
mkitti: My thinking here is that the error I wrote would be more descriptive than the following since this potentially faces an end user. I suppose I could try catch the MethodError though.
```julia
ERROR: MethodError: Cannot `convert` an object of type String to an object of type Bool
Closest candidates are:
  convert(::Type{T}, ::Ptr) where T<:Integer at pointer.jl:23
  convert(::Type{T}, ::T) where T<:Number at number.jl:6
  convert(::Type{T}, ::Number) where T<:Number at number.jl:7
```

simonbyrne: I guess? Seems like a bit of a weird error case to be worried about though.

mkitti:
> Is there a reason why the tests call `HDF5.API.h5_close()`?

I don't know, but it sure does help catch unclosed handles.

src/api/lock.jl
mkitti: Actually the type assertion a global will not work with Julia 1.6, and I think we're still trying to maintain compat with Julia 1.3

test/virtual_dataset.jl
musm: Probably best to pull this into a different PR

mkitti: I can pull this into a different pull request, but this needs to be here as well since CI will error otherwise.

mkitti:
Closing and resubmitting PR directly to master.

PR #1023, opened 2022-11-17
Fix virtual dataset tests, from mkitti/api_lock_preferences

mkitti:
See https://github.com/JuliaIO/HDF5.jl/pull/1022#discussion_r1018402306 .

cc: @musm


mkitti:
I'm going to merge this since it was discussed previously.

simonbyrne:
thanks!

PR #1024, opened 2022-11-18
API Lock Preferences

mkitti:
See #1022 for prior discussion. This pull request is directly to master.


mkitti:
Just to reiterate https://github.com/JuliaIO/HDF5.jl/pull/1021#issuecomment-1309998089, you can get a 15% speed boost by disabling the locks. This is especially applicable when one is only using one Julia thread and the locks serve no purpose at all.

PR #1026, opened 2022-11-22
Remove use of `tempname()` from `setindex!` of an `AttributeDict`

nhz2:
Currently `setindex!` uses `tempname()` to get a random unique name for a temporary attribute. However, `tempname()` is quite slow and requires reading the file system to check if the names it generates are unique. In this PR I am replacing `tempname()` with `"-><7;*<_NY" * bytes2hex(rand(Random.RandomDevice(), UInt8, 16))`

```julia
julia> randname1() = "-><7;*<_NY" * bytes2hex(rand(Random.RandomDevice(), UInt8, 16))
randname1 (generic function with 1 method)

julia> randname2() = tempname()
randname2 (generic function with 1 method)

julia> @btime randname1()
  256.557 ns (4 allocations: 264 bytes)
"-><7;*<_NY2390f39a585aec3b0caac0f127a0a83a"

julia> @btime randname2()
  2.085 μs (9 allocations: 1.22 KiB)
"/tmp/jl_yjyjMTO1uh"

```

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -250,17 +250,8 @@ end
 function Base.setindex!(attrdict::AttributeDict, val, name::AbstractString)
     if haskey(attrdict, name)
         # in case of an error, we write first to a temporary, then rename
-        max_tries = 100 # number of times to try making a new name
-        _name = nothing
-        for i in 1:max_tries
-            _name = "-><7;*<_NY"*bytes2hex(rand(UInt8, 16))
-            if haskey(attrdict, _name)
-                _name = nothing
-            else
-                break
-            end
-        end
-        isnothing(_name) && error("tempname: max_tries exhausted")
+        _name = "-><7;*<_NY"*bytes2hex(rand(UInt8, 16))
+        haskey(attrdict, _name) && error("`attrdict` has the temp name somehow")
         try
             write_attribute(attrdict.parent, _name, val)
             delete_attribute(attrdict.parent, name)
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -250,7 +250,7 @@ end
 function Base.setindex!(attrdict::AttributeDict, val, name::AbstractString)
     if haskey(attrdict, name)
         # in case of an error, we write first to a temporary, then rename
-        _name = "-><7;*<_NY" * bytes2hex(rand(UInt8, 16))
+        _name = "-><7;*<_NY" * bytes2hex(rand(Random.RandomDevice(), UInt8, 16))
         haskey(attrdict, _name) && error("`attrdict` has the temp name somehow")
         try
             write_attribute(attrdict.parent, _name, val)
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -160,8 +160,8 @@ rename_attribute(
 
 Delete the [`Attribute`](@ref) named `name` on the object `parent`.
 """
-delete_attribute(parent::Union{File,Object}, path::AbstractString) =
-    API.h5a_delete(checkvalid(parent), path)
+delete_attribute(parent::Union{File,Object}, name::AbstractString) =
+    API.h5a_delete(checkvalid(parent), name)
 
 """
     h5writeattr(filename, name::AbstractString, data::Dict)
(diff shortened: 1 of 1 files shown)


src/attributes.jl
```diff
@@ -250,7 +250,8 @@ end
 function Base.setindex!(attrdict::AttributeDict, val, name::AbstractString)
     if haskey(attrdict, name)
         # in case of an error, we write first to a temporary, then rename
-        _name = tempname()
+        _name = "-><7;*<_NY" * bytes2hex(rand(Random.RandomDevice(), UInt8, 16))
+        haskey(attrdict, _name) && error("`attrdict` has the temp name somehow")
```
musm, line 254: Instead of failing this should try a new random name until it succeeds

src/attributes.jl
```diff
@@ -250,7 +250,8 @@ end
 function Base.setindex!(attrdict::AttributeDict, val, name::AbstractString)
     if haskey(attrdict, name)
         # in case of an error, we write first to a temporary, then rename
-        _name = tempname()
+        _name = "-><7;*<_NY" * bytes2hex(rand(Random.RandomDevice(), UInt8, 16))
```
musm, line 253: What's the rationale of the initial string? In other words, why not rely on the bytes2hex(' random number string).

mkitti:
Do we actually need a random name?

mkitti:
Could we use some predictable sequence of names or perhaps some variant of the old the name?

nhz2:
In terms of needing the temperary name, from what I can tell the HDF5 C library doesn't have a way to change the datatype, dataspace, and value of an attribute at the same time. h5py is doing something similar https://github.com/h5py/h5py/blob/fcaca1d1b81d25c0d83b11d5bdf497469b5980e9/h5py/_hl/attrs.py#L191-L219 using a uuid4. I think the `API.h5a_write` can change the value of an attribute but cannot change the type and space, so maybe there should be a check to see if the type and space need to be changed and if not just update the value?

src/attributes.jl
nhz2: I initially had that to make collisions even less likely, but trying random names multiple times is probably even better.

mkitti:
uuid4 or other uuids seem like a good idea. Why don't we do that? Is that slow?

Also, why not make the name related to the old name in case there is an error and we never complete the operation?

```julia
julia> using UUIDs

julia> name = "helloworld"
"helloworld"

julia> _name = name * "_hdf5jl_" * string(uuid4())
"helloworld_hdf5jl_452f6a32-c1db-4756-bbfa-42b2fd7d0a29"
```

mkitti:
```julia
julia> randname1() = "-><7;*<_NY" * bytes2hex(rand(Random.RandomDevice(), UInt8, 16))
randname1 (generic function with 1 method)

julia> randname2() = tempname()
randname2 (generic function with 1 method)

julia> randname3(name) = name * "_hdf5jl_" * string(uuid4())
randname3 (generic function with 1 method)


julia> @btime randname1()
  631.651 ns (4 allocations: 264 bytes)
"-><7;*<_NY3ac27142c20bbb1c3129af9503747718"

julia> @btime randname2()
  6.007 μs (9 allocations: 1.22 KiB)
"/tmp/jl_jEhEJkVQre"

julia> @btime randname3($name)
  608.423 ns (3 allocations: 192 bytes)
"helloworld_hdf5jl_7ca25e59-4c2e-406e-90e0-80332425c05e"
```

mkitti (approved):

mkitti:
@simonbyrne want to take a quick look?

src/attributes.jl
```diff
@@ -250,7 +250,17 @@ end
 function Base.setindex!(attrdict::AttributeDict, val, name::AbstractString)
     if haskey(attrdict, name)
         # in case of an error, we write first to a temporary, then rename
-        _name = tempname()
+        max_tries = 100 # number of times to try making a new name
+        _name = nothing
+        for i in 1:max_tries
```
musm, line 255: Since we've moved to uuid, I really don't expect there to be a clash using uuid4 since the probability of collision is so incredibly low. Perhaps we could just remove this and simply this.

src/attributes.jl
```diff
@@ -250,7 +250,17 @@ end
 function Base.setindex!(attrdict::AttributeDict, val, name::AbstractString)
     if haskey(attrdict, name)
         # in case of an error, we write first to a temporary, then rename
-        _name = tempname()
+        max_tries = 100 # number of times to try making a new name
+        _name = nothing
```
simonbyrne, line 254: I don't think `*(::Nothing, ::String)` is defined.

simonbyrne (approved):

PR #1027, opened 2022-12-19
Make `begin` as the first index of a `Dataset` work

machakann:
Consider that `d` is a `Dataset` with two dimensions. Currently, `d[:, end]` works, however, `d[:, begin]` doesn't. This PR intended to make the latter work.

```julia
using HDF5
fn = tempname()
fid = h5open(fn, "w")
g = create_group(fid, "foo")
d = create_dataset(g, "bar", Int, (1, 3))
d[1, 1:3] = 1:3

@show d[1, end]  # OK
@show d[1, begin] # Error
```

The error message is:
```
MethodError: no method matching axes(::HDF5.Dataset, ::Int64)
Closest candidates are:
  axes(::Base.Broadcast.Broadcasted{<:Any, <:Tuple{Vararg{T, N}} where T}, ::Integer) where N at broadcast.jl:227
  axes(::Number, ::Integer) at number.jl:83
  axes(::AbstractArray{T, N}, ::Any) where {T, N} at abstractarray.jl:72
  ...

Stacktrace:
 [1] firstindex(a::HDF5.Dataset, d::Int64)
   @ Base .\abstractarray.jl:402
 [2] top-level scope
   @ show.jl:1047
 [3] eval
   @ .\boot.jl:368 [inlined]
 [4] include_string(mapexpr::typeof(REPL.softscope), mod::Module, code::String, filename::String)
   @ Base .\loading.jl:1428
```

### DIFF
--- a/src/datasets.jl
+++ b/src/datasets.jl
@@ -310,6 +310,7 @@ Base.write(parent::Union{File,Group}, name::Union{AbstractString,Nothing}, data;
 
 Base.eachindex(::IndexLinear, A::Dataset) = Base.OneTo(length(A))
 Base.axes(dset::Dataset) = map(Base.OneTo, size(dset))
+Base.axes(dset::Dataset, d::Integer) = Base.OneTo(size(dset, d))
 
 # Write to a subset of a dataset using array slices: dataset[:,:,10] = array
 
--- a/src/datasets.jl
+++ b/src/datasets.jl
@@ -131,6 +131,7 @@ Base.flush(ds::Dataset) = API.h5d_flush(checkvalid(ds))
 # Array constructor for datasets
 Base.Array(x::Dataset) = read(x)
 
+# The next two lines are kept for v"1.4" <= VERSION <= v"1.5"
 Base.lastindex(dset::Dataset) = length(dset)
 Base.lastindex(dset::Dataset, d::Int) = size(dset, d)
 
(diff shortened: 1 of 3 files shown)


machakann:
Additionally, this is optional but the definition of `lastindex` is no longer needed, probably. `Base.lastindex` also uses `axes` to work. 

https://github.com/JuliaIO/HDF5.jl/blob/205518d6c1d09dc8586b577d88a2f3633a9ea061/src/datasets.jl#L134-L135

simonbyrne (approved):

musm:
Julia 1.0/1.3 have test failures
```
┌ Warning: Deprecated syntax `"begin" inside indexing expression` at /home/runner/work/HDF5.jl/HDF5.jl/test/extend_test.jl:26.

```

might need to version out the tests and or changes.

musm:
> Additionally, this is optional but the definition of `lastindex` is no longer needed, probably. `Base.lastindex` also uses `axes` to work.

If we don't need them, better to remove them for simplicity than cary uneccessary definitions.

machakann:
Thank you for your comments. I will work on it. I still need some tweaks for tests.

I found that the test failed without the two lines of `lastindex` in julia v1.4.2 and v1.5.4. So, the two lines should be kept for v1.4 and v1.5. Probably, https://github.com/JuliaLang/julia/pull/38742 is required to remove them. I will just add a comment on it. (`begin` just works without explicit `firstindex` definitions.)

musm:
Does the `@static` need to be pulled out of the  testset ?  Seems like it's not being respected

machakann:
Finally, all tests passed. Since `begin` is interpreted as a begin-end block in v1.3, I separate the file to avoid LoadError.

musm:
Great thanks!

PR #1032, opened 2023-01-13
correct docs on JULIA_HDF5_PATH search path

stevengj:
The docs say that it searches `$JULIA_HDF5_PATH/lib` and `$JULIA_HDF5_PATH/lib64` for the libraries, but [according to the code](https://github.com/JuliaIO/HDF5.jl/blob/259763f719b69f9e7b0f997eb1d8a539c014cc72/deps/build.jl#L43) it also searches `$JULIA_HDF5_PATH` directly.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -27,7 +27,7 @@ Starting from Julia 1.3, the HDF5 binaries are by default downloaded using the `
 
 ### Using custom or system provided HDF5 binaries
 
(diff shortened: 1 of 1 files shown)


mkitti:
I'm ok with merging this, but I think we should probably deprecate this mechanism in favor of the Preferences.jl based JLL override mechanism. One could then either override the location via Overrides.toml in the Julia depot or LocalPreferences.toml in the project directory.

Issue #1015, opened 2022-10-04
Better error messages when trying to use MPI without a compatible binary


simonbyrne:
This isn't very helpful:
<img width="659" alt="Screen Shot 2022-10-04 at 12 00 11 PM" src="https://user-images.githubusercontent.com/1692009/193930255-a525ac02-d5ec-4435-adff-dc3cd78e9a5e.png">

simonbyrne:
I guess the solution would be to get rid of the conditional here:
https://github.com/JuliaIO/HDF5.jl/blob/259763f719b69f9e7b0f997eb1d8a539c014cc72/src/drivers/drivers.jl#L92

mkitti:
Maybe we should turn it into and `if` ... `else`. If we don't have MPIO, we include another file that returns an error if someone tries to use MPI?

simonbyrne:
It should throw an error here:
https://github.com/JuliaIO/HDF5.jl/blob/0120cb2d32fc8d57727b3dcb17860c63c00a0cd3/src/drivers/mpio.jl#L43-L48

PR #1038, opened 2023-01-21
Clean up MPI

simonbyrne:
- Simplify API function definitions
- Improve error messages (#1015)


src/api/helpers.jl
```diff
@@ -949,18 +949,19 @@ end
...
-end
+See `libhdf5` documentation for [`H5Pget_fapl_mpio`](https://portal.hdfgroup.org/display/HDF5/H5P_GET_FAPL_MPIO32).
+"""
+function h5p_set_fapl_mpio end
+
+"""
+    h5p_set_fapl_mpio32(fapl_id::hid_t, comm::MPI.API.MPI_Comm, info::MPI.API.MPI_Info)
+
+See `libhdf5` documentation for [`H5Pset_fapl_mpio`](https://portal.hdfgroup.org/display/HDF5/H5P_SET_FAPL_MPIO32).
+"""
```
musm, line 966:
```suggestion
```

Something here seems off. This all should be deleted right?

musm (approved):

src/api/helpers.jl
simonbyrne: It's just a stub for the docs, but I have the names the wrong way around.

PR #1040, opened 2023-01-22
Bump H5Zblosc version to 0.1.1

mkitti:
H5Zblosc was last update in January 2022, there have been updates since then


mkitti:
@JuliaRegistrator register subdir=filters/H5Zblosc

PR #1043, opened 2023-01-25
Parse doxygen tag files

mkitti:
I parsed the Doxygen tag file and generated a tab separated value file, [hdf5_func_urls.tsv](https://github.com/JuliaIO/HDF5.jl/compare/master...mkitti:HDF5.jl:mkitti/parse_doxygen_tag_files?expand=1#diff-759a509ea6ee133395dff7cd4f9227ea12121458c24acb8fd9b4701cb02906f5)

The next step is to use these URLs when generating the documentation.


simonbyrne:
Haha, I had been working on this as well. I'll open a draft PR with what I have, maybe we can combine.

mkitti:
I'm still interested in taking some ideas from #1044  such as replacing the `for` loop with a xpath query, but I"m running short on time.

mkitti:
We still need to modify h5doc: https://github.com/JuliaIO/HDF5.jl/blob/ffc0b66cdab5be3c8e29a637c4514c97d7e126d8/src/HDF5.jl#L65

mkitti:
I changed this to use the HDF5 v1.14 documentation rather than develop to narrow the gap with #1044.

I'm going to merge this over #1044. My main reasons for preferring this over #1044 are as follows.
* Here in 1043, the documentation URLs are statically embedded statically at generate time within the documentation rather than at parse time via macro that does that a xpath lookup.
* The generation step here does a single pass over the XML tag file to generate a tsv file that is easily converted into a Julia `Dict`. The xpath lookup ends up multiple passes over the XML file.

I do want to revisit some of the ideas in #1044 later, especially with a pure Julia XML parser.

The current errors are due HDF5_jll binaries losing SZIP support. That should be fixed shortly with HDF5 1.14.2.

PR #1049, opened 2023-02-03
Implement try_close_finalizer, fix #1048

### LINKED ISSUE #1048: Finalizer error when rapidly writing to HDF5
Hi,

I have the following setup:

1. A main thread which is idling
2. A thread which writes to a HDF5 dataset faster than 1 ms delay between successive writes, i.e. (just pseudo-MWE)
```julia
Threads.@spawn begin
     for i=1:1_000_000
          lock(some_filelock) do
              hdf5_dataset[startidx:stopidx,1] = somedata
          end
     end
end
```

If the writes are too fast, the task/thread crashes with the following trace:

```
error in running finalizer: ErrorException("val already in a list")
error at .\error.jl:35
push! at .\linked_list.jl:53 [inlined]
_wait2 at .\condition.jl:87
#wait#621 at .\condition.jl:127
wait at .\condition.jl:125 [inlined]
slowlock at .\lock.jl:156
lock at .\lock.jl:147 [inlined]
h5i_is_valid at C:\Users\user\.julia\packages\HDF5\TcavY\src\api\functions.jl:1960
isvalid at C:\Users\user\.julia\packages\HDF5\TcavY\src\objects.jl:2 [inlined]
close at C:\Users\user\.julia\packages\HDF5\TcavY\src\dataspaces.jl:25
unknown function (ip: 000002027f7c9b96)
run_finalizer at C:/workdir/src\gc.c:298
jl_gc_run_finalizers_in_list at C:/workdir/src\gc.c:388
run_finalizers at C:/workdir/src\gc.c:434
jl_mutex_unlock at C:/workdir/src\julia_locks.h:81 [inlined]
ijl_task_get_next at C:/workdir/src\partr.c:385
poptask at .\task.jl:963
wait at .\task.jl:972
#wait#621 at .\condition.jl:130
wait at .\condition.jl:125 [inlined]
take_buffered at .\channels.jl:456
take! at .\channels.jl:450 [inlined]
iterate at .\channels.jl:592
macro expansion at c:\Users\user\.vscode\extensions\julialang.language-julia-1.41.1\scripts\packages\JSONRPC\src\core.jl:103 [inlined]
#1 at .\task.jl:514
unknown function (ip: 000002027f6ffb33)
jl_apply at C:/workdir/src\julia.h:1874 [inlined]
start_task at C:/workdir/src\task.c:1092
```

Versioninfo:

```
Julia Version 1.9.0-beta3
Commit 24204a7344 (2023-01-18 07:20 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 8 × Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, skylake)
  Threads: 8 on 8 virtual cores
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 8
```

mkitti:
The formatting is a bit weird.

```julia
function try_close_finalizer(x)
    if !islocked(liblock) && trylock(liblock) do
        close(x)
        true
    end
    else
        finalizer(try_close_finalizer, x)
    end
end
```

The MWE of #1048 seems to pass now. How do I turn this into a test?

```julia
julia> begin
       using HDF5

       fid = h5open(tempname() * ".h5", "w")

       d = create_dataset(fid, "data", datatype(Int), ((1_000_000,1), (-1,1)), chunk=(1,1))

       t = Threads.@spawn begin
           for i=1:1_000_000
               d[i,1] = 1
           end
       end
       wait(t)

       close(fid)
       end
```

### DIFF
--- a/src/api/lock.jl
+++ b/src/api/lock.jl
@@ -11,4 +11,5 @@ function try_close_finalizer(x)
     else
         finalizer(try_close_finalizer, x)
     end
+    return nothing
 end
--- a/src/api/lock.jl
+++ b/src/api/lock.jl
@@ -4,13 +4,11 @@ const liblock = ReentrantLock()
 # Otherwise, defer finalization
 # https://github.com/JuliaIO/HDF5.jl/issues/1048
 function try_close_finalizer(x)
-    if !islocked(liblock) &&
-        trylock(liblock) do
-            close(x)
-            true
-        end
+    if !islocked(liblock) && trylock(liblock) do
+        close(x)
+        true
+    end
     else
         finalizer(try_close_finalizer, x)
     end
 end
-
--- a/src/api/lock.jl
+++ b/src/api/lock.jl
@@ -1,0 +1,16 @@
+const liblock = ReentrantLock()
+
+# Try to acquire the lock (test-test-set) and close if successful
+# Otherwise, defer finalization
+# https://github.com/JuliaIO/HDF5.jl/issues/1048
+function try_close_finalizer(x)
+    if !islocked(liblock) &&
+        trylock(liblock) do
+            close(x)
+            true
+        end
+    else
+        finalizer(try_close_finalizer, x)
+    end
+end
+
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -62,6 +62,8 @@ export @read,
 # H5DataStore, Attribute, File, Group, Dataset, Datatype, Opaque,
 # Dataspace, Object, Properties, VLen, ChunkStorage, Reference
 
+
+
 h5doc(name) = "[`$name`](https://portal.hdfgroup.org/display/HDF5/$(name))"
 
 include("api/api.jl")
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -62,8 +62,6 @@ export @read,
 # H5DataStore, Attribute, File, Group, Dataset, Datatype, Opaque,
 # Dataspace, Object, Properties, VLen, ChunkStorage, Reference
 
-
-
 h5doc(name) = "[`$name`](https://portal.hdfgroup.org/display/HDF5/$(name))"
 
 include("api/api.jl")
--- a/src/api/api.jl
+++ b/src/api/api.jl
@@ -13,8 +13,7 @@ else
     )
 end
 
-const liblock = ReentrantLock()
-
+include("lock.jl")
 include("types.jl")
 include("error.jl")
 include("functions.jl") # core API ccall wrappers
--- a/src/properties.jl
+++ b/src/properties.jl
@@ -104,7 +104,7 @@ macro propertyclass(name, classid)
             id::API.hid_t
             function $name(id::API.hid_t)
                 obj = new(id)
-                finalizer(close, obj)
+                finalizer(API.try_close_finalizer, obj)
                 obj
             end
         end
--- a/src/types.jl
+++ b/src/types.jl
@@ -5,6 +5,7 @@
 # Supertype of HDF5.File, HDF5.Group, JldFile, JldGroup, Matlabv5File, and MatlabHDF5File.
 abstract type H5DataStore end
 
+
 # Read a list of variables, read(parent, "A", "B", "x", ...)
 function Base.read(parent::H5DataStore, name::AbstractString...)
     tuple((read(parent, x) for x in name)...)
@@ -41,7 +42,7 @@ mutable struct File <: H5DataStore
     function File(id, filename, toclose::Bool=true)
         f = new(id, filename)
         if toclose
-            finalizer(close, f)
+            finalizer(API.try_close_finalizer, f)
         end
         f
     end
@@ -55,7 +56,7 @@ mutable struct Group <: H5DataStore
 
     function Group(id, file)
         g = new(id, file)
-        finalizer(close, g)
+        finalizer(API.try_close_finalizer, g)
         g
     end
 end
@@ -69,7 +70,7 @@ mutable struct Dataset
 
     function Dataset(id, file, xfer=DatasetTransferProperties())
         dset = new(id, file, xfer)
-        finalizer(close, dset)
+        finalizer(API.try_close_finalizer, dset)
         dset
     end
 end
@@ -84,14 +85,14 @@ mutable struct Datatype
     function Datatype(id, toclose::Bool=true)
         nt = new(id, toclose)
         if toclose
-            finalizer(close, nt)
+            finalizer(API.try_close_finalizer, nt)
         end
         nt
     end
     function Datatype(id, file::File, toclose::Bool=true)
         nt = new(id, toclose, file)
         if toclose
-            finalizer(close, nt)
+            finalizer(API.try_close_finalizer, nt)
         end
         nt
     end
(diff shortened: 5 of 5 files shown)


mkitti:
@tachawkes does this branch resolve the issue?

```Julia
using Pkg
pkg"activate --temp"
pkg"add https://github.com/mkitti/HDF5.jl#mkitti/try_close_finalizer"
begin
       using HDF5

       fid = h5open(tempname() * ".h5", "w")

       d = create_dataset(fid, "data", datatype(Int), ((1_000_000,1), (-1,1)), chunk=(1,1))

       t = Threads.@spawn begin
           for i=1:1_000_000
               d[i,1] = 1
           end
       end
       wait(t)

       close(fid)
end
```

TacHawkes:
Yes! Thank you so much for locking into this so quickly.

With this branch my MWE does not crash anymore.

mkitti:
@TacHawkes does this fix the primary code where you initially observed the issue?

By the way, why are you using multithreading here? The locks we have installed basically limit access to the HDF5 library to a single thread at a time. The underlying HDF5 C library is not thread safe.

TacHawkes:
Actually, it does!

I'm using this in a larger Julia package where analog data is read from a fast ADC, processed by Julia and written in chunks to the disk using HDF5. The rest of the software is using multi-threading and I have written a small convenience layer around HDF5.jl for abstracting all details of datasets, file handling  etc. and these threads are calling to the library from their threads. Therefore I have added another lock in my layer to guard the HDF5 file as I am aware that only single-threaded access is possible. Actually, I do not call `write` that fast in my project but there is another multi-threading issue currently being investigated and while debugging, I found this issue in HDF5.jl.

src/api/lock.jl
```diff
@@ -0,0 +1,14 @@
+const liblock = ReentrantLock()
+
+# Try to acquire the lock (test-test-set) and close if successful
+# Otherwise, defer finalization
+# https://github.com/JuliaIO/HDF5.jl/issues/1048
+function try_close_finalizer(x)
+    if !islocked(liblock) && trylock(liblock) do
+        close(x)
+        true
+    end
```
simonbyrne, line 10: How does this work? `trylock(f, liblock)` isn't documented, but I assume that inside the `do` block `liblock` will be locked. But doesn't `close` then call an API function which itself will try to get a lock?

mkitti: `trylock(f, liblock)` is just

```julia
function trylock(f, l::AbstractLock)
    if trylock(l)
        try
            return f()
        finally
            unlock(l)
        end
    end
    return false
end
```

You can call `lock` multiple times from the same `Task`, but each `lock` has to be paired with an `unlock`.

```julia
julia> const liblock = ReentrantLock()
ReentrantLock(nothing, 0x00000000, 0x00, Base.GenericCondition{Base.Threads.SpinLock}(Base.InvasiveLinkedList{Task}(nothing, nothing), Base.Threads.SpinLock(0)), (8, 1980324864016, 1980324864048))

julia> lock(liblock)

julia> lock(liblock)

julia> lock(liblock)

julia> islocked(liblock)
true

julia> unlock(liblock)

julia> unlock(liblock)

julia> islocked(liblock)
true

julia> unlock(liblock)

julia> islocked(liblock)
false
```

`trylock` also has to be paired with an `unlock` if successful, but the `do` lambda takes care of that for us.

1. We use `islocked` first because it does not involve atomics and is thus cheaper.
2. `trylock` is only attempted if `islocked` is `false`.

3a. If `trylock` can `lock`, then we enter the `do` block successfully, `close`, and return `true` from the outer `trylock`. 
4a. We then enter the empty `if` block and return `nothing`.

3b. If we cannot `lock`, then we add this to the end of the finalization queue the next time the garbage collection runs.
4b. We return whatever `finalizer` returns, which I think is `x`. Perhaps we should make this type stable and return `nothing`?

simonbyrne: > You can call lock multiple times from the same Task, but each lock has to be paired with an unlock.

Ah, TIL! That makes sense.

src/api/lock.jl
```diff
@@ -0,0 +1,14 @@
+const liblock = ReentrantLock()
+
+# Try to acquire the lock (test-test-set) and close if successful
+# Otherwise, defer finalization
+# https://github.com/JuliaIO/HDF5.jl/issues/1048
+function try_close_finalizer(x)
+    if !islocked(liblock) && trylock(liblock) do
+        close(x)
+        true
```
simonbyrne, line 9: I don't think we need this here, as it is handled by `trylock`
```suggestion
```

src/api/lock.jl
simonbyrne: > 4b. We return whatever `finalizer` returns, which I think is `x`. Perhaps we should make this type stable and return `nothing`?

That seems like a good idea

src/api/lock.jl
mkitti: I think we do need this. The two argument `trylock` above does `return f()`. The result of `true && nothing` is `nothing`, and that will be provided to the `if` statement. If the `if` statement receives a `nothing`, an error will be thrown.

```julia
julia> if true && nothing
       else
           println("Hello")
       end
ERROR: TypeError: non-boolean (Nothing) used in boolean context
Stacktrace:
 [1] top-level scope
   @ REPL[27]:1
```

src/api/lock.jl
```diff
@@ -0,0 +1,14 @@
...
+    end
+    else
+        finalizer(try_close_finalizer, x)
+    end
+end
```
mkitti, line 14:
```suggestion
    else
        finalizer(try_close_finalizer, x)
    end
    return nothing
end
```

src/api/lock.jl
simonbyrne: Oh, I didn't realize it returns `f()`.

simonbyrne (approved):

src/api/lock.jl
```diff
@@ -0,0 +1,15 @@
...
+function try_close_finalizer(x)
+    if !islocked(liblock) && trylock(liblock) do
+        close(x)
+        true
+    end
+    else
+        finalizer(try_close_finalizer, x)
+    end
+    return nothing
+end
```
simonbyrne, line 15: It's a bit weird that this calls `close`, which hasn't yet been defined?

mkitti: `close` here is `Base.close`, so technically it has been defined.

mkitti: If we moved it,  where would we put it?

After the types are defined?

simonbyrne: Create a new file `common.jl` before `properties.jl`? That said, I don't feel too strongly about this.

mkitti: I frankly do not see the difference. Something will appear out of order any way we do this. If we put this in after types.jl, then we refer to `try_close_finalizer` before we define it.

Moreover I like it where it is. When someone goes to figure out see how the lock works, they'll see this weird piece of code nearby, which may be useful for future needs.

mkitti:
I'm considering merging and releasing this within the next 12 hours since a second user has encountered this issue.

cc: @musm @simonbyrne

mkitti:
Fix resolves issues for two users. Merging. Happy to discuss organization and naming in later pull requests.

PR #1052, opened 2023-02-18
docs: mention that low-level functions are in API submodule

jwahlstrand:
In the section on low-level routines, there was previously no mention of the API submodule where the wrapped functions are defined. Also fix a typo.


musm:
Thanks

PR #1060, opened 2023-03-25
Tuple support

lxvm:
Since there is already support for reading/writing compound types, I wanted to try adding support for `Tuple`s as well, since sometimes saving tuple data is convenient, or compound types have tuple data (e.g. `SArray`). This pr writes tuples to compound types with member names corresponding to the indices of each element, i.e. `var"1", var"2", var"3",...`. I understand that there is no native HDF5 type corresponding to a tuple, however this pr is non-breaking and since tuples will be read back as named tuples, which inherit the majority of behavior from tuples, this only provides additional features. Please advise.

I also made the following changes:
- allow `dset[i] = e` for `e` a compound type by loosening method signatures
- add a method for `datatype(::Type{T})` so that `create_dataset(parent, path, T, dataspace)` works for compound types `T`

I also prepared a follow-up pr showing that we can save StaticArrays as compound types (https://github.com/lxvm/HDF5.jl/tree/pr_sarray), although one method would change in a breaking way.


src/typeconversions.jl
```diff
@@ -64,14 +64,15 @@ end
 
 # These will use finalizers. Close them eagerly to avoid issues.
 datatype(::T) where {T} = Datatype(hdf5_type_id(T), true)
+datatype(T::Type) = Datatype(hdf5_type_id(T), true)
```
mkitti, line 67: I believe we actually want this.

```suggestion
datatype(::Type{T}) where T = Datatype(hdf5_type_id(T), true)
```

src/typeconversions.jl
```diff
@@ -64,14 +64,15 @@ end
...
 datatype(x::AbstractArray{T}) where {T} = Datatype(hdf5_type_id(T), true)
 
 hdf5_type_id(::Type{T}) where {T} = hdf5_type_id(T, Val(isstructtype(T)))
 function hdf5_type_id(::Type{T}, isstruct::Val{true}) where {T}
     dtype = API.h5t_create(API.H5T_COMPOUND, sizeof(T))
     for (idx, fn) in enumerate(fieldnames(T))
         ftype = fieldtype(T, idx)
-        API.h5t_insert(dtype, fn, fieldoffset(T, idx), hdf5_type_id(ftype))
+        API.h5t_insert(dtype, Symbol(fn), fieldoffset(T, idx), hdf5_type_id(ftype))
```
mkitti, line 75: What does this change do?

mkitti:
I'm going to try to get this formatted and merged soon to do a release.

PR #1067, opened 2023-05-13
Functional API with Strings / variable-size strings

mkitti:
This is a response to https://discourse.julialang.org/t/hdf5-jl-variable-length-string/98808 adding support for just passing `String` as a type to `create_dataset`.

This allow allows `Base.setindex!` with Strings.

Traditionally, we omitted this because variable length strings are probably not a great idea to store in HDF5. Preferably, one would use fixed-length strings.

Todo:
- [x] : Add basic tests
- [x] : Test with other `AbstractString`
- [ ] : Consider implications for use with [InlineStrings.jl](https://github.com/JuliaStrings/InlineStrings.jl)
- [ ] : Consider implications for use with [StaticStrings.jl](https://github.com/mkitti/StaticStrings.jl)

### DIFF
--- a/src/datasets.jl
+++ b/src/datasets.jl
@@ -351,7 +351,9 @@ function Base.setindex!(dset::Dataset, X::Array{T}, I::IndexType...) where {T}
     return X
 end
 
-function Base.setindex!(dset::Dataset, X::Array{S}, I::IndexType...) where {S <: AbstractString}
+function Base.setindex!(
+    dset::Dataset, X::Array{S}, I::IndexType...
+) where {S<:AbstractString}
     !isconcretetype(S) && error("type $S is not concrete")
     U = get_jl_type(dset)
 
@@ -372,7 +374,6 @@ function Base.setindex!(dset::Dataset, X::Array{S}, I::IndexType...) where {S <:
         error("number of elements in src and dest arrays must be equal")
     end
 
-
     p = Ref{Cstring}(X)
     try
         API.h5d_write(dset, memtype, memspace, dspace, dset.xfer, p)
@@ -383,7 +384,6 @@ function Base.setindex!(dset::Dataset, X::Array{S}, I::IndexType...) where {S <:
     end
 
     return X
-
 end
 
 function Base.setindex!(dset::Dataset, x::T, I::IndexType...) where {T<:Number}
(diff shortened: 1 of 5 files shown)


mkitti:
Demonstration

```julia
julia> using HDF5

julia> h5open("jltest.h5", "w") do f
           ds = create_dataset(f, "strings", String, (4,))
           ds[1] = "Hello"
           ds[2] = "Hi"
           ds[3] = "Bonjour"
           ds[4] = "Hola"
       end
"Hola"

julia> h5f = h5open("jltest.h5")
🗂️ HDF5.File: (read-only) jltest.h5
└─ 🔢 strings

julia> h5f["strings"]
🔢 HDF5.Dataset: /strings (file: jltest.h5 xfer_mode: 0)

julia> h5f["strings"][1]
"Hello"

julia> h5f["strings"][2]
"Hi"

julia> h5f["strings"][3]
"Bonjour"

julia> h5f["strings"][4]
"Hola"

```

src/datasets.jl
```diff
@@ -351,12 +351,53 @@ function Base.setindex!(dset::Dataset, X::Array{T}, I::IndexType...) where {T}
     return X
 end
 
+function Base.setindex!(
+    dset::Dataset, X::Array{S}, I::IndexType...
+) where {S<:AbstractString}
```
mkitti, line 356: There is a fair amount of copying between this method and the above method.

The main difference is the need for `p = Ref{Cstring}(X)`. Can we abstract this somehow?

src/datasets.jl
```diff
@@ -351,12 +351,53 @@ function Base.setindex!(dset::Dataset, X::Array{T}, I::IndexType...) where {T}
...
+
 function Base.setindex!(dset::Dataset, x::T, I::IndexType...) where {T<:Number}
     indices = Base.to_indices(dset, I)
     X = fill(x, map(length, indices))
     Base.setindex!(dset, X, indices...)
 end
 
+function Base.setindex!(dset::Dataset, x::T, I::IndexType...) where {T<:AbstractString}
+    indices = Base.to_indices(dset, I)
+    X = fill(x, map(length, indices))
```
mkitti, line 397: Is `fill` a good thing to do for a string?

mkitti:
I'm still debating if we should fully integrate InlineStrings.jl and/or StaticStrings.jl or if we should the package extension mechanism for those. It's probably easier to move those forward as a package extension.

PR #1072, opened 2023-05-24
Handle partial chunks

0:
The extent is not guaranteed to be a multiple of chunk size. When partial chunks are present, the number of chunks should be rounded up in order to include them.

### DIFF
--- a/src/api_midlevel.jl
+++ b/src/api_midlevel.jl
@@ -57,7 +57,7 @@ function get_chunk_offset(dataset_id, index)
     extent = size(dataset_id)
     chunk = get_chunk(dataset_id)
     chunk_indices = CartesianIndices(
-        ntuple(i -> 0:(extent[i] ÷ chunk[i] - 1), length(extent))
+        ntuple(i -> 0:(cld(extent[i], chunk[i]) - 1), length(extent))
     )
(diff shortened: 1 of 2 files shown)


mkitti:
Looks promising. I'm not sure why Windows seems to be a special case here.

Also run the formatter. See contrib/format.

0:
Reformatted.

I think the failure is caused by a newer version of libhdf5. The three Windows checks that fail all report "libhdf5 v1.12.2", but the others have lower versions: v1.10.4 and v1.12.1. (Confusingly, the ones that claim to be using v1.12.1 seem to get it through HDF5_jll v1.12.2+2, but that's probably a separate issue.) I can reproduce the failure locally by using the system version of libhdf5, which is newer than v1.12.1.

I may have found the relevant [HDF5 commit](https://github.com/HDFGroup/hdf5/commit/758e97c1e5f02869dbae81a7a428d270a016464e#diff-fb91e91c512c86c97c58e32aadcef8dffa6039a2df09149b9a7784d952942e57L414-R392), which was added between v1.12.1 and v1.12.2, and appears to change the logic for determining what constitutes a partial allocation. The failing test might be checking something that's too internal to libhdf5. Is it worth keeping it?

0:
Modified failing test to be version-dependent.

PR #1077, opened 2023-06-03
Provide high level syntax for dealing with HDF5 files in memory

mkitti:
- Fix get_driver for Core
- Enhance creating a file from memory using the Core driver

### DIFF
--- a/src/api/helpers.jl
+++ b/src/api/helpers.jl
@@ -30,6 +30,27 @@ function h5f_get_dset_no_attrs_hint(file_id)::Bool
     return minimize[]
 end
 
+"""
+    h5f_get_file_image(file_id)
+
+Return a `Vector{UInt8}` containing the file image. Does not include the user block.
+"""
+function h5f_get_file_image(file_id)
+    buffer_length = h5f_get_file_image(file_id, C_NULL, 0)
+    buffer = Vector{UInt8}(undef, buffer_length)
+    h5f_get_file_image(file_id, buffer, buffer_length)
+    return buffer
+end
+
+"""
+    h5f_get_file_image(file_id, buffer::Vector{UInt8})
+
+Store the file image in the provided buffer.
+"""
+function h5f_get_file_image(file_id, buffer::Vector{UInt8})
+    h5f_get_file_image(fild_id, buffer, length(buffer))
+end
+
 ###
 ### Attribute Interface
 ###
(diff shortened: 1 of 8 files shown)


src/api/helpers.jl
```diff
@@ -30,6 +30,27 @@ function h5f_get_dset_no_attrs_hint(file_id)::Bool
...
+    h5f_get_file_image(file_id)
+
+Return a `Vector{UInt8}` containing the file image. Does not include the user block.
+"""
+function h5f_get_file_image(file_id)
+    buffer_length = h5f_get_file_image(file_id, C_NULL, 0)
+    buffer = Vector{UInt8}(undef, buffer_length)
+    h5f_get_file_image(file_id, buffer, buffer_length)
+    return buffer
+end
```
simonbyrne, line 43: Perhaps should be a different function since it has a different return value.

src/api/helpers.jl
```diff
@@ -790,6 +811,27 @@ function h5p_get_virtual_view(dapl_id)
...
+"""
+    h5p_get_file_image(fapl_id)::Vector{UInt8}
+
+Retrieve a file image of the appropriate size in a `Vector{UInt8}`.
+"""
+function h5p_get_file_image(fapl_id)::Vector{UInt8}
+    buf_ptr_ref = Ref{Ptr{Nothing}}()
+    buf_len_ref = Ref{Csize_t}(0)
+    h5p_get_file_image(fapl_id, buf_ptr_ref, buf_len_ref)
+    unsafe_wrap(Array{UInt8}, Ptr{UInt8}(buf_ptr_ref[]), buf_len_ref[]; own=true)
```
simonbyrne, line 823: I'm always a little bit worried about `own=true`: I know the docs say to call `free`, but I thought that this wasn't correct on Windows?

simonbyrne: It might be safer to do the double call: first call with `buf_ptr_ref` as `C_NULL` to get the size of the buffer, then allocate it ourselves, and call it again?

src/api/helpers.jl
mkitti: I have to look back through, but I think this is pretty common to have a helper return a different value.

src/api/helpers.jl
simonbyrne: Oh, i was wrong. You would need to define [`H5Pset_file_image_callbacks`](https://docs.hdfgroup.org/hdf5/v1_12/group___f_a_p_l.html#ga14ea3598215afd078b964b672b40d63c)

simonbyrne: What's strange is that HDF5 explicitly provides [`H5_FREE_MEMORY`](https://portal.hdfgroup.org/display/HDF5/H5_FREE_MEMORY) for this reason elsewhere (see #677).

mkitti: We probably should use that. I'll check the HDF5 source later.

src/api/helpers.jl
mkitti: Did you mean that `h5f_get_file_image(file_id)` should have a different name than `h5f_get_file_image(file_id, buffer::Vector{UInt8})` or `h5f_get_file_image(file_id, buffer, buffer_length)`?

mkitti: That applies to most of the methods in helpers.jl though, does it not?

* The 3-arg method of `h5_get_libversion` returns `0` or throws. The 0-arg helper function returns a `VersionNumber.
* The 1-arg method of `h5_is_library_threadsafe` returns `0` or throws. The 0-arg helper function returns a `Bool`.
* The 3-arg method of `h5a_get_name` returns `0` or throws. The 1-arg helper function returns a `String`.

I believe I'm following page precedent for a function that contains `get` in the name. We have the original C-function as a method but also a Julia friendly helper method that returns what you are getting. Am I misunderstanding your thought here?

simonbyrne: I guess that's true.

src/api/helpers.jl
mkitti: I'm resolving this for now by only allowing the one-arg version with `own=true` if `h5p_get_file_image_callbacks` returns a null pointer for `image_free`.

mkitti: Maybe we should still use `HDF5.API.h5_free_memory`.

PR #1095, opened 2023-08-14
doc typo: attributes(parent), not attribute(parent)

stevengj:


mkitti:
Looks good.

The test failures stem from losing SZIP on Windows, but that is unrelated to the change.

Thanks.

PR #1100, opened 2023-08-30
allow keyword args to open_group, open_dataset, open_datatype

simonbyrne:
It's a bit odd that we allow them for the `create_*` functions, but not the `open_*` functions.

### DIFF
--- a/src/groups.jl
+++ b/src/groups.jl
@@ -50,16 +50,21 @@ function create_group(parent::Union{File,Group}, path::AbstractString; pv...)
 end
 
 """
-    open_group(parent::Union{File,Group}, path::AbstratString)
+    open_group(parent::Union{File,Group}, path::AbstratString; properties...)
 
 Open an existing [`Group`](@ref) at `path` under the `parent` object.
+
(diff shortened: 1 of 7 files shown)


src/groups.jl
```diff
@@ -50,17 +50,22 @@ function create_group(parent::Union{File,Group}, path::AbstractString; pv...)
...
-    parent::Union{File,Group},
-    name::AbstractString,
-    gapl::GroupAccessProperties=GroupAccessProperties()
+    parent::Union{File,Group}, name::AbstractString, gapl::GroupAccessProperties
 )
     return Group(API.h5g_open(checkvalid(parent), name, gapl), file(parent))
 end
+function open_group(parent::Union{File,Group}, name::AbstractString; pv...)
+    gapl = GroupAccessProperties(; pv...)
+    return open_group(parent, path, gapl)
```
mkitti, line 67:
```suggestion
    return open_group(parent, name, gapl)
```

mkitti (changes requested):
A few typos to fix

mkitti:
Looks good to me

PR #1105, opened 2023-09-05
debug windows failures

simonbyrne:


mkitti:
Which Windows failures are you trying to debug? The SZIP test?

simonbyrne:
Yes: any idea what is going on?

I was wondering if it was somehow linked to Julia 1.9.3 (as that is roughly when it appeared), but apparently not.

mkitti:
If it's the SZIP tests, the problem is upstream.

MSYS2 uses cmake to build HDF5.

https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-hdf5/PKGBUILD

HDF Group changed cmake to deprecate SZIP in favor of libaec.

However, HDF Group does this weird thing where they use their own cmake build for libaec, which is different than the way everyone else builds libaec.

https://github.com/HDFGroup/hdf5/blob/develop/config/cmake/LIBAEC/CMakeLists.txt

I'm trying to figure out how to point HDF5 to the libsz that Yggdrasil builds.

PR #1113, opened 2023-09-08
In docs, suggest passing HDF5 module to set_preferences!() instead of UUID

johnomotani:
In the docs, to use custom or system-provided binaries, it is currently suggested to pass the UUID for HDF5.jl to `set_preferences!()`. It is simpler to load the module with `using HDF5`, and pass `HDF5` directly to `set_preferences!()`. Is there a reason this is a bad idea?

This PR upates the docs to suggest the method proposed above.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -63,6 +63,17 @@ set_preferences!(
     "libhdf5_hl" => "/usr/lib/x86_64-linux-gnu/hdf5/mpich/libhdf5_hl.so", force = true)
 ```
 
+If HDF5 cannot be loaded, it may be useful to use the UUID to change these settings:
+
+```julia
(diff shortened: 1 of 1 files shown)


JoshuaLampert:
After providing a convenience function (#1114) this will even simplify further.

mkitti:
We should keep the UUID example as a backup.

One frustrating situation that can occur is that the misconfiguration of the libraries might prevent HDF5.jl from loading, making it difficult to pass HDF5 as an argument to remedy the situation.

johnomotani:
> We should keep the UUID example as a backup.

@mkitti sounds good to me - I've put it back.

mkitti:
LGTM

PR #1115, opened 2023-09-08
Add convenience function `set_libraries!`

### LINKED ISSUE #1114: Convenience function to use custom or system provided HDF5 binaries
Using `Preferences.jl` for the settings to use custom binaries is great! But the method to set this up is now less convenient. Instead of being able to do something short like
```
ENV["JULIA_HDF5_PATH"] = "/path/to/hdf5/dir"
using Pkg, Pkg.build()
```
it is now necessary to import `Preferences`, create two separate paths, and give them both to `set_preferences!()`. It would be lovely to wrap up the process in a convenience function (copying the name from the MPI.jl equivalent), e.g.
```
using HDF5
HDF5.use_system_binary("/path/to/hdf5/dir")
```
If it was possible to auto-detect the directory for system binaries in common cases, that would be awesome, but I don't know how possible that is - maybe parse the output of `h5cc -show`?

JoshuaLampert:
This adds a function `set_libraries!` to conveniently set the preferences for a custom HDF5 library.

closes #1114

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -144,7 +144,7 @@ version = "0.1.1"
 
 [[deps.HDF5]]
 deps = ["Compat", "HDF5_jll", "Libdl", "MPIPreferences", "Mmap", "Preferences", "Printf", "Random", "Requires", "UUIDs"]
-path = ".."
+path = "/home/ldebsao6/.julia/dev/HDF5.jl"
 uuid = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 version = "0.17.0"
 weakdeps = ["MPI"]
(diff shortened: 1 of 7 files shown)


JoshuaLampert:
Do we (alternatively? additionally?)  want to allow the user to pass one path to a directory, where both libraries are located?

JoshuaLampert:
How can I set the docstring without the documenter complaining?

mkitti:
You should add `set_libraries!` to the list of documented functions.

JoshuaLampert:
Thanks @mkitti! I also removed `not_low_level_api`, which appears to be unused and was quite confusing to me.

mkitti:
> Do we (alternatively? additionally?) want to allow the user to pass one path to a directory, where both libraries are located?

I'm still thinking about this. It seems to mainly add complexity

PR #1118, opened 2023-09-15
Update documentation about using files in memory

denglerchr:
An entry in the doc and a small correction of the Drivers.Core doc-string. For recently added functionality in https://github.com/JuliaIO/HDF5.jl/issues/552

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -346,6 +346,32 @@ mtx_dset = create_dataset(g, "M", datatype(Float64), dataspace(100, 1000),
 mtx = HDF5.readmmap(mtx_dset) # succeeds immediately
 ```
 
+## In-memory HDF5 files
+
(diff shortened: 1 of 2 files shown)


mkitti:
Thanks. I will look this over this weekend.

mkitti:
This is mergeable.

I would like to investigate how the default was documented wrong.

Also, I would prefer the `do` syntax with `h5open` as well as using `jldoctest`.

Let me know if you are interested in iterating or not.

denglerchr:
Seems like the default for the Core driver has been documented like this from its creation https://github.com/JuliaIO/HDF5.jl/commit/0eaec093399de9e7fb96ee3cd7b190a71871aed2#diff-eaf75d58da3a8bb4a3287bf539f3167efd5c7e8dc58c23289560013c98ac7b16 
Ill changed to the `do` syntax for the examples

mkitti:
Right, but how did that happen. Is there a mismatch between upstream and this package?

PR #1119, opened 2023-09-18
Fix `h5readattr` docstring

Sbozzolo:
The function does not take a third argument.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -242,7 +242,7 @@ function h5writeattr(filename, name::AbstractString, data::Dict)
 end
 
 """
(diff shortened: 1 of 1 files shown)


mkitti:
You are correct. I think this was just a typo introduced in  
8338e09f05d1e05c83b9e7bf64fb26797191b563 by @simonbyrne .

We could show a type assertion for the return type, but I'm not sure if that is currently the norm.

mkitti:
Test failure is due to `h5a_iterate` failing to throw on Ubuntu.

mkitti:
Fix is valid and only affects documentation. Merging.

PR #1138, opened 2024-01-19
Update helpers.jl with initial value for Ref{Cuint}

### LINKED ISSUE #1137: h5_is_library_threadsafe() gives unreliable results due to unspecified initial value
`HDF5.API.h5_is_library_threadsafe()` has an unspecified initial value in `src/api/helpers.jl`:

```julia
function h5_is_library_threadsafe()
    is_ts = Ref{Cuint}()
    h5_is_library_threadsafe(is_ts)
    return is_ts[] > 0
end
```
and `H5is_library_threadsafe` only updates the low 8 bits:

```Julia
julia> is_ts=Ref{Cuint}(-1 % Cuint)
Base.RefValue{UInt32}(0xffffffff)

julia> ccall((:H5is_library_threadsafe, HDF5.API.libhdf5), HDF5.API.herr_t, (Ref{Cuint},), is_ts);

julia> is_ts[]
0xffffff00
```

david-macmahon:
Fixes #1137

### DIFF
--- a/gen/api_defs.jl
+++ b/gen/api_defs.jl
@@ -27,7 +27,7 @@
 @bind h5_free_memory(buf::Ptr{Cvoid})::herr_t "Error freeing memory"
 @bind h5_garbage_collect()::herr_t "Error on garbage collect"
 @bind h5_get_libversion(majnum::Ref{Cuint}, minnum::Ref{Cuint}, relnum::Ref{Cuint})::herr_t "Error getting HDF5 library version"
-@bind h5_is_library_threadsafe(is_ts::Ref{Cuint})::herr_t "Error determining thread safety"
+@bind h5_is_library_threadsafe(is_ts::Ref{Cuchar})::herr_t "Error determining thread safety"
 @bind h5_open()::herr_t "Error initializing the HDF5 library"
 @bind h5_set_free_list_limits(reg_global_lim::Cint, reg_list_lim::Cint, arr_global_lim::Cint, arr_list_lim::Cint, blk_global_lim::Cint, blk_list_lim::Cint)::herr_t "Error setting limits on free lists"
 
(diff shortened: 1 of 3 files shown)


mkitti:
This is supposed to be a C99 `bool` aka `_Bool`. The appropriate type for the `Ref` should be [`Cuchar`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/#man-bits-types).

https://docs.hdfgroup.org/hdf5/develop/group___h5.html#ga70bfde4acd009cdd7bcd2f54c594e28a

mkitti:
I think we also want to change to Cuchar here:
https://github.com/JuliaIO/HDF5.jl/blob/91ef2841dcc5ddd52ef3094120b8921e0c1c5749/gen/api_defs.jl#L30

david-macmahon:
Yeah, I didn't realize/forgot that a lot of the api files are generated.  I guess the generator doesn't provide an initial value when generating so it relies on having the type size correct.  Maybe best to change it to `is_ts = Ref{Cchar}()` in `helpers.jl` and `Cuchar` in `gen_api.jl`.  I can make the changes, but I don't have cycles to test them.  Does CI regenerate the api files?  I guess we should add a test for this, but how does one test something that relies on an uninitialized value being (or not being) a certain value?

mkitti:
Once you change api_defs.jl, just run the gen_wrappers.jl script as indicated below. If you cannot, I would be happy to do it.

https://github.com/JuliaIO/HDF5.jl/blob/91ef2841dcc5ddd52ef3094120b8921e0c1c5749/gen/gen_wrappers.jl#L2

simonbyrne:
Maybe we should switch to Clang.jl-generated wrappers.

mkitti:
I worked on that a few years ago in LibHDF5.jl:
https://github.com/mkitti/LibHDF5.jl/blob/main/src/bind.jl

mkitti:
I'm going to try to push this through soon

PR #1151, opened 2024-05-26
Fix H5Zzstd error reporting and decompressed size behavior

mkitti:
https://github.com/JuliaIO/JLD2.jl/pull/560 revealed that error reporting was not occuring when there were errors present.

1. `ZSTD_getDecompressedSize` can result in an error, resulting in a return value of 0.
2. `Libc.malloc(0)` may not return a null pointer.

Therefore we need to check the return value of `ZSTD_getDecompressedSize` directly before
trying to malloc.

Also improve error handing overall.


mkitti:
Also note that `ZSTD_getDecompressedSize` is deprecated.

PR #1152, opened 2024-05-27
Guard malloc in compression filters

mkitti:
Providing a length of `0` bytes to malloc could result in a valid pointer being generated. This is probably an error.

This pull request detects this condition and errors if detected. Also improves error messaging for failed allocation.

### DIFF
--- a/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
+++ b/filters/H5Zbitshuffle/src/H5Zbitshuffle.jl
@@ -196,6 +196,7 @@ function H5Z_filter_bitshuffle(
         end
 
         size = nbytes_uncomp ÷ elem_size
+        buf_size_out <= 0 && error("bitshuffle_h5plugin: Non-positive buf_size_out for malloc: $buf_size_out")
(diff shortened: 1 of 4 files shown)


mkitti:
Thanks for the fix. Also, should there be a way to just accept the formatter suggestions?

musm:
Do you mean to use https://github.com/julia-actions/julia-format 
```
name: Format suggestions
on:
  pull_request:
    # this argument is not required if you don't use the `suggestion-label` input
    types: [ opened, reopened, synchronize, labeled, unlabeled ]
jobs:
  code-style:
    runs-on: ubuntu-latest
    steps:
      - uses: julia-actions/julia-format@v3
        with:
          version: '1' # Set `version` to '1.0.54' if you need to use JuliaFormatter.jl v1.0.54 (default: '1')
          suggestion-label: 'format-suggest' # leave this unset or empty to show suggestions for all PRs
```

mkitti:
Trying julia-format in #1153 , might merge that into a copy of this branch to see how it behaves.

PR #1153, opened 2024-05-30
Switch to julia-format action

mkitti:
Switch to the julia-format action:
https://github.com/marketplace/actions/format-suggestion-with-juliaformatter-jl
https://github.com/julia-actions/julia-format


mkitti:
I have a feeling that we may want to customize this:
https://github.com/julia-actions/julia-format/blob/be8e90b6b386d0dbcff7504ba067aa2cd96c5948/action.yml#L28

PR #1160, opened 2024-06-27
Convert H5Z*.jl packages into package extensions

mkitti:
The H5Z*.jl packages such as H5Zblosc.jl implement HDF5 filter plugins.
These were written before the advent of package extensions in Julia 1.9.

This pull request converts the implementations of the H5Z*.jl packages
into package extensions. The H5Z*.jl packages are transitional packages
that are shells of their former selves. They only serve two purposes:

1. Load the weak dependency to trigger the package extension to load.
2. Export the filter

After this pull request loading the H5Z*.jl packages could be replaced
by loading the weak dependency directly. For example H5Zblosc.jl could
be replaced by loading Blosc.jl. Another example is that H5Zlz4.jl
could be replaced by loading CodecLZ4.jl.

Exporting the filter could be provided by leaving a function stub
with no methods in HDF5.jl. For example, `function BloscFilter end`
could be declared in HDF5.jl and then implemented in the package
extensions.


simonbyrne:
I like this idea: it would be much simpler to maintain.

PR #1162, opened 2024-07-17
Fix Base.parent by splitting on / only, fix #1161

### LINKED ISSUE #1161: Base.parent does not work if dataset contains a backslash
For a dataset with a name of "/X/somelab/RW10557/031021/Pos0/segmentation/RegB/045/E10\\", `Base.parent` does not work. It attempts to look for "/X/somelab/RW10557/031021/Pos0/segmentation/RegB/045/E10" since that is the result of `dirname`.

Instead we should strictly look for `/`. This is still possibly incorrect since the name of the dataset could contain a `/`.

mkitti:
Fix `Base.parent`, fix #1161, by splitting only on `/` and not `\`.


nhz2 (approved):
This looks great

PR #1166, opened 2024-10-09
Move FileIO support into extension

topolarity:
This moves one of the two remaining uses of Requires.jl into a pair of extensions.


topolarity:
Invalidations check appears to be an unrelated failure

mkitti:
@simonbyrne @musm any thoughts about this pull request? Does this require a version bump?

simonbyrne (approved):

mkitti:
I will try to sort these out this weekend.

mkitti:
I'm leaning towards dropping Requires entirely with 0.18

PR #1167, opened 2024-10-09
Move `set_blosc!` methods to H5ZBlosc.jl

topolarity:
That's the last one gone!

As far as I can tell, Requires was not necessary here since the package it was "extending" already depends on HDF5.jl


mkitti:
I need to think about the interaction with #1160. These methods were mainly there was backwards compatibility.

topolarity:
For what it's worth, the methods are still there - I think this should behave pretty much as it did before, except that because we can't set retroactive `[compat]` bounds, it's possible to have, e.g., an old version of H5Zblosc with the new HDF5.jl that will be missing the methods.

mkitti:
@simonbyrne @musm any thoughts about this pull request? Does this require a version bump?

Project.toml
```diff
@@ -22,6 +22,7 @@ MPIExt = "MPI"
 
 [compat]
 Compat = "3.1.0, 4"
+H5Zblosc = "≥ 0.1.3"
```
simonbyrne, line 25: Shouldn't this be covered by the compat in H5Zblosc?

simonbyrne (approved):
I don't _think_ it should be breaking, but I guess it could break version resolution?

PR #1169, opened 2024-10-16
Update docs build

ajwheeler:
This PR:
- deletes `test/Manifest.toml`, which I think is being tracked by mistake
- adds a compat entry for Documenter in the test `Project.toml`
- tweaks `docs/make.jl` to work with the newest Documenter versions
- Drops `UNLIMITED` and `is_selection_valid` from `@docs` blocks because they don't seem to be defined.
    - I don't understand what's going on with this one.  I see them in the source, but they don't seem to be in the current docs, and I can't access them if I load the package myself.

**edit:** sorry for submitting something which fails, but I think this should be straightforward to fix for someone who understands what's going on with `UNLIMITED` and `is_selection_valid`.


mkitti:
Thank you for the help and the hints for how to fix things. I'm heading in a slightly different direction with the size_threshold for now.

PR #1170, opened 2024-10-23
Fix H5Oexist test result after HDF5 bugfix

derobins:
A bugfix in the HDF5 library changed the return value of the H5Oexists(_by_name) API call. In earlier versions of the library, when the ID was a file ID and the object didn't exist, the call would return FAIL (-1). The correct behavior is to return false (0). This change will appear in the next released version of HDF5, which is planned to be 2.0.0.

This change updates the object test to expect FAIL for 1.14.5 and earlier and false for later versions.

### DIFF
--- a/test/objects.jl
+++ b/test/objects.jl
@@ -16,7 +16,13 @@ using HDF5.API
     h5open(fn, "r") do h5f
         @test API.h5o_exists_by_name(h5f, "data")
         @test API.h5o_exists_by_name(h5f, "lore")
-        @test_throws API.H5Error API.h5o_exists_by_name(h5f, "noonian")
+        @static if HDF5.API.h5_get_libversion() <= v"1.14.5"
+            # Buggy behavior in earlier versions of HDF5 returns FAIL (-1)
+            @test_throws API.H5Error API.h5o_exists_by_name(h5f, "noonian")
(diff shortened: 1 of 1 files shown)


derobins:
I'm applying this patch in our CI, so getting this in ASAP isn't critical

simonbyrne (approved):

test/objects.jl
```diff
@@ -16,7 +16,13 @@ using HDF5.API
     h5open(fn, "r") do h5f
         @test API.h5o_exists_by_name(h5f, "data")
         @test API.h5o_exists_by_name(h5f, "lore")
-        @test_throws API.H5Error API.h5o_exists_by_name(h5f, "noonian")
+        @static if HDF5.API.h5_get_libversion() <= v"1.14.5"
```
simonbyrne, line 19: @derobins Are there no further planned patch releases on 0.14? Would it be safer to use
```suggestion
        @static if HDF5.API.h5_get_libversion() <= v"2"
```

derobins: If there were another release of 1.14, it would include this bugfix, so >= 1.14.5 is probably safer.

mkitti:
Thank you, @derobins .

PR #1176, opened 2024-11-14
Drop Requires.jl, bump to 0.18

mkitti:
This pull request is in prepration for a 0.18 release.

* Minimum Julia version is now Julia 1.9 for extension support
* Requires.jl support is dropped in favor of the extensions
* Filter packages are just helpers that trigger the extension mechanism

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "HDF5"
 uuid = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
-version = "0.17.1"
+version = "0.18.0"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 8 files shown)


bclyons12:
What's the status of this? I'm trying to build a relocatable sysimage with PackageCompiler.jl that includes HDF5.jl, and I'm running into the issues noted here with the FileIO @require statement.
https://github.com/JuliaPackaging/Requires.jl/issues/112
https://discourse.julialang.org/t/relocability-problems-about-requires-jl/96110

bclyons12:
Our full sysimage is nontrivial to test. I tried a simplified version but couldn't reproduce the error. Nevertheless, if this PR removes all the @require statements, then I can't see how this wouldn't solve the problem.

hz-xiaxz:
> Yes, but I'm not sure if I would release immediately after merging this.

sorry, it looks like it's yet released.

Issue #1216, opened 2025-12-28
Make `public` certain documented functions


JeffFessler:
According to the [history file](https://github.com/JuliaIO/HDF5.jl/blob/master/HISTORY.md), in v0.14 a bunch of methods were unexported and require the `HDF5` module prefix, which is fine.  See list below.
Nevertheless, should these methods all be declared `public` so that developers can rely on them?

The one I care about the most is `readmmap` and that one is described in the [docs](https://juliaio.github.io/HDF5.jl/stable/#Memory-mapping) which certainly seems to imply it is `public`
I am unsure which of the rest of this list should be considered public (and hence stable).

```julia
file
filename
name
get_chunk
get_datasets
iscontiguous
ishdf5
ismmappable
root
readmmap
set_dims!
get_access_properties
get_create_properties
create_external_dataset
```

mkitti:
Thank you for the list. We have not used the `public` keyword extensively yet, but this might be a good starting point 

Would you like to submit a pull request?

JeffFessler:
Sure - see #1217 as a starting point.  Thanks.

JeffFessler:
closed by #1217

PR #1217, opened 2025-12-31
Declare `public` methods

JeffFessler:
Addresses #1216 by declaring public methods as such (those that are not exported and thuse require module scoping).

I used the union of the list of methods in the history file:
https://github.com/JuliaIO/HDF5.jl/blob/master/HISTORY.md#v014
and the methods mentioned in `HDF5.jl`

Any maintainer (or I) can edit the list in this PR to remove any names that should not be public.

This works only in Jula v0.11.0 and beyond.  Currently, `eval Expr` is needed to avoid problems with code coverage in earlier versions.  Eventually, when the long-term release is >= 0.11, then this code can be simplified to avoid `eval Expr`

I learned this Eval trick over at https://github.com/JuliaMath/FFTW.jl/pull/329

### DIFF
--- a/src/HDF5.jl
+++ b/src/HDF5.jl
@@ -51,6 +51,7 @@ export @read,
 
 @static if VERSION ≥ v"1.11.0"
     eval(Expr(
+        :public,
         :create_external,
         :create_external_dataset,
         :file,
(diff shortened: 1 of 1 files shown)


mkitti:
Thank you!

mkitti (approved):

PR #1230, opened 2026-05-01
Remove unneeded `cglobal` in `cglobal(dlsym(...))`

topolarity:
The `cglobal` is not needed here.

We'd like to consider deprecating / removing this `cglobal(::Ptr)` support in Julia 1.14 (https://github.com/JuliaLang/julia/pull/61707).

### DIFF
--- a/src/api/types.jl
+++ b/src/api/types.jl
@@ -253,7 +253,7 @@ end
 # we can precompile for improved latency.
 const libhdf5handle = Ref(dlopen(libhdf5))
 ccall(dlsym(libhdf5handle[], :H5open), herr_t, ())
-_read_const(sym::Symbol) = unsafe_load(cglobal(dlsym(libhdf5handle[], sym), hid_t))
+_read_const(sym::Symbol) = unsafe_load(Ptr{hid_t}(dlsym(libhdf5handle[], sym)))
 _has_symbol(sym::Symbol) = dlsym(libhdf5handle[], sym; throw_error=false) !== nothing
 
 # iteration order constants


mkitti:
> ERROR: LoadError: TypeError: in cglobal: first argument not a pointer or valid constant expression, expected Ptr, got a value of type Tuple{Ptr{Nothing}, Symbol}

This does not seem to work...

mkitti:
Do we need a macro here instead?

topolarity:
Ah right this is a dynamic usage, sorry about that.

The error is a bit obtuse, but it means that the argument is not known statically. The new form with `Ptr{T}(dlsym(...))` should work (and be equivalent to what was there before).

mkitti:
Kind of. We actually use all of these statically...
https://github.com/JuliaIO/HDF5.jl/blob/66ee074039ddbc1cbe23a32bfda07afd808656e6/src/api/types.jl#L385-L401

mkitti:
Would there be any advantage or disadvantage to doing:

```julia-repl
julia> macro hdf5_global(ex)
           quote
               unsafe_load(Ptr{UInt64}(cglobal(($ex, HDF5_jll.libhdf5_path))))
           end
       end
@hdf5_global (macro with 1 method)

julia> @hdf5_global :H5T_STD_I16BE_g
0x030000000000004e
```

topolarity:
Yes, that will opt in to Julia's built-in caching, rather than calling `dlsym` on every load

The macro would be a performance improvement over the status quo, in that case

mkitti:
At the moment, we assign everything to a top-level const, so I guess this results in another layer of caching that gets stored in the precompile cache and may be a bit fragile.

```julia
 const H5P_DATASET_ACCESS   = _read_const(:H5P_CLS_DATASET_ACCESS_ID_g) 
```

If someone switched out the HDF5 library in such a way that does not HDF5.jl to recompile, these constants might be invalid. I will merge this for now and consider refactoring later.

PR #1244, opened 2026-09-16
Split HDF5 C API doc-link generation into HDF5DocURLs.jl

mkitti:
## Summary
- Extracts the Doxygen-tag-parsing tool, vendored `hdf5.tag`, and derived URL-lookup TSVs into a new, independent package: [HDF5DocURLs.jl](https://github.com/mkitti/HDF5DocURLs.jl) (pending registration in the General registry; a `[sources]` entry in `Project.toml` points at the GitHub repo in the meantime).
- Docstrings in `src/api/functions.jl`/`src/api/helpers.jl` no longer bake resolved doc URLs as literal strings at generation time. Instead, `gen/bind_generator.jl` now emits escaped `$(_hdf5_func_doc_url("H5Fopen"))`-style interpolation text into the generated source. Julia evaluates that interpolation **exactly once**, when `HDF5.jl`'s own `src/api/functions.jl` is loaded during precompilation — so doc links reflect whichever `HDF5DocURLs` version is installed at precompile time, without requiring a new HDF5.jl release to pick up a refreshed link. Verified directly: `@doc HDF5.API.h5f_open` resolves to a real, live URL, not literal interpolation text.
- Fixes the 6 hardcoded, stale `portal.hdfgroup.org` links in `src/api/helpers.jl` (two of which linked to the entirely wrong function, `H5Oopen`) to use the same mechanism with corrected canonical C function names.
- Deletes `gen/DoxygenTagParser/` — that tooling now lives in `HDF5DocURLs.jl`'s own `gen/` directory.

This is the first of a 3-part stack (doc-links package → `AbstractDict` conformance on `H5DataStore` for REPL tab completion → `DiskArrays.jl` interface on `Dataset`) constituting the HDF5.jl **v0.18** release.

## Test plan
- [x] `Pkg.resolve()`/`Pkg.instantiate()` succeed, resolving `HDF5DocURLs` from its GitHub repo via the temporary `[sources]` entry
- [x] Regenerated `src/api/functions.jl` diffed to confirm only doc-URL construction text changed (505 sites), not function signatures/bodies; `docs/src/api_bindings.md` unchanged (same underlying data)
- [x] `@doc HDF5.API.h5f_open` in a fresh session resolves to a real `support.hdfgroup.org` URL, confirming the precompile-time interpolation mechanism works as intended
- [x] Full test suite passes: 1514 passed, 4 pre-existing broken, 0 failures

## Follow-up (not part of this PR)
Once `HDF5DocURLs.jl` is registered, a separate PR backports the same dependency + mechanism to `release-0.17` (and similar), so doc-link fixes ship to every HDF5.jl release line via pure `HDF5DocURLs` releases, without a new HDF5.jl release.


mkitti:
**Update: HDF5DocURLs restructured as a monorepo subdirectory package.**

Per further discussion, `HDF5DocURLs` will live inside this repo at `HDF5DocURLs/` (matching the existing `filters/H5Zblosc`-style monorepo-subpackage convention) rather than as a standalone repo at `mkitti/HDF5DocURLs.jl`.

This also happens to fix the CI failure seen on this PR (`ERROR: LoadError: expected package HDF5DocURLs to be registered`): `Pkg.test()`'s sandboxed test environment failed to resolve the dependency when it was sourced via an external git URL (`[sources] = {url = "...", rev = "main"}`), but resolves cleanly when sourced via a local `path =` entry pointing at a directory that's part of the same checkout — verified locally, full test suite passes (1514 passed, 4 pre-existing broken, 0 failures), and the precompile-time doc-link interpolation mechanism is unaffected.

The corrected commit is pushed to [`mkitti/HDF5.jl@hdf5-doc-urls-split-monorepo`](https://github.com/mkitti/HDF5.jl/tree/hdf5-doc-urls-split-monorepo) (commit `8c1146c4`). This still needs to be force-pushed onto this PR's `hdf5-doc-urls-split` branch to update it in place — that step requires manual confirmation and hasn't been done yet.

The standalone `mkitti/HDF5DocURLs.jl` repo is left as-is for now; that's a separate decision for later.

PR #1249, opened 2026-09-17
Support negative compression levels in ZstdFilter

bjarthur:
The Zstandard HDF5 filter stores the compression level as an unsigned integer. The Julia port converts it with `Cint(unsafe_load(cd_values))`, which throws an `InexactError` for the bit pattern of a negative level, and then clamps levels below 1 to 1. The upstream C plugin (HDFGroup/hdf5_plugins@de55b7c4) uses the cast `(int)cd_values[0]` and now allows negative levels, which are faster at the cost of compression, clamping to `ZSTD_minCLevel()..ZSTD_maxCLevel()`.

This PR mirrors that behaviour in `CodecZstdExt`:

- reinterpret the stored value with `% Cint` instead of the checked conversion
- clamp to `ZSTD_minCLevel():ZSTD_maxCLevel()` (same as the C plugin)
- `ZstdFilter(clevel::Integer)` stores the bit pattern, so `ZstdFilter(-3)` constructs; `show` prints the signed level
- test in `test/filter.jl` writing and reading back a dataset with `ZstdFilter(-3)`, checking the level round-trips (`clevel % Cint == -3`)

Motivation: netCDF (`nc_def_var_zstandard`) accepts negative levels, and writing such a variable through NCDatasets.jl with H5Zzstd loaded currently produces a filter failure (`H5Zzstd Non-Fatal ERROR: InexactError: convert(Int32, 0xfffffffd)`); see JuliaGeo/NCDatasets.jl#318. Note the released H5Zzstd 0.1.x has the same code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bjarthur:
@mkitti

mkitti:
Yes, I changed this upstream:

https://github.com/HDFGroup/hdf5_plugins/pull/192

mkitti:
Did this really require Fable?

bjarthur:
yea, i should've switched models at least when fixing the formatting.

relately, consider upgrading to JuliaFormatter v2.  there are differences with the v1 currently used here.

